PR #2, opened 2018-03-20
Adding ETW Events, Global Exception handlers(Middleware)

ShekharGupta1988:
ETW Provider : "Microsoft-Azure-AppService-Diagnostics"

Events ID Range:
Compiler Host : 1000 - 1999
Runtime Host : 2000 - 2499
Source Watcher : 2500 - 2599
Compiler Host Client : 2600 - 2699
Data Providers : 3000 - 3999


src/Diagnostics.RuntimeHost/Controllers/SitesController.cs
```diff
@@ -62,6 +63,8 @@ public SitesController(ICompilerHostClient compilerHostClient, ISourceWatcherSer
                 return BadRequest(errorMessage);
             }
 
+            string requestId = this.Request.Headers[HeaderConstants.RequestIdHeaderName];
```
stevenernst131: this will return null and not throw key not found exception right?

ShekharGupta1988: I should do try get value. Fixing at all places

src/Diagnostics.Logger/ApiMetricsLogger.cs
```diff
@@ -0,0 +1,183 @@
...
+            try
+            {
+                if (String.IsNullOrWhiteSpace(httpContext.Request.Path.ToString()))
+                {
+                    return;
+                }
+
+                var parts = httpContext.Request.Path.ToString().ToLowerInvariant().Split('?')[0].Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+                var cursor = ((IEnumerable<string>)parts).GetEnumerator();
+                while (cursor.MoveNext())
```
hforeste: Instead of this way of parsing what you want from the request path, consider using regular expressions and match groups. https://www.dotnetperls.com/regex-groups

Example: subscriptions/SomeGuid/resourceGroups/SomeName/providers/Microsoft.Web/sites/SomeSite

You can have a regex like `"subscribtions/(.*)/resourceGroups/(.*)/providers/Microsoft.Web/sites/(.*)"` and each of these can match to a group and you basically just say `_subscriptionId = Regex.Group(1)`

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            WriteDiagnosticsEvent(1001,
+                RequestId,
+                Source,
+                ExceptionType,
+                ExceptionDetails);
+        }
+
+        [Event(1002, Level = EventLevel.Informational, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogCompilerHostAPISummary)]
+        public void LogCompilerHostAPISummary(string RequestId, string Address, string Verb, int StatusCode, long LatencyInMilliseconds, string StartTime, string EndTime)
```
hforeste: We should also log the Id of the detector id specified in the csx file. I can see near future value in tracking that kind of telemetry.

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
```diff
@@ -0,0 +1,110 @@
...
+                    string.Empty,
+                    string.Empty,
+                    logEx.GetType().ToString(),
+                    logEx.ToString());
+            }
+        }
+
+        private void GenerateMissingRequestHeaders(ref HttpContext httpContext)
+        {
+            if (!httpContext.Request.Headers.TryGetValue(HeaderConstants.RequestIdHeaderName, out StringValues values)
```
hforeste: Is there a scenario where RequestId won't be in the headers?

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
```diff
@@ -0,0 +1,110 @@
...
+                {
+                    logger = new ApiMetricsLogger(context);
+                }
+
+                logger.LogRuntimeHostUnhandledException(context, ex);
+            }
+            catch (Exception logEx)
+            {
+                DiagnosticsETWProvider.Instance.LogRuntimeHostUnhandledException(
+                    string.Empty,
```
hforeste: I think we should track requestId here for correlation purposes.

src/Diagnostics.CompilerHost/Middleware/CompilerRequestMiddleware.cs
```diff
@@ -0,0 +1,96 @@
...
+                {
+                    logger = new ApiMetricsLogger(context);
+                }
+
+                logger.LogCompilerHostUnhandledException(context, ex);
+            }
+            catch (Exception logEx)
+            {
+                DiagnosticsETWProvider.Instance.LogCompilerHostUnhandledException(
+                    string.Empty,
```
hforeste: Same as with the DiagnosticRequestMiddleware, track requestId for correlation purposes

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
ShekharGupta1988: for that, you can use request id as that is getting passed

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
ShekharGupta1988: yes. Headers is not a mandatory requirement from client perspective.

ShekharGupta1988: one of the clients of this host is w3wp process, which is passing the header. Other client could be local cli which might not be passing the header..

src/Diagnostics.CompilerHost/Middleware/CompilerRequestMiddleware.cs
ShekharGupta1988: sure.. This code path will get executed only when there is an exception in ETW logging. (I dont expect this to get hit lot of times. but ya, I can put reqquest id there.)

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
hforeste: It would be difficult to derive useful insights about a detector with the requestid since there is a unique requestid for each compilation.

 It might be helpful to have detectorId so that you can pivot on that and get useful telemetry. For example: average amount time roselyn has been taking to compile this detector in the past 7 days. That information could be useful for us to optimize our platform.

ShekharGupta1988: for compiler API request, there is no detector Id. The input is a general script csx file.  
Those kind of insights can easily be drawn doing a join between 2002(detector api event) and 1002(compiler api event)

src/Diagnostics.CompilerHost/Middleware/CompilerRequestMiddleware.cs
hforeste: Ya i was thinking the same about the ETW logging too. But I guess in those edge cases maybe this will help

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
ShekharGupta1988: oh no buddy. 
They will not go to production kusto. Local development events will not go to kusto as there is no mds logger running on your box.

hforeste: Ah true. Thanks for clarification. So where will the logs go then, EventViewer?

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
hforeste: I see so I would do a join on both of those tables and look for the detector name in the address is the way to do it right?

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
ShekharGupta1988: No. These are ETW events (generated from Event source). To locally view them, you need to use perfview

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
ShekharGupta1988: See the compilation is not happening for every detector (since the assemblies are already compiled and saved). Compilation only happens for one API (/query). So that insight you mentioned will not be useful.
What you want to know would be what's the average compilation time for script, how many times the memory threshold is breached (from my experiements, I haven't seen  any variations in compilation time per detector)

PR #3, opened 2018-03-21
Adding rendering models

stevenernst131:


stevenernst131:
These are not final, but first cut. Will finalize before I go with geomaster change

PR #4, opened 2018-03-21
Integrate observer pass through api

hforeste:


src/Diagnostics.DataProviders/DataProviders/ISupportObserverDataProvider.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Diagnostics.DataProviders
+{
+    public interface ISupportObserverDataProvider
+    {
+        Task<dynamic> GetSite(string siteName);
+        Task<dynamic> GetSite(string stampName, string siteName);
+        Task<IEnumerable<string>> GetSiteHostNames(string siteName);
```
stevenernst131: Can you add the getsitehostnames(siteName, stampName) for which there is an open bug about? Or just combine them in the first go

tests/DataProviderTests/SupportObserverTests.cs
```diff
@@ -0,0 +1,42 @@
...
+            var configurationFactory = new MockDataProviderConfigurationFactory();
+            var configuration = configurationFactory.LoadConfigurations();
+            _dataProvider = new SupportObserverDataProvider(null, configuration.SupportObserverConfiguration);
+        }
+
+        [Fact]
+        public void GetHostnamesTest()
+        {
+            var hostnames = _dataProvider.GetSiteHostNames("realforsikring").Result;
+            Assert.Equal(3, hostnames.Count());
```
ShekharGupta1988: use async await. XUnit supports that

hforeste:
Quick Note. SupportObserverDataProvider is not implementing everything from the ISupportObserverDataProvider interface. I just implemented the things that will quickly give Puneet the ability to write some detectors. If this is approved as it is then I will do work to integrate the entity fraemwork and its own datamodels and the such very soon afterwards

tests/DataProviderTests/SupportObserverTests.cs
hforeste: awesome!

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task<string> GetServerFarmWebspaceName(string subscriptionId, string serverFarm)
+        {
+            throw new NotImplementedException();
+        }
+
+        public async Task<IEnumerable<string>> GetSiteHostNames(string siteName)
```
stevenernst131: Yeah given the bug I just mentioned, it seems like it is incorrect to ship this API as is, since it can give incorrect data

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            var response = await GetObserverResource($"sites/{siteName}/hostnames?api-version=2.0");
+            var hostnames = response.Remove(0, 1).Remove(response.Length - 2, 1).Split(new char[] { ',' });
```
stevenernst131: What are these removes doing?

hforeste:
Looks like the build failed because I'm making AAD calls to get token and the client id and appkey is not supplied. I may need to remove those tests

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
hforeste: Honestly, I may remove this entirely seeing that we require hostnames to be passed through the API

ShekharGupta1988: Will hostnames ask be fulfilled by geomaster work?

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
hforeste: The response is like a JSON Array. I'm just removing the opening and closing brackets

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
stevenernst131: Even without the site information in the request body, we have them from the query parameters. Probably don't need a method to get them again, at least I can't think of a reason.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            return await GetSite(siteName);
+        }
+
+        private async Task<string> GetObserverResource(string url)
+        {
+            var request = new HttpRequestMessage(HttpMethod.Get, url);
+            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", await GetAccessToken());
+            var response = await _httpClient.SendAsync(request);
+            response.EnsureSuccessStatusCode();
```
ShekharGupta1988: How is the exception thrown here handled at caller side. Should we log an event indicating DataSource Errored out.

Right now, there is no ETW event in Data providers project as we recently added ETW for main API calls.
As a separate work-item, It will be nice to add few common events like summary event about latency and exception to some base class so that every data provider gets it automatically.

hforeste: The exception here is going to be tracked by the ETW events issued at the API level. The diagnostic logger middleware that you created should capture the exception details and show the stack trace in our kusto logs so that we can see where the exception is thrown from.

Yes I agree about the summary event for data providers. I added an [issue](https://github.com/ShekharGupta1988/AppLensV2/issues/1061) for this

ShekharGupta1988: oh cool.

ShekharGupta1988:
Looks good. Just resolve conflicts and merge

PR #12, opened 2018-04-06
Add observer passthrough pt2

hforeste:
-Reference WAWS Observer API endpoint to get database data.
-Create sample detector. Example, BackupCheckDetector
-Still reference support bay api for RuntimeSiteSlotMap data

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/MockSupportObserverDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/MockSupportObserverDataProvider.cs
@@ -88,7 +88,7 @@ namespace Diagnostics.DataProviders
 
         public async Task<dynamic> GetSite(string siteName)
         {
-            return await GetSiteInternal();
+            throw new NotImplementedException();
         }
 
         public async Task<dynamic> GetSite(string stampName, string siteName)
--- a/src/Diagnostics.DataProviders/DataProviders/MockSupportObserverDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/MockSupportObserverDataProvider.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Text;
 using System.Threading.Tasks;
 using Diagnostics.ModelsAndUtils.Models;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Diagnostics.DataProviders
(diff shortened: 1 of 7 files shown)


src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -53,16 +53,22 @@ public Task<JObject> GetAppServiceEnvironmentDetails(string hostingEnvironmentNa
...
-            var slotTimeRangeCaseSensitiveDictionary = JsonConvert.DeserializeObject<Dictionary<string, List<RuntimeSitenameTimeRange>>>(response);
+            return await GetRuntimeSiteSlotMapInternal(stampName, siteName);
+        }
+
+        private async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMapInternal(string stampName, string siteName)
+        {
+            var request = new HttpRequestMessage(HttpMethod.Get, string.IsNullOrEmpty(stampName) ? $"/sites/{siteName}/runtimesiteslotmap" : $"stamp/{stampName}/sites/{siteName}/runtimesiteslotmap");
+            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", await GetAccessToken(_configuration.RuntimeSiteSlotMapResourceUri));
+            var response = await _httpClient.SendAsync(request);
+            response.EnsureSuccessStatusCode();
```
stevenernst131: Do we want to handle errors in a way that we can log any error coming from the API?

src/Diagnostics.DataProviders/DataProviders/MockSupportObserverDataProvider.cs
```diff
@@ -63,10 +64,10 @@ public Task<JObject> GetAppServiceEnvironmentDetails(string hostingEnvironmentNa
```
stevenernst131: Above it appears like you are throwing a null exception when stamp is null, but in the method right above you are explicitly calling it with stamp as null so that will never work right?

hforeste: No actually it was unintentional. I don't think you have to worry over this one though. If anything I should remove it because we are always going to call the overloaded version with stampName and siteName

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -53,16 +53,22 @@ public Task<JObject> GetAppServiceEnvironmentDetails(string hostingEnvironmentNa
...
         public async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string stampName, string siteName)
         {
-            var response = await GetObserverResource($"stamp/{stampName}/sites/{siteName}/runtimesiteslotmap");
-            var slotTimeRangeCaseSensitiveDictionary = JsonConvert.DeserializeObject<Dictionary<string, List<RuntimeSitenameTimeRange>>>(response);
+            return await GetRuntimeSiteSlotMapInternal(stampName, siteName);
+        }
+
+        private async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMapInternal(string stampName, string siteName)
+        {
+            var request = new HttpRequestMessage(HttpMethod.Get, string.IsNullOrEmpty(stampName) ? $"/sites/{siteName}/runtimesiteslotmap" : $"stamp/{stampName}/sites/{siteName}/runtimesiteslotmap");
```
ShekharGupta1988: use IsNullOrWhiteSpace

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -116,28 +122,30 @@ public Task<string> GetWebspaceResourceGroupName(string subscriptionId, string w
...
             return result;
         }
 
-        private async Task<string> GetAccessToken()
+        private async Task<string> GetAccessToken(string resourceId = null)
         {
             if (_authContext == null)
             {
                 _aadCredentials = new ClientCredential(_configuration.ClientId, _configuration.AppKey);
                 _authContext = new AuthenticationContext("https://login.microsoftonline.com/microsoft.onmicrosoft.com", TokenCache.DefaultShared);
```
ShekharGupta1988: This is subjected to race when multiple calls hit this code piece.

tests/DataProviderTests/SupportObserverTests.cs
```diff
@@ -36,15 +36,49 @@ public SupportObserverTests()
...
 
                 Response result = (Response)await invoker.Invoke(new object[] { dataProviders, operationContext, response });
 
-                Assert.Equal("thor-api__a88nf", result.Dataset.First().Table.Rows[1][1]);
+                Assert.Equal("my-api__a88nf", result.Dataset.First().Table.Rows[1][1]);
+            }
+        }
+
+        [Fact]
+        public async System.Threading.Tasks.Task E2E_Test_WAWSObserverAsync()
```
ShekharGupta1988: you can remove the Fully Qualified name

PR #14, opened 2018-04-10
Use Datatable in scripts but morph into correct return type

stevenernst131:
_FYI will not be pushing this out until after CSS demo wednesday 4/11_

The motivation for this change is the feedback we got from @puneetg1983 about the difficulty with working with DataTableResponseObjects. We want script authors to be able to use DataTables to post process data because we think that makes for a better experience. This means the following:

1. The kusto data provider will not return queries as data tables
2. The response object that is exposed to script writers will have a data table as the data type in the DiagnosticData object. 
3. Still TODO: We need to create methods that will take a list of objects and make it into a data table. I may try to do that as well as part of this change if I have a chance before I actually merge. 
4. After the CSX returns the data, we will morph it into the correct return type, so that they UI code does not need to change.


src/Diagnostics.ModelsAndUtils/DataTableUtility.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            dataTableResponseObject.Columns = columns;
+            dataTableResponseObject.Rows = rows.ToArray();
+
+            return dataTableResponseObject;
+        }
+
+        internal static Type GetColumnType(string type, string datatype)
+        {
+            return Type.GetType($"System.{datatype}");
```
hforeste: Is `GetType()` case-sensitive? Also check this out, I think these are all datatypes supported by kusto that could possibly return from the HTTP call. https://kusto.azurewebsites.net/docs/queryLanguage/scalar-data-types/bool.html.

I notice that one of the data types is `real` and .NET doesn't have `System.Real`.

stevenernst131: real maps to System.Double, so a real column returns something like this:

`
{
  columnName: "ColumnName"
  dataType: "Double"
  columnType: "real"
}
`

In this case I am looking at data type, which is System.xxxx and the column type is the kusto scalar type

src/Diagnostics.ModelsAndUtils/DataTableUtility.cs
```diff
@@ -0,0 +1,73 @@
+﻿using Diagnostics.ModelsAndUtils;
+using Diagnostics.ModelsAndUtils.Models;
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Linq;
+
+namespace Diagnostics.ModelsAndUtils
+{
+    public static class DataTableUtility
```
ShekharGupta1988: There are 2 classes for DataTableUtility? One in data providers and other in ModelsAndUtils

src/Diagnostics.ModelsAndUtils/DataTableUtility.cs
```diff
@@ -0,0 +1,73 @@
+﻿using Diagnostics.ModelsAndUtils;
+using Diagnostics.ModelsAndUtils.Models;
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Linq;
+
+namespace Diagnostics.ModelsAndUtils
```
ShekharGupta1988: If this is a utility, move this inside the utilities folder and update the namespace

src/Diagnostics.ModelsAndUtils/DataTableUtility.cs
stevenernst131: No I moved it, is it still showing up in DataProviders?

src/Diagnostics.ModelsAndUtils/DataTableUtility.cs
stevenernst131: I didn't know if this was considered a "ScriptUtility" which is the only utility folder. Should I add it there or a new folder?

src/Diagnostics.ModelsAndUtils/DataTableUtility.cs
ShekharGupta1988: Oh I think you just removed it from csproj but not deleted it. Its in your PR

stevenernst131: You are correct. I have deleted it.

src/Diagnostics.ModelsAndUtils/DataTableUtility.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public static class DataTableUtility
+    {
+        public static DataTable GetDataTable(DataTableResponseObject dataTableResponse)
```
ShekharGupta1988: Instead of creating this as methods, do you think it will nice to create extensions for DataTableResponseObject.

Like "res.ToDataTable()"

src/Diagnostics.ModelsAndUtils/DataTableUtility.cs
ShekharGupta1988: Consider my other comment of creating two extension methods
response.ToDataTable()
dt.ToDataTableResponse()

Perhaps then you dont need a separate file. Just move this class into DataTableResponseObject class file

src/Diagnostics.ModelsAndUtils/DataTableUtility.cs
stevenernst131: I think this would be fine. These specific methods are not supposed to be used by the script authors, as we use it inside kusto data provider to go from DataTableResponseObject => DataTable and then later after the script runs to go the opposite transformation. 

I can create extensions for DataTableResponseObject/DataTable that will do these transformations

src/Diagnostics.ModelsAndUtils/DataTableUtility.cs
stevenernst131: The response.ToDataTable() I am not clear as to what that would be for. I think we need that for changing something like a list of objects into a datatable but I don't believe it is needed for converting between DataTable and DataTableResponseObject, since as I mentioned in the other comment, that does not happen inside the script

PR #15, opened 2018-04-10
Filters Attribute

ShekharGupta1988:
This is work in progress.


src/Diagnostics.ModelsAndUtils/Attributes/Filter.cs
```diff
@@ -0,0 +1,67 @@
...
+namespace Diagnostics.ModelsAndUtils.Attributes
+{
+    public class Filter : Attribute
+    {
+    }
+
+    /// <summary>
+    /// Enum representing Resource type or collection of Resource Types
+    /// </summary>
+    public enum ResourceType
```
stevenernst131: Do we want to split this out to separate resource type and stamp type. Because this could be relative to both sites on ASE and for ASE detectors

ShekharGupta1988: This is still work in progress. I'll be updating these filters in my main checkin. Just want to get the new namespace changes deployed

PR #18, opened 2018-04-17
Adding the GeoMasterDataProvider

puneetg1983:
First pull request for GeoMasterDataProvider. Can cleanup based on some feedback and suggestions but wanted to submit this to check if I am on the right path. Have tested this against a CSM client and a private DiagRole and I believe for this to work on PROD DiagRole, there shouldnt be any changes required.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
+++ b/src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
@@ -16,8 +16,8 @@ namespace Diagnostics.DataProviders
         /// <summary>
         /// GeomasterCertThumbprint
         /// </summary>
-        [ConfigurationName("GeoCertThumbprint")]
-        public string GeoCertThumbprint { get; set; }
+        [ConfigurationName("GeoRegionCertThumbprint")]
+        public string GeoRegionCertThumbprint { get; set; }
 
         /// <summary>
         /// GeomasterEndpoint
--- a/src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
+++ b/src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
@@ -1,0 +1,34 @@
+﻿
+namespace Diagnostics.DataProviders
+{
+    [DataSourceConfiguration("GeoMaster")]
+    public class GeoMasterDataProviderConfiguration : IDataProviderConfiguration
+    {
+        
+        public void PostInitialize()
+        {
+        }
+
+        public GeoMasterDataProviderConfiguration()
+        {
+        }
+
+        /// <summary>
+        /// GeomasterCertThumbprint
+        /// </summary>
+        [ConfigurationName("GeoCertThumbprint")]
+        public string GeoCertThumbprint { get; set; }
+
+        /// <summary>
+        /// GeomasterEndpoint
+        /// </summary>
+        [ConfigurationName("GeoEndpointAddress")]
+        public string GeoEndpointAddress { get; set; }
+
+        /// <summary>
+        /// IsInternal
+        /// </summary>
+        [ConfigurationName("Token")]
+        public string Token { get; set; }
+    }
+}
--- a/src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
+++ b/src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
@@ -16,8 +16,8 @@ namespace Diagnostics.DataProviders
         /// <summary>
         /// GeomasterCertThumbprint
         /// </summary>
-        [ConfigurationName("GeoRegionCertThumbprint")]
-        public string GeoRegionCertThumbprint { get; set; }
+        [ConfigurationName("GeoCertThumbprint")]
+        public string GeoCertThumbprint { get; set; }
 
         /// <summary>
         /// GeomasterEndpoint
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -24,7 +24,7 @@ namespace Diagnostics.DataProviders
         private IGeoMasterClient InitClient()
         {
             IGeoMasterClient geoMasterClient;
-            bool onDiagRole = !string.IsNullOrWhiteSpace(_configuration.GeoCertThumbprint);
+            bool onDiagRole = !string.IsNullOrWhiteSpace(_configuration.GeoRegionCertThumbprint);
             if (onDiagRole)
             {
                 geoMasterClient = new GeoMasterCertClient(_configuration);
@@ -53,12 +53,24 @@ namespace Diagnostics.DataProviders
             return appSettings;
         }
 
-        public async Task<IDictionary<string, dynamic>> GetStickySlotSettingNames(string subscriptionId, string resourceGroupName, string name)
+        public async Task<IDictionary<string, string[]>> GetStickySlotSettingNames(string subscriptionId, string resourceGroupName, string name)
         {
             string path = SitePathUtility.GetSitePath(subscriptionId, resourceGroupName, name);
             path = path + "/config/slotConfigNames";
-            var geoMasterResponse = await HttpGet<GeoMasterResponse>(path);
-            return geoMasterResponse.Properties;
+            GeoMasterResponse geoMasterResponse = null;
+            geoMasterResponse = await HttpGet<GeoMasterResponse>(path);
+            Dictionary<string, string[]> stickyToSlotSettings = new Dictionary<string, string[]>();
+            foreach (var item in geoMasterResponse.Properties)
+            {
+                var val = new string[0];
+                if (item.Value != null && item.Value is Newtonsoft.Json.Linq.JArray)
+                {
+                    var jArray = (Newtonsoft.Json.Linq.JArray)item.Value;
+                    val = jArray.ToObject<string[]>();
+                }
+                stickyToSlotSettings.Add(item.Key, val);
+            }
+            return stickyToSlotSettings;
         }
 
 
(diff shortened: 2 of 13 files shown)


src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public GeoMasterDataProviderConfiguration()
+        {
+        }
+
+        /// <summary>
+        /// GeomasterCertThumbprint
+        /// </summary>
+        [ConfigurationName("GeoRegionCertThumbprint")]
+        public string GeoRegionCertThumbprint { get; set; }
```
hforeste: nitpik: Could you name this GeoMasterProdCertThumbprint instead. This name gives me the impression that you are trying to authenticate with GR instead of GeoMaster.

src/Diagnostics.DataProviders/GeoMasterUtility.cs
```diff
@@ -0,0 +1,161 @@
...
+        public const string Functions = "functions";
+        public const string ResourceProviderName = "Microsoft.Web";
+        public const string NetworkResourceProviderName = "Microsoft.Network";
+        public const string ResourceProviderDisplayName = "Microsoft Web Apps";
+        public const string ResourceGroupSegment = "resourceGroups";
+        public const string ResourceGroupRoot = "subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}";
+        public const string ResourceRoot =
+                "subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/" + ResourceProviderName;
+    }
+    public class DataSourceCancelledException : Exception
```
stevenernst131: Should this be in a seperate file? Is it supposed to be used across any data provider or just geomaster?

If you leave it here please add a empty line between this and the class above

src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
puneetg1983: Actually this comment is interesting. I am just trying to be in exact sync with the RoleConfiguration setting for diagrole so it is easy to relate what is what. Earlier I had it named as GeoMasterCertThumbprint and I was reading this setting from roleconfiguration but then I ended reading the totally wrong certificate and wasted many hours. I am not sure why role configuration has it named this way but I am thinking changing the name now will cause more confusion? Or not?

src/Diagnostics.DataProviders/GeoMasterUtility.cs
puneetg1983: I would like to understand better on how we are handling cancellations and where should we be logging exceptions. Then we can decide if this should be same or different file.

src/Diagnostics.DataProviders/GeoMasterUtility.cs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                return string.Format(FmtSiteWithSlotName, siteName, slotName);
+            }
+
+            return siteName;
+        }
+
+        public static string GenerateHostPrefixWithSlotName(string siteName, string slotName)
+        {
+            if (!string.Equals(slotName, GeoMasterConstants.ProductionSlot, StringComparison.OrdinalIgnoreCase))
```
stevenernst131: what if slot name is empty? Don't know exactly how you are using this.

src/Diagnostics.DataProviders/GeoMasterUtility.cs
hforeste: Actually, can we make this a in a different file like CustomExceptions.cs or something. Even before tackling the logging story this exception would be useful across all the data providers.

src/Diagnostics.DataProviders/GeoMasterUtility.cs
puneetg1983: All that I have is exact copy of existing code in the GeoMasterClient...

stevenernst131: So it looks like this will work now but only because it is currently only called with the output of ParseSiteWithSlotName, which returns 'Production'.  It is probably work checking for an empty slot name, in case someone uses it in the future without passing 'Production' when there is no slot name

src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
stevenernst131: If you look in Microsoft.Web.Hosting.Diagnstics.Common CommonRuntimeConfiguration.cs you will see this:

image

puneetg1983: Ouch...that's a bigger problem then. I will validate and update everywhere

src/Diagnostics.DataProviders/GeoMasterUtility.cs
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+
+    public static class SitePathUtility
+    {
+        public static string GetSitePath(string subscriptionId, string resourceGroupName, string name = null)
+        {
+            var path = GetResourceRootPath(subscriptionId, resourceGroupName) + "/" + GeoMasterConstants.SitesResource;
+
+            if (name != null)
```
ShekharGupta1988: ue string.IsNullOrWhiteSpace

src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
```diff
@@ -0,0 +1,34 @@
...
+        /// GeomasterEndpoint
+        /// </summary>
+        [ConfigurationName("GeoEndpointAddress")]
+        public string GeoEndpointAddress { get; set; }
+
+        /// <summary>
+        /// IsInternal
+        /// </summary>
+        [ConfigurationName("Token")]
+        public string Token { get; set; }
```
ShekharGupta1988: Do you have a flag to control ARM vs Geomaster call for private testing?

src/Diagnostics.DataProviders/GeoMasterTokenClient.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+
+namespace Diagnostics.DataProviders
+{
+    class GeoMasterTokenClient : IGeoMasterClient
```
ShekharGupta1988: Better name - "ArmClient"

src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
ShekharGupta1988: I guess my question how do I switch between Geomasterclient and ArmClient

ShekharGupta1988:
@puneetg1983 - please add a unit test also. It will help catching basic regressions in future. Check out the Diagnostics.Tests project in tests folder. To run the tests, execute "dotnet test"

src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
hforeste: @ShekharGupta1988 , he is checking to see if the GeoCertThumprint is set and if not then he uses GeoMasterTokenClient (ArmClient)

hforeste: https://github.com/Azure/Azure-AppServices-Diagnostics/pull/18/files/446c5c618a47f0d83f2d0712316fb5bf84258d1b#diff-dfb9d8e7d9a336d190de41b69312b210

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -0,0 +1,145 @@
...
+        
+        public GeoMasterDataProvider(OperationDataCache cache, GeoMasterDataProviderConfiguration configuration) : base(cache)
+        {
+            _configuration = configuration;
+            _geoMasterClient = InitClient();
+        }
+
+        private IGeoMasterClient InitClient()
+        {
+            IGeoMasterClient geoMasterClient;
```
hforeste: Create GeoMasterMockClient for testing

src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
puneetg1983: Yup, as Hawk said :-)

src/Diagnostics.DataProviders/GeoMasterUtility.cs
puneetg1983: Created a common folder and moved this file over there...

src/Diagnostics.DataProviders/GeoMasterUtility.cs
puneetg1983: done

src/Diagnostics.DataProviders/GeoMasterUtility.cs
puneetg1983: For now, leaving it as is. We can revisit this if required

src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
puneetg1983: Changed from GeoRegionCertThumbprint to GeoCertThumbprint

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
puneetg1983: I have never written test projects so this is something I would like to understand and post that I can add the right clients. This is something I was planning to discuss next week anyways...

puneetg1983:
For sure @ShekharGupta1988  got all the attention for his PR ;-)

@ShekharGupta1988 , @stevenernst131 , @hforeste requesting review on this.

PR #19, opened 2018-04-18
Introducing Filters : Support for Different App Types, OS Types, ASE, Stacks etc.

ShekharGupta1988:
Added Resource Level Filters which enables the developer to write detectors for a particular resource type (like app or ase) and add more levels of filtering like platform linux or not, stack is php or node , ase verskion v1 or v2 etc.

Examples:

**[AppFilter]**  :  This Filter signifies that detector is for WebApp Resource and Windows Platform (defaults)

**[AppFilter(AppType = AppType.WebApp, PlatformType = PlatformType.Linux, StackType = StackType.PHP)]**   :   This Filter signifies that detector is for php web app on Linux.

**[HostingEnvironmentFilter]**  :  This filter indicates that detector is for ASE (Windows Platform, which is default)

**[HostingEnvironmentFilter(HostingEnvironmentType = HostingEnvironmentType.V2, PlatformType = PlatformType.Linux)]**   :  This filter signifies detector for ASE V2 on Linux


src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/Enumerations.cs
```diff
@@ -0,0 +1,94 @@
...
+        All = 255
+    }
+
+    public enum SkuType
+    {
+        Free = 1,
+        Shared = 2,
+        Standard = 4,
+        Premium = 8,
+        All = 255
```
hforeste: Also add Isolated SKU

data/templates/Detector_HostingEnvironment.csx
```diff
@@ -0,0 +1,19 @@
+private static string GetQuery(OperationContext<HostingEnvironment> cxt)
+{
+    return
+    $@"<YOUR_TABLE_NAME>
```
stevenernst131: This may be an annoying ask, but what do you think about making the template queries in the form of something like

`
let TenantList = {Utilities.getTenantList(cxt.Resource)}
let startTime = {cxt.StartTime}
let endTime = {cxt.EndTime}
Table | where TIMESTAMP > startTime and TIMESTAMP < endTime and Tenant in TenantList
| rest of query

I was thinking about this recently, and it seems to me that this would be easier and more intuitive for the user to transfer from kusto. Also it seems like most other people on the team use let statements for variables in their queries as standard practice.

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/HostingEnvironmentFilter.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        /// <summary>
+        /// Platform Type (Windows, Linux ...)
+        /// </summary>
+        public PlatformType PlatformType;
+
+        /// <summary>
+        /// Hosting Environment Type (ASE V1, ASE V2 ...)
+        /// </summary>
+        public HostingEnvironmentType HostingEnvironmentType;
```
hforeste: I think you can overload StampType here instead of creating a separate type like the HostingEnvironmentType. In the case that someone tries to set this to something that doesn't match ASEv1 or ASEv2 or All we can throw an InvalidOperationException.

src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
```diff
@@ -7,34 +7,48 @@ namespace Diagnostics.ModelsAndUtils.Models
...
+        /// See <see cref="App"/> and <see cref="HostingEnvironment"/>
         /// </summary>
-        public SiteResource Resource;
+        public TResource Resource { get; private set; }
 
         /// <summary>
         /// Start Time(UTC) for data measurement
         /// </summary>
-        public string StartTime;
+        public string StartTime { get; private set; }
```
stevenernst131: Should we do the work of turning the start and end time string into an actual datetime before passing it to the csx? I guess if the helper methods use string then it makes sense, but it seems like if they need to parse the start or end time in csx then it could lead to people doing it wrong.

src/Diagnostics.RuntimeHost/Controllers/HostingEnvironment/HostingEnvironmentController.cs
```diff
@@ -0,0 +1,86 @@
...
+    [Route(UriElements.HostingEnvironmentResource + UriElements.Diagnostics)]
+    public sealed class HostingEnvironmentController : HostingEnvironmentControllerBase
+    {
+        public HostingEnvironmentController(IStampService stampService, ICompilerHostClient compilerHostClient, ISourceWatcherService sourceWatcherService, IInvokerCacheService invokerCache, IDataSourcesConfigurationService dataSourcesConfigService)
+            : base(stampService, compilerHostClient, sourceWatcherService, invokerCache, dataSourcesConfigService)
+        {
+        }
+
+        [HttpPost(UriElements.Query)]
+        public async Task<IActionResult> Post(string subscriptionId, string resourceGroupName, string hostingEnvironmentName, string[] hostNames, string stampName, [FromBody]CompilationBostBody<DiagnosticStampData> jsonBody, string startTime = null, string endTime = null, string timeGrain = null)
```
stevenernst131: CompilationBostBody supposed to be CompilationPostBody?

src/Diagnostics.RuntimeHost/Controllers/ControllerBase.cs
```diff
@@ -1,56 +1,171 @@
...
+        {
+            this.Request.Headers.TryGetValue(HeaderConstants.RequestIdHeaderName, out StringValues requestIds);
+            this.Request.Headers.TryGetValue(HeaderConstants.InternalCallHeaderName, out StringValues internalCallHeader);
+            bool isInternalRequest = false;
+            if (internalCallHeader.Any())
+            {
+                bool.TryParse(internalCallHeader.First(), out isInternalRequest);
+            }
+
+            return new OperationContext<TResource>(
```
stevenernst131: A comment here about the datetime seems more relevant, now I actually see what you are doing. I don't feel super strongly but it seems to me that it makes more sense to keep it in datetime format, and then in utilities that are used in kusto queries, correctly print it for kusto

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
```diff
@@ -0,0 +1,82 @@
...
+            }
+
+            if (string.IsNullOrWhiteSpace(jsonBody.Script))
+            {
+                return BadRequest("Missing script in body");
+            }
+
+            if(jsonBody.Resource == null)
+            {
+                return BadRequest("Missing Site data in body");
```
stevenernst131: This worries me, I think it is possible that we want to be able to call this without the site data in the body. For example, I was trying to test support center with these APIs, which would enable me to test generic detectors without access to ARM for that resource. However, if you enforce this, that would not be possible.

src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
```diff
@@ -7,34 +7,48 @@ namespace Diagnostics.ModelsAndUtils.Models
...
+        /// <summary>
+        /// Request Id
+        /// </summary>
+        public string RequestId { get; private set; }
 
         /// <summary>
         /// TimeGrain in minutes for aggregating data.
         /// </summary>
-        public string TimeGrain;
+        public string TimeGrain { get; private set; }
```
hforeste: Don't have to add now but in case you want to, we can also add UserAgent here for the near future to distinguish those on the cli tool, applens, prod

src/Diagnostics.RuntimeHost/Services/ResourceService/SiteService.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                return StackType.None;
+            }
+
+            return GetAppStackType(stackTable.Rows[0][0].ToString().ToLower());
+        }
+
+        private StackType GetAppStackType(string stackString)
+        {
+            switch (stackString)
```
stevenernst131: Nice

src/Diagnostics.ModelsAndUtils/Models/Resource/App.cs
```diff
@@ -0,0 +1,78 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.Models
+{
+    /// <summary>
+    /// Resource representing an app service
+    /// </summary>
+    public class App : AppFilter, IResource
```
hforeste: So potentially, since App is inheriting from AppFilter, App class can be used as an attribute right?

src/Diagnostics.Scripts/EntityInvoker.cs
```diff
@@ -125,9 +127,8 @@ public void InitializeEntryPoint(Assembly asm)
             }
             catch (Exception ex)
             {
-                if (ex is ScriptCompilationException)
+                if (ex is ScriptCompilationException scriptEx)
```
stevenernst131: Why not just do
```
catch(ScriptCompilationException scriptEx) {
   ...
}
```

Wouldn't that accomplish the same thing?

src/Diagnostics.ModelsAndUtils/Models/Resource/HostingEnvironment.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        public HostingEnvironment(string subscriptionId, string resourceGroup, string name)
+        {
+            this.SubscriptionId = subscriptionId;
+            this.ResourceGroup = resourceGroup;
+            this.Name = name;
+            this.TenantIdList = new List<string>();
+        }
+
+        public bool IsApplicable(IResourceFilter filter)
```
hforeste: Do we need to make this public? Will this be useful to applens developers?

src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
stevenernst131: I made another comment about this in a more relevant place below so look for that one

src/Diagnostics.ModelsAndUtils/Models/Resource/IResource.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// Type of Resource
+        /// </summary>
+        ResourceType ResourceType { get; set; }
+        
+        /// <summary>
+        /// Determines whether the resource is applicable after filtering.
+        /// </summary>
+        /// <param name="filter">Resource Filter</param>
+        /// <returns>True, if resource passes the filter. False otherwise</returns>
+        bool IsApplicable(IResourceFilter filter);
```
hforeste: Same as comment above. I'm not sure if I see developers creating filters to see if the resource is applicable.

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Utilities.cs
```diff
@@ -11,75 +11,85 @@ public class Utilities
...
         {
-            return $"{timeColumnName} >= datetime({cxt.StartTime}) and {timeColumnName} <= datetime({cxt.EndTime})";
+            return $"{timeColumnName} >= datetime({startTime}) and {timeColumnName} <= datetime({endTime})";
         }
 
         /// <summary>
         /// Creates a kusto query part for Tenant filters. This is a recommended way to use Tenant filters as this takes care of mega-stamps also.
         /// </summary>
-        /// <param name="cxt"><see cref="OperationContext"/></param>
+        /// <param name="cxt"><see cref="OperationContext{TResource}"/></param>
```
hforeste: Remove this to match IResource

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Utilities.cs
```diff
@@ -11,75 +11,85 @@ public class Utilities
...
+            }
+
+            return $"Tenant in ({string.Join(",", tenantIds.Select(t => $@"""{t}"""))})";
         }
 
         /// <summary>
         /// Creates a kusto query part for Time and Tenant filter together.
         /// </summary>
-        /// <param name="cxt"><see cref="OperationContext"/></param>
+        /// <param name="cxt"><see cref="OperationContext{TResource}"/></param>
```
hforeste: Same here

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Utilities.cs
```diff
@@ -11,75 +11,85 @@ public class Utilities
...
         {
-            return $"{TimeFilterQuery(cxt, timeColumnName)} and {TenantFilterQuery(cxt)}";
+            return $"{TimeFilterQuery(startTime, endTime, timeColumnName)} and {TenantFilterQuery(resource)}";
         }
 
         /// <summary>
         /// Creates kusto query part to cover hostnames. Takes care for wildcard hostnames also.
         /// </summary>
-        /// <param name="cxt"><see cref="OperationContext"/></param>
+        /// <param name="cxt"><see cref="OperationContext{App}"/></param>
```
hforeste: Change to match methods params

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
ShekharGupta1988: the next change would be to fetch the data if its not present. I'll use Hawk's work when that gets checked in.

src/Diagnostics.Scripts/EntityInvoker.cs
ShekharGupta1988: ya, I was planning to add more custom exception types like invocationexception to pass through the invocation errors a part of the api

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Utilities.cs
ShekharGupta1988: I need to correct the comments. I'll work on the documentation later.

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Utilities.cs
ShekharGupta1988: same as above.

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/Enumerations.cs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    public enum StampType
+    {
+        Public = 1,
+        ASEV1 = 2,
+        ASEV2 = 4,
+        All = 255
+    }
+}
```
puneetg1983: Megastamp?

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/ResourceFilterBase.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// </summary>
+        public bool InternalOnly { get; set; }
+        
+        public ResourceFilterBase(ResourceType resourceType, bool internalOnly)
+        {
+            this.ResourceType = resourceType;
+            this.InternalOnly = internalOnly;
+        }
+    }
+}
```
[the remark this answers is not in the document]
puneetg1983: Never mind, just found `IsInternalCall `

src/Diagnostics.ModelsAndUtils/Models/Resource/HostingEnvironment.cs
```diff
@@ -0,0 +1,92 @@
...
+        public string ResourceGroup { get; set; }
+
+        /// <summary>
+        /// Internal Name (For Example:- waws-prod-....)
+        /// </summary>
+        public string InternalName { get; set; }
+
+        /// <summary>
+        /// Service Address
+        /// </summary>
```
puneetg1983: Add VNET name for ASE ?

hforeste: I can worry about adding additional things for ASE, this filter work is a big work item alone. Lets just get this out there.

src/Diagnostics.RuntimeHost/Controllers/HostingEnvironment/HostingEnvironmentController.cs
ShekharGupta1988: lol .. will fix

stevenernst131: Aspaan comes back to haunt us!

src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
ShekharGupta1988: UserAgent in Operation Context? we should add User Agents in calls made by our backend, right.

src/Diagnostics.ModelsAndUtils/Models/Resource/HostingEnvironment.cs
ShekharGupta1988: good point. I can mark it internal

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/Enumerations.cs
ShekharGupta1988: you see a scenario applicable only for megastamp. I can add that.

src/Diagnostics.ModelsAndUtils/Models/Resource/HostingEnvironment.cs
ShekharGupta1988: ya :)

data/templates/Detector_HostingEnvironment.csx
ShekharGupta1988: see these improvements can be done separately also. Doesnt have to be part of filters work.

src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
hforeste: lets table this for discussion later since its not critical to this change

src/Diagnostics.ModelsAndUtils/Models/Resource/HostingEnvironment.cs
puneetg1983: Well we have the vnet name at least...other detectors can get this and get vnet info through the geoMaster provider I guess...

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/Enumerations.cs
puneetg1983: Actually, there is also a SKU called PremiumV2

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/Enumerations.cs
puneetg1983: I am trying to think why this is needed and now I am realizing most of the times it is for people when forming KUSTO queries so it may not be that needed. Actually rather than an Enum type, it would make sense to have a method that one can call IsMegastamp? I cannot think of real world scenario but sometimes we have specific issues only on Megstamps. Anyways not a high priority.

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/Enumerations.cs
ShekharGupta1988: good point. I'll add those 2

src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
ShekharGupta1988: Good point. I will include that. That will lead to few changes down the pipe till data providers, but i guess right time to do this change

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/ResourceFilterBase.cs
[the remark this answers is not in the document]
ShekharGupta1988: yes. The IsInternalCall is part of the context getting passed to script

PR #22, opened 2018-04-23
Support for Linux Apps

ShekharGupta1988:
Linux Workers emit info into LinuxRoleInstanceHeartBeats instead of RoleInstanceHeartbeats. So fixing that.

The approach is to look in both the tables and combine the tenant-ids.


src/Diagnostics.RuntimeHost/Controllers/ControllerBase.cs
```diff
@@ -126,11 +126,11 @@ protected OperationContext<TResource> PrepareContext<TResource>(TResource resour
             return Ok(queryRes);
         }
 
-        protected async Task<HostingEnvironment> GetHostingEnvironment(string subscriptionId, string resoourceGroup, string name, DiagnosticStampData stampPostBody, DateTime startTime, DateTime endTime)
+        protected async Task<HostingEnvironment> GetHostingEnvironment(string subscriptionId, string resourceGroup, string name, DiagnosticStampData stampPostBody, DateTime startTime, DateTime endTime)
```
hforeste: nitpik: Move to HostingEnvironmentControllerBase class

ShekharGupta1988: naa. This is needed by SiteController as well.

src/Diagnostics.RuntimeHost/Services/ResourceService/StampService.cs
```diff
@@ -13,56 +14,88 @@ namespace Diagnostics.RuntimeHost.Services
...
+            List<string> linuxTenantIds = new List<string>();
 
-            string query =
-                $@"RoleInstanceHeartbeat
-                | where TIMESTAMP >= datetime({startTimeStr}) and TIMESTAMP <= datetime({endTimeStr}) and PublicHost startswith ""{stamp}""
-                | summarize by Tenant, PublicHost";
+            string windowsQuery = GetTenantIdQuery(stamp, startTime, endTime, PlatformType.Windows);
+            string linuxQuery = GetTenantIdQuery(stamp, startTime, endTime, PlatformType.Linux);
+
+            var windowsTask = _dataProviders.Kusto.ExecuteQuery(windowsQuery, stamp, requestId, "GetTenantIdForStamp-Windows");
```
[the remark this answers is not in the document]
hforeste: Are you going to add `GetTenantIdForStamp-Windows` in the DataProviderConstants class?

src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
```diff
@@ -25,20 +25,20 @@ public SiteControllerBase(IStampService stampService, ISiteService siteService,
...
+            if (this.Request.Headers.TryGetValue(HeaderConstants.RequestIdHeaderName, out StringValues requestIds))
             {
                 requestId = requestIds.FirstOrDefault() ?? string.Empty;
             }
 
             App app = new App(subscriptionId, resourceGroup, appName)
             {
                 DefaultHostName = postBody.DefaultHostName,
-                Hostnames = postBody.HostNames.Select(p => p.Name),
+                Hostnames = postBody.HostNames != null ? postBody.HostNames.Select(p => p.Name) : new List<string>(),
```
stevenernst131: We really need to figure out how to return the kusto query that is run, because if for some reason this is null, it could cause a lot of confusion if there is no data coming back.

PR #23, opened 2018-04-25
Add compilation check for illegal file path chars in Id

ShekharGupta1988:
right now checking Id against empty and illegal file name characters allowed


src/Diagnostics.Scripts/EntityInvoker.cs
```diff
@@ -225,6 +227,27 @@ private ScriptOptions GetScriptOptions(ImmutableArray<string> frameworkReference
...
+        {
+            if(this._entryPointDefinitionAttribute != null)
+            {
+                string id = this._entryPointDefinitionAttribute.Id;
+                if (string.IsNullOrWhiteSpace(id))
+                {
+                    throw new ScriptCompilationException("Id cannot be empty in Definition attribute");
+                }
+
+                List<char> invalidChars = Path.GetInvalidFileNameChars().ToList();
```
hforeste: >Path.GetInvalidFileNameChars() [](start = 42, length = 30)

Didn't know this exist. Awesome.:blush:

tests/ScriptsTests/EntityInvokerTests.cs
```diff
@@ -177,5 +177,22 @@ public class EntityInvokerTests
                 });
             }
         }
+
+        [Theory]
+        [InlineData("")]
+        [InlineData("<someId>")]
```
hforeste: Nice. Didn't know xUnit offers this. I was just reading about it.

PR #24, opened 2018-04-26
Adding VNET Verifier and deployments list to GeoMaster DataProvider

puneetg1983:


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -73,6 +73,27 @@ private IGeoMasterClient InitClient()
             return stickyToSlotSettings;
         }
 
+        public async Task<VnetValidationFailureDetails> VerifyHostingEnvironmentVnet(string subscriptionId, string vnetResourceGroup, string vnetName, string vnetSubnetName, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var path = string.Format(@"subscriptions/{0}/providers/Microsoft.Web/verifyHostingEnvironmentVnet", subscriptionId);
```
hforeste: Do you want to use the SitePathUtility? Just to keep things consistent.

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -73,6 +73,27 @@ private IGeoMasterClient InitClient()
...
+            var vnetParameters = new VnetParameters { VnetResourceGroup = vnetResourceGroup, VnetName = vnetName, VnetSubnetName = vnetSubnetName };
+            var result = await HttpPost<VnetValidationFailureDetails, VnetParameters>(path, vnetParameters, "", GeoMasterConstants.March2016Version, cancellationToken);
+            return result;
+        }
+
+        public async Task<List<IDictionary<string, dynamic>>> GetDeployments(string subscriptionId, string resourceGroupName, string name)
```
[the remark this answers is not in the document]
ShekharGupta1988: got it. 
An orthogonal question : Are we adding specific methods in geomaster provider and eventually add a pas through ability also?

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/VnetParameters.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.Models
+{
+    public class VnetParameters
```
ShekharGupta1988: I see this is an input model is used only by geomaster data provider
Is this also getting used in csx scripts?

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/GeoMasterResponse.cs
```diff
@@ -14,4 +14,9 @@ public class GeoMasterResponse
         public IDictionary<string, dynamic> Properties { get; set; }
     }
 
+    public class GeoMasterResponseValue
```
ShekharGupta1988: same here. Is this class getting used by scripts or just the data provider

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/VnetParameters.cs
hforeste: No. GeoMasterDataProvider just takes exact string parameters for subscriptionid, vnet subnet name and etc,
@puneetg1983 , you should probably change the access modifier for this model to internal so as to not expose this to developers creating their csx file.

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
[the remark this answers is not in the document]
hforeste: @puneetg1983  and I was talking about that. We can chat offline about this.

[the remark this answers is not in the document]
puneetg1983: This is actually temporary for now. We might come up with a generic pass through method in future that allows any to call a GeoMaster route passing the POST body etc. if required. Some of the these methods *may* be used a lot more often hence I created them.

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/VnetParameters.cs
puneetg1983: This one I can make internal.

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/GeoMasterResponse.cs
puneetg1983: Once we add a pass through kind of method, the detectors may benefit directly with this class so I am thinking of leaving this one as is

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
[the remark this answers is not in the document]
puneetg1983: Changed this method to GetAppDeployments

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
puneetg1983: I don't believe that function can be used directly here as this path has no site name in it...

PR #27, opened 2018-05-02
Markdown and Detector List rendering

stevenernst131:
Adding a markdown rendering type.
Also adding list of detectors rendering type, that just returns the list
of ids that should be displayed. In order to do this, I had to add a
status as part of the detector response

NOTE: For detector list, I think I will need to iterate on this more once I see what the feedback is.


stevenernst131:
I have intentionally not added documentation for detector list because I want to work with a few people first before making it discoverable.

ShekharGupta1988:
Not sure if you guys discussed this, people would have ri explicitly set detector status now while writing detectors ? I was thinking it can be automatically deduced by the insights present in the detector.

stevenernst131:
That is an interesting idea. I think it makes sense to do that and then also allow it to be overridden. One other thing I added last minute was the ability to put a summary status message, which would really be needed to show a message in a view that shows detectors in a list format with the status marks. We should probably discuss more, the reason I wasn't going to publicize this is that I just wanted to get this change in so I could try out some UI designs to see what works. 

I can add some logic to deduce the status from the insights, and we can build off that into a more complete solution.

stevenernst131:
I am looking into making the change that I discussed above, but it is occurring to me that it makes sense for there to be a single enum for level (critical, warning, etc) that is used by the insight as well as the detector response. This however would break the model and cause us to need to recompile detectors. Do you think that is worth it?

hforeste:
Hey @stevenernst131, I agree that it does make sense to have a single enum but for now don't worry about it. In the mean time, I think we should get in a room and discuss versioning. It feels like we can't push this any further because as new ideas come out and as we refactor the backend code, its ivetiable that we will break detectors especially because the project is so young.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorView.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.Models.ResponseExtensions
+{
+    public static class ResponseDetectorViewExtensions
+    {
+        public static DiagnosticData AddDetectorView(this Response response, List<string> detectorIds)
```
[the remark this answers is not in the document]
stevenernst131: Sure

src/Diagnostics.ModelsAndUtils/Models/Rendering.cs
```diff
@@ -24,6 +24,17 @@ public Rendering(RenderingType type)
         }
     }
 
+    public class DetectorRendering : Rendering
```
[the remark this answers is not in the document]
stevenernst131: Yeah that sounds better, I forgot I was going to make it a list when I named it.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Markdown.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// {
+        ///     var markdown = @"
+        ///     ## This is header
+        ///     ";
+        ///     
+        ///     res.AddMarkdownView(markdown, "Title of markdown section");
+        /// }
+        /// </code>
+        /// </example>
+        public static DiagnosticData AddMarkdownView(this Response response, string markdown, string title = null)
```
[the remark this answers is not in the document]
hforeste: Maybe your right. I don't think its a big deal if we don't have AddTextBoxView, we can just go with this.

I just figured we should make this experience as fluid as possible. I can imagine a sample that says `AddTextBoxView` and it just intuitively makes sense on how to use it.

Though markdown is easy to learn, the lesser things to learn the better. Without intellisense they are already learning much.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorCollection.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.Models.ResponseExtensions
+{
+    public static class ResponseDetectorViewExtensions
+    {
+        public static DiagnosticData AddDetectorCollection(this Response response, List<string> detectorIds)
```
ShekharGupta1988: How does this work in the UI. Do you show a loading message for every detector? 
If yes, Where does the loading message comes from? It cant be from detector response. So it has to be defined here. 
Or are you using the name of the detector?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorStatus.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <example> 
+        /// This sample shows how to use <see cref="SetDetectorStatus"/> method.
+        /// <code>
+        /// public async static Task<![CDATA[<Response>]]> Run(DataProviders dp, OperationContext cxt, Response res)
+        /// {
+        ///     res.SetDetectorStatus(DetectorStatus.Critical, "This is an error message that will come from the detector");
+        /// }
+        /// </code>
+        /// </example>
+        public static void SetDetectorStatus(this Response response, DetectorStatus health, string message = null)
```
ShekharGupta1988: nitpick : probably name the "health" parameter as "status" as the method says "SetDetectorStatus"

Also see my other comment of trying to deduce the status automatically **if the developer has not set any detector status explicitly**.

**Any Insight with Level Critical** : Detector Status is Critical, Message = "<Detector Name> found issues"
etc..

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorCollection.cs
stevenernst131: I had a more detailed conversation with Hawk last week, I can share with you what my vision is as well.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorStatus.cs
```diff
@@ -21,13 +23,48 @@ public static class ResponseDetectorStatusExtensions
...
+                            var insightStatusInt = (int)Enum.Parse(typeof(InsightStatus), row["Status"].ToString());
+                            if (lowestStatus > insightStatusInt)
+                            {
+                                lowestStatus = insightStatusInt;
+                            }
+                        }
+                        return lowestStatus;
+                    }
+                    
+                }).OrderBy(st => st).FirstOrDefault();
```
hforeste: Since you reordered the enum values for InsightStatus then this Ordery and FirstOrDefault should return back the most critical status (no pun intended) of the insights right?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorStatus.cs
```diff
@@ -21,13 +23,48 @@ public static class ResponseDetectorStatusExtensions
...
-                StatusId = health,
+                StatusId = status,
                 Message = message
             };
         }
+
+        public static void UpdateDetectorStatusFromInsights(this Response response)
+        {
+
+            if (response.Status != null)
```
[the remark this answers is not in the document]
stevenernst131: No if this is not manually set by the csx file, then I want to set it.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorStatus.cs
stevenernst131: That is the intent yes, the status of the detector will be most critical of its insights.

stevenernst131:
@ShekharGupta1988 @hforeste let me know if there are any more concerns about this. Otherwise please approve. If you guys are merging shekhars change, please also merge this one.

PR #31, opened 2018-05-14
Add data provider logging

hforeste:


hforeste:
I see a test failure, looking into it

ShekharGupta1988:
@hforeste - until you get the versioning story, can you think of any other way to roll this out to not cause downtime or minimize the impact. Note that the detectors are now public and any exception would be customer facing. Also the weekly SLA for Microsoft.Web will be affected.

Here is one approach you can take:
Before deploying your changes to PROD:
1.  Compile all the detectors and upload them to a separate branch (say "master2.0").
2.  In the prod definitions, update the Github branch to be "master2.0".

So when an UD gets upgraded, it will be pointing to "master2.0" with your backend changes. Hence, it should work.
For the UDs which are not upgraded, they will still be pointing to "master" with old backend changes, so they will continue to work.

hforeste:
Hey @ShekharGupta1988 , I love that idea. Lets do it!

ShekharGupta1988:
One other thing I forgot to mention: Once msftintdm3 roll out finishes, update the appsetting in applens also to point to "master2.0" for publishing new detectors to this branch.

src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        private async Task<T> MakeDependencyCall<T>(string requestId, Task<T> dataProviderTask, [CallerMemberName]string dataProviderOperation = "")
+        {
+            Exception dataProviderException = null;
+            DateTime startTime = DateTime.UtcNow, endTime;
+            try
+            {
+                return await dataProviderTask;
+            }catch(Exception ex)
```
ShekharGupta1988: nit : fix spacing

src/Diagnostics.DataProviders/Diagnostics.DataProviders.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
```
stevenernst131: Did you mean to change this?

ShekharGupta1988: intentional?

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -151,6 +151,38 @@ public void LogCompilerHostClientWarning(string RequestId, string Source, string
         #endregion
 
         #region Data Provider Events (ID Range : 3000 - 3999)
+        [Event(3000, Level = EventLevel.Informational, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogDataProviderMessage)]
+        public void LogDataProviderMessage(string RequestId, string Source, string message)
```
ShekharGupta1988: Rename the "message" argument to be  "Message". Otherwise this will result it one additional column in kusto

src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Data;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Diagnostics.Logger;
+using Diagnostics.ModelsAndUtils.Models;
+
+namespace Diagnostics.DataProviders
+{
+    internal class DataProviderLogDecorator : IKustoDataProvider, IGeoMasterDataProvider, ISupportObserverDataProvider
```
ShekharGupta1988: Do you forsee this class to explode as it has methods from all the data providers?

hforeste: Since kusto and observer are following that parameterized approach (Kusto.ExecuteQuery and Observer.GetResource), I don't see this class blowing up in the near future. Hopefully geomaster will follow the same approach.

src/Diagnostics.DataProviders/Diagnostics.DataProviders.csproj
hforeste: I answered this in steve's comment above

src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
```diff
@@ -0,0 +1,114 @@
...
+            return await MakeDependencyCall(null, _geomasterDataProvider.GetStickySlotSettingNames(subscriptionId, resourceGroupName, name));
+        }
+
+        public async Task<VnetValidationRespone> VerifyHostingEnvironmentVnet(string subscriptionId, string vnetResourceGroup, string vnetName, string vnetSubnetName, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return await MakeDependencyCall(null, _geomasterDataProvider.VerifyHostingEnvironmentVnet(subscriptionId, vnetResourceGroup, vnetName, vnetSubnetName, cancellationToken));
+        }
+
+        private async Task<T> MakeDependencyCall<T>(string requestId, Task<T> dataProviderTask, [CallerMemberName]string dataProviderOperation = "")
+        {
```
stevenernst131: Do you see a problem here that we will not be able to measure how individual kusto queries will perform? Basically `dataProviderOperation` this is going to be `ExecuteQuery` every time. I am not sure what a good solution to this is but it seems like we will want greater granularity in terms of how specific queries are doing.

hforeste: I thought about this. But just to make sure we're on the same page. Who would be the stakeholders who are interested in this data? Would it be us the platform developers or the developers creating detectors?

PR #32, opened 2018-05-15
Introducing SupportTopic for detectors

ShekharGupta1988:
This will help detector owners to enable their detector in case submission flow.

image


ShekharGupta1988:
One more check needs to happen :- if the detector is not public and someone is using support topic, we should give a warning.

src/Diagnostics.Scripts/EntityInvoker.cs
```diff
@@ -251,6 +256,26 @@ private void Validate()
...
                 }
+                
+                // Validate Empty author
+                if (string.IsNullOrWhiteSpace(this._entryPointDefinitionAttribute.Author))
+                {
+                    this.CompilationOutput.Concat(new List<string>() { "Warning : Author not specified in Definition attribute" });
+                }
+
+                // Validate Support Topic Attributes
+                this._entryPointDefinitionAttribute.SupportTopicList.ToList().ForEach(item =>
```
stevenernst131: Could SupportTopicList be null here?

src/Diagnostics.Scripts/EntityInvoker.cs
```diff
@@ -251,6 +256,26 @@ private void Validate()
                 {
                     throw new ScriptCompilationException("Name cannot be empty in Definition attribute");
                 }
+                
+                // Validate Empty author
+                if (string.IsNullOrWhiteSpace(this._entryPointDefinitionAttribute.Author))
+                {
+                    this.CompilationOutput.Concat(new List<string>() { "Warning : Author not specified in Definition attribute" });
```
hforeste: This would be a breaking change but it may be helpful that we make this mandatory. This would be particularly useful when people create a detector with a support topic id that is already used by an existing detector. When they use the information from the complication output to look at the existing detector who is using the same support topic id then they can also see the author to possibly ping them before changing the existing detector

PR #34, opened 2018-05-22
Add SupportObserver APIs for Migration and Fix Bug in KustoClient

hforeste:
-Unblock migration detector creation by adding support observer APIs
-Fix for issue https://github.com/ShekharGupta1988/AppLensV2/issues/1135

### DIFF
--- a/src/Diagnostics.DataProviders/KustoClient.cs
+++ b/src/Diagnostics.DataProviders/KustoClient.cs
@@ -79,7 +79,14 @@ namespace Diagnostics.DataProviders
 
             DataTableResponseObjectCollection dataSet = JsonConvert.DeserializeObject<DataTableResponseObjectCollection>(content);
 
-            return dataSet.Tables.FirstOrDefault().ToDataTable();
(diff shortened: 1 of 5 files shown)


src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -79,7 +79,14 @@ public KustoClient(KustoDataProviderConfiguration configuration)
 
             DataTableResponseObjectCollection dataSet = JsonConvert.DeserializeObject<DataTableResponseObjectCollection>(content);
 
-            return dataSet.Tables.FirstOrDefault().ToDataTable();
+            if (dataSet.Tables == null)
```
ShekharGupta1988: nitpick : maybe just check for dataSet null and also dataSet.Tables.Any() to make it more solid and less error prone.

hforeste: Doing `dataSet.Tables.Any()` wont work in this case because it will throw NullReferenceException.

PR #37, opened 2018-06-02
Fix GetTenantCall in StampService

ShekharGupta1988:
Assumption : All the Secondary stamps have just **One letter(a-z)** after the main stamp name.  so it was also remove dr stamps from the list.

Did some research on the existing data  in RoleInstance heartbeat table and so far the assumption looks correct.


stevenernst131:
Good fix, should have done this from the start! (my fault I remember writing the current implementation 👍  )

PR #38, opened 2018-06-05
Adding support to call the site root path in GeoMasterDataProvider

puneetg1983:
This will allow detectors to get properties that are defined at the site root level and not in observer and looks like there are a lot of them which will benefit from this change.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -284,7 +284,10 @@ namespace Diagnostics.DataProviders
         /// </example>
         public async Task<T> MakeHttpGetRequest<T>(string subscriptionId, string resourceGroupName, string name, string path = "")
         {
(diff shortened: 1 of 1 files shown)


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -263,20 +263,37 @@ private IGeoMasterClient InitClient()
...
+        ///     }
         /// }
         /// </code>
         /// </example>
-        public async Task<T> MakeHttpGetRequest<T>(string subscriptionId, string resourceGroupName, string name, string path)
+        public async Task<T> MakeHttpGetRequest<T>(string subscriptionId, string resourceGroupName, string name, string path = "")
         {
-            path = path.StartsWith("/") ? path.Substring(1): path;
-            path = $"{SitePathUtility.GetSitePath(subscriptionId, resourceGroupName, name)}/{path}";
+            path = path.StartsWith("/") ? path.Substring(1) : path;
```
ShekharGupta1988: this will break if path is null

PR #39, opened 2018-06-07
Insights Take 1

stevenernst131:
This is the first draft of the insights change, I want to get eyes on it ASAP.

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
@@ -8,6 +8,19 @@ using System.Text;
 
 namespace Diagnostics.ModelsAndUtils.Models.GeoMaster
 {
+    public class AzureSupportCenterInsightEnvelope
+    {
+        public int TotalInsightsFound { get; set; }
+
+        public string ErrorMessage { get; set; }
+
+        public Guid CorrelationId { get; set; }
+
+        public string Metadata1 { get; set; }
+
+        public IEnumerable<AzureSupportCenterInsight> Insights { get; set; }
+    }
+
     public class AzureSupportCenterInsight
     {
         public Guid Id { get; set; }
@@ -30,14 +43,13 @@ namespace Diagnostics.ModelsAndUtils.Models.GeoMaster
 
         public IEnumerable<Link> Links { get; set; }
 
-        public string ConfidenceLevel { get; set; }
+        public InsightConfidenceLevel ConfidenceLevel { get; set; }
 
-        public string Scope { get; set; }
-    }
+        public InsightScope Scope { get; set; }
 
-    public enum ImportanceLevel
-    {
-        SomeLevel
+        public IEnumerable<Link> ApplensLinks { get; set; }
+
+        public IEnumerable<KeyValuePair<string, string>> AdditionalDetails { get; set; }
     }
 
     public class CustomerReadyContent
@@ -70,95 +82,34 @@ namespace Diagnostics.ModelsAndUtils.Models.GeoMaster
         public Text Text { get; set; }
     }
 
-    public static class AzureSupportCenterInsightUtilites
+    public enum ImportanceLevel
     {
-        public static AzureSupportCenterInsight CreateInsight<TResource>(Insight insight, OperationContext<TResource> context, Definition detector)
-            where TResource : IResource
-        {
+        Critical = 0,
 
-            // Logic in line before would have to be updated if we added more resource types. 
-            // TODO: May make sense to have resource type enums contain the resource type string
-            var resourceTypeString = context.Resource.ResourceType == ResourceType.App ? "sites" : "hostingEnvironments";
-            var applensPath = $"subscriptions/{context.Resource.SubscriptionId}/resourceGroups/{context.Resource.ResourceGroup}/{resourceTypeString}/{context.Resource.Name}/detectors/{detector.Id}";
+        Warning = 1,
 
-            var recommendedAction = GetTextObjectFromData("recommended action", insight.Body);
-            var description = GetTextObjectFromData("description", insight.Body);
-            var customerReadyContent = GetTextObjectFromData("customer ready content", insight.Body);
+        Info = 2,
+    }
 
-            var supportCenterInsight = new AzureSupportCenterInsight()
-            {
-                Id = GetDetectorGuid(detector.Id),
-                Title = insight.Message,
-                ImportanceLevel = ImportanceLevel.SomeLevel,
-                InsightFriendlyName = insight.Message,
-                IssueCategory = detector.Name.ToUpper(),
-                IssueSubCategory = detector.Name,
-                Description = description,
-                Links = new List<Link>()
-                    {
-                        new Link()
-                        {
-                            Type = 0,
-                            Text = "Applens Link",
-                            Uri = $"https://applens.azurewebsites.net/{applensPath}"
-                        }
-                    },
-                RecommendedAction = recommendedAction == null ? null : 
-                    new RecommendedAction()
-                    {
-                        Id = new Guid(),
-                        Text = recommendedAction
-                    },
-                CustomerReadyContent = customerReadyContent == null ? null :
-                    new CustomerReadyContent()
-                    {
-                        Text = customerReadyContent
-                    },
-                ConfidenceLevel = "High",
-                Scope = "ResourceLevel"
-            };
+    public enum InsightScope
+    {
+        ResourceLevel = 0,
 
-            return supportCenterInsight;
-        }
+        SubscriptionLevel = 1,
+    }
 
-        private static Guid GetDetectorGuid(string detector)
-        {
-            try
-            {
-                Encoding utf8 = Encoding.UTF8;
-                int count = utf8.GetByteCount(detector);
-                byte[] bytes = new byte[count > 16 ? count : 16];
-                Encoding.UTF8.GetBytes(detector, 0, detector.Length, bytes, 0);
-                // Guid only takes byte array of length 16
-                return new Guid(bytes.Take(16).ToArray());
-            }
-            catch(Exception ex)
-            {
-                Console.WriteLine(ex.ToString());
-            }
-            return new Guid();
-        }
 
-        // Returns true if markdown
-        private static Text GetTextObjectFromData(string searchKey, Dictionary<string, string> data)
-        {
-            var matchingNameValuePairs = data.Where(x => x.Key.ToLower() == searchKey);
-            var matchingString = matchingNameValuePairs.Any() ? matchingNameValuePairs.FirstOrDefault().Value : null;
 
-            if (string.IsNullOrWhiteSpace(matchingString))
-            {
-                return null;
-            }
+    public enum InsightConfidenceLevel
+    {
+        VeryLow = 0,
 
-            var isMarkdown = matchingString.Contains("<markdown>");
+        Low = 1,
 
-            var output = new Text()
-            {
-                IsMarkdown = isMarkdown,
-                Value = isMarkdown ? matchingString.Replace("<markdown>", "").Replace("</markdown>", "") : matchingString
-            };
+        Medium = 2,
 
-            return output;
-        }
+        High = 3,
+
+        VeryHigh = 4,
     }
 }
--- a/src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
@@ -1,0 +1,164 @@
+﻿using Diagnostics.ModelsAndUtils.Attributes;
+using Diagnostics.ModelsAndUtils.Models.ResponseExtensions;
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Linq;
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.Models.GeoMaster
+{
+    public class AzureSupportCenterInsight
+    {
+        public Guid Id { get; set; }
+
+        public string Title { get; set; }
+
+        public ImportanceLevel ImportanceLevel { get; set; }
+
+        public string InsightFriendlyName { get; set; }
+
+        public string IssueCategory { get; set; }
+
+        public string IssueSubCategory { get; set; }
+
+        public Text Description { get; set; }
+
+        public RecommendedAction RecommendedAction { get; set; }
+
+        public CustomerReadyContent CustomerReadyContent { get; set; }
+
+        public IEnumerable<Link> Links { get; set; }
+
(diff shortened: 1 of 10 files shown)


src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
```diff
@@ -76,7 +80,50 @@ public SitesController(IStampService stampService, ISiteService siteService, ICo
...
+
+        [HttpPost(UriElements.Insights)]
+        public async Task<IActionResult> GetInsights(string subscriptionId, string resourceGroupName, string siteName, [FromBody] DiagnosticSiteData postBody, string supportTopicId, string startTime = null, string endTime = null, string timeGrain = null)
+        {
+            if (postBody == null)
+            {
+                return BadRequest("Post Body missing.");
+            }
+
+            if (siteName == "steve")
```
ShekharGupta1988: Forgot to remove this?  I guess this was for testing purpose

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
```diff
@@ -76,7 +80,50 @@ public SitesController(IStampService stampService, ISiteService siteService, ICo
...
+                    new DiagnosticHostnameData()
+                    {
+                        Name = "nodebakery.azurewebsites.net",
+                        Type = 0
+                    }
+                };
+            }
+
+            DateTimeHelper.PrepareStartEndTimeWithTimeGrain(string.Empty, string.Empty, string.Empty, out DateTime startTimeUtc, out DateTime endTimeUtc, out TimeSpan timeGrainTimeSpan, out string errorMessage);
+            App app = await GetAppResource(subscriptionId, resourceGroupName, siteName, postBody, startTimeUtc, endTimeUtc);
```
ShekharGupta1988: Shouldn't you move this detector to Base and have this Controller inherit that.  Check any other API above.
Its a good idea to keep common things there in base. It will help us onboard new resources quickly.

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
```diff
@@ -76,7 +80,50 @@ public SitesController(IStampService stampService, ISiteService siteService, ICo
...
+            OperationContext<App> cxt = PrepareContext<App>(app, startTimeUtc, endTimeUtc, true);
+
+            var applicableDetectors = await base.GetDetectorsBySupportTopicId(cxt, supportTopicId);
+
+            var insightGroups = await Task.WhenAll(applicableDetectors.Select(detector => base.GetInsightsFromDetector(cxt, detector)));
```
ShekharGupta1988: deleted the older comment. I meant about this one. Shouldn't this logic be encapsulated in Controller Base as its common for any resource.

src/Diagnostics.RuntimeHost/Controllers/ControllerBase.cs
```diff
@@ -67,7 +68,16 @@ protected OperationContext<TResource> PrepareContext<TResource>(TResource resour
                 .Select(p => new DiagnosticApiResponse { Metadata = RemovePIIFromDefinition(p.EntryPointDefinitionAttribute, context.IsInternalCall) });
         }
 
-        protected async Task<IActionResult> GetDetector<TResource>(string detectorId, OperationContext<TResource> context)
+        protected async Task<IActionResult> GetDetectorResponse<TResource>(string detectorId, OperationContext<TResource> context)
+            where TResource : IResource
+        {
+            var detectorResponse = await GetDetector(detectorId, context);
+
```
ShekharGupta1988: nit : Remove Extra line

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    public static class AzureSupportCenterInsightUtilites
+    {
+        public static AzureSupportCenterInsight CreateInsight<TResource>(Insight insight, OperationContext<TResource> context, Definition detector)
+            where TResource : IResource
+        {
+
+            // Logic in line before would have to be updated if we added more resource types. 
+            // TODO: May make sense to have resource type enums contain the resource type string
```
ShekharGupta1988: +1 on TODO. We want to minimize such conditions as it will be hard to onboard a new resource in future.

src/Diagnostics.RuntimeHost/Controllers/ControllerBase.cs
```diff
@@ -93,10 +103,66 @@ protected OperationContext<TResource> PrepareContext<TResource>(TResource resour
...
+            // Not found can occur if invalid detector is put in detector list
+            if (response == null)
+            {
+                return null;
+            }
+
+            List<AzureSupportCenterInsight> supportCenterInsights = new List<AzureSupportCenterInsight>();
+
+            // Take max one insight per detector, only critical or warning, pick the most critical
+            var mostCriticalInsight = response.Insights.Where(insight => ((int)insight.Status) < ((int)InsightStatus.Warning)).OrderBy(insight => insight.Status).FirstOrDefault();
```
hforeste: >var mostCriticalInsight = response.Insights.Where(insight => ((int)insight.Status) < ((int)InsightStatus.Warning)).OrderBy(insight => insight.Status).FirstOrDefault(); [](start = 12, length = 167)

Why can't we show warnings and success insights? Does ASC only show problem areas?

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
stevenernst131: I mentioned in the teams chat that I had this just for testing and I will remove :)

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
ShekharGupta1988: How about **return base.GetInsights(cxt, supportTopicId);**

src/Diagnostics.RuntimeHost/Controllers/ControllerBase.cs
stevenernst131: right now, yes. I actually was going to make it a query string parameter to filter only critical/warning. Still need to get from them what the enum values they have for status

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
stevenernst131: yes I need to change this logic, we will have multiple query string parameters that may act as filters, good point that it should be common in base

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
```diff
@@ -0,0 +1,164 @@
...
+                Links = new List<Link>()
+                    {
+                        new Link()
+                        {
+                            Type = 0,
+                            Text = "Applens Link",
+                            Uri = $"https://applens.azurewebsites.net/{applensPath}"
+                        }
+                    },
+                RecommendedAction = recommendedAction == null ? null :
```
ShekharGupta1988: nit : you can use ??

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
```diff
@@ -0,0 +1,164 @@
...
+                            Uri = $"https://applens.azurewebsites.net/{applensPath}"
+                        }
+                    },
+                RecommendedAction = recommendedAction == null ? null : 
+                    new RecommendedAction()
+                    {
+                        Id = new Guid(),
+                        Text = recommendedAction
+                    },
+                CustomerReadyContent = customerReadyContent == null ? null :
```
ShekharGupta1988: same here. (??)

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
stevenernst131: too used to typescript where they don't have that!

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
```diff
@@ -0,0 +1,164 @@
...
+                Encoding utf8 = Encoding.UTF8;
+                int count = utf8.GetByteCount(detector);
+                byte[] bytes = new byte[count > 16 ? count : 16];
+                Encoding.UTF8.GetBytes(detector, 0, detector.Length, bytes, 0);
+                // Guid only takes byte array of length 16
+                return new Guid(bytes.Take(16).ToArray());
+            }
+            catch(Exception ex)
+            {
+                Console.WriteLine(ex.ToString());
```
ShekharGupta1988: :) . Remove this.

stevenernst131: Oh yes this is another piece left over from testing, good catch

src/Diagnostics.RuntimeHost/Controllers/ControllerBase.cs
```diff
@@ -53,7 +54,7 @@ protected OperationContext<TResource> PrepareContext<TResource>(TResource resour
                 resource,
                 DateTimeHelper.GetDateTimeInUtcFormat(startTime).ToString(DataProviderConstants.KustoTimeFormat),
                 DateTimeHelper.GetDateTimeInUtcFormat(endTime).ToString(DataProviderConstants.KustoTimeFormat),
-                isInternalRequest,
+                isInternalRequest || forceInternal,
```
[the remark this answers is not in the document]
stevenernst131: Basically these calls are coming from geomaster but they are still internal requests essentially. I am open to another option, this was the quickest way for me. 

Right now this API is always internal, at least in my opinion, even if it doesn't have the applens header, which it won't. I could add the applens healder for all calls to this API, but that seems incorrect since it isn't really coming from applens. Thoughts?

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
```diff
@@ -0,0 +1,164 @@
...
+    public class RecommendedAction
+    {
+        public Guid Id { get; set; }
+
+        public Text Text { get; set; }
+    }
+
+    public static class AzureSupportCenterInsightUtilites
+    {
+        public static AzureSupportCenterInsight CreateInsight<TResource>(Insight insight, OperationContext<TResource> context, Definition detector)
```
hforeste: >public static AzureSupportCenterInsight CreateInsight<TResource>(Insight insight, OperationContext<TResource> context, Definition detector) [](start = 8, length = 139)

So technically this is exposed to detector writers. From what I understand, the platform is only interested in using this right?

src/Diagnostics.RuntimeHost/Controllers/ControllerBase.cs
hforeste: Yea query string parameter sounds like a good idea to save yourself a deployment

---
In reply to: [193861176](https://github.com/Azure/Azure-AppServices-Diagnostics/pull/39#discussion_r193861176) [](ancestors = 193861176)

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
stevenernst131: Yes, but since it is used from RuntimeHost.dll it has to be public right?

src/Diagnostics.RuntimeHost/Controllers/ControllerBase.cs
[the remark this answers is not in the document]
ShekharGupta1988: I see. So this flag is set to true in the API we are making for ASC . 
Thats fine then.
Another approach could be we add a user-agent (or something in the headers) at the caller side(geomaster, applens controller) for AppLens and Azure Support Center and then decide whether its internal or external.
Same complexity .. but it can help logging in future.

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
hforeste: Since this is technically a helper for the RuntimeHost I would move it there but because we don't have good discoverability in the detector editing experience you may not need to worry about it.

---
In reply to: [193864564](https://github.com/Azure/Azure-AppServices-Diagnostics/pull/39#discussion_r193864564) [](ancestors = 193864564)

src/Diagnostics.RuntimeHost/Controllers/ControllerBase.cs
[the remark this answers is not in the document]
stevenernst131: Yeah i was thinking user agent as well, I think that is a better approach

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
stevenernst131: I am fine with that, though I am not that worried about someone accidentally using it. @ShekharGupta1988 thoughts? Only since you specifically created a project called models and utils haha

stevenernst131: A side affect of putting the utils there is that they all have to be public

src/Diagnostics.RuntimeHost/Controllers/ControllerBase.cs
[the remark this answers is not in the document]
stevenernst131: I am going to leave this for now because it doesn't seem I can set header as is in geomaster, want to keep changes there to a minimum.

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
ShekharGupta1988: I like Hawk's Idea about moving this method to runtime host project. My motive of creating that Models/Utils dll was to expose functionality to detector authoring and sharing models between various projects. 
But specific helper methods which only one project is going to use can reside in that project itself.

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/AzureSupportCenterInsight.cs
stevenernst131: Actually for these recommendedAction is a Text object but RecommendedAction is a RecommendedAction object so I cannot use the ?? operator, as the two things being compared have to be of the same type.

ShekharGupta1988:
The recent changes also look good. Go forward

PR #40, opened 2018-06-13
Add Category to definition

stevenernst131:
This is just a proof of concept, I will add more predefined categories before I would check in.

Here are the reasons I am doing it this way:

- Enums don't have a way for assigning a string value (with spaces)
- This would give hard coded defaults for people to use so that we are in control of the categories
- If someone wants a category that we have not added, they can use the Custom(name) method, which is really meant to be a stopgap way of not having to wait for a deployment. We would need to push people to add hard coded values later and then updating the detectors. I have some worry about that happening successfully. 
- The UI will just group detectors by category and render them in those groups, no hard coding needed in the UI (applens and support center)
  - The exception to this is old detectors and Puneet's tools etc., those will continue to be hard coded for now.
- They are not defined per resource, we expect our filtering logic to only return detectors of categories that make sense. 

The downside is that with the custom method, people could theoretically add whatever they want and make it public, but we are already at risk for people adding whatever detector they want publicly. I am basically choosing not having to do a georegion deployment to update the categories to the possibility of someone adding one that doesn't make sense.

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/Category.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/Category.cs
@@ -18,7 +18,41 @@ namespace Diagnostics.ModelsAndUtils.Models
             return new Category(name);
         }
 
+        // The predefined categories below are broken out by resource type, 
+        // but it does not mean they can only by used by that resource type.
+
+        // General
         public static readonly Category AvailabilityAndPerformance = new Category("Availability and Performance");
         public static readonly Category ConfigurationAndManagement = new Category("Configuration and Management");
+        public static readonly Category Deployment = new Category("Deployment");
+        public static readonly Category OpenSourceTechnologies = new Category("Open Source Technologies");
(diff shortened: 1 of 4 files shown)


src/Diagnostics.ModelsAndUtils/Models/Category.cs
```diff
@@ -18,7 +18,41 @@ public static Category Custom(string name)
...
+
+        // Site Specific
+        public static readonly Category ProblemsWithAse = new Category("Problems with ASE"); // For site on ASE
+        public static readonly Category ProblemsWithWebjobs = new Category("Problems with Webjobs");
+
+        // Linux Specific
+        public static readonly Category DockerContainers = new Category("DockerContainers");
+
+        // ASE Specific
+        public static readonly Category Networking = new Category("Problems with Webjobs");
```
hforeste: Add the following:

-Network Configuration
-ASE Deployment
-Scaling
-Management

PR #44, opened 2018-06-19
Categories2

stevenernst131:


src/Diagnostics.ModelsAndUtils/Models/Category.cs
```diff
@@ -4,58 +4,46 @@
 
 namespace Diagnostics.ModelsAndUtils.Models
 {
-    public class Category 
+    public static class Category
```
hforeste: nitpik: Can you rename this as `Categories`

PR #46, opened 2018-06-20
Adding additional method to GeomasterDataProvider to call any path

puneetg1983:
This would allow to query any Geomaster API path over HTTP GET and this should help in all those scenarios where we need data from an endpoint other than 'Microsoft.Web/sites'


puneetg1983:
ping @ShekharGupta1988 , @stevenernst131  @hforeste

hforeste:
Hey @puneetg1983 , do you plan to remove the MakeHttpGetRequest since MakeHttpGetRequestWithFullPath is kind of like a superset of that?

puneetg1983:
@hforeste  I am ok with having both the versions of the function as using the existing function querying endpoints under the site are easier where as the new version is specifically written for folks to query end points other than microsoft.web/site.

PR #48, opened 2018-06-26
Onboarding LogicApp, APIM, Certificates and Domain Resources to Generic Platform

ShekharGupta1988:


ShekharGupta1988:
Fixing tests and adding few more for new Resources

ShekharGupta1988:
The kusto Change is actually a breaking change.  So Tests are rightfully failing:
Existing scripts will fail with error :
(5,64): error CS0121: The call is ambiguous between the following methods or properties: 'IKustoDataProvider.ExecuteQuery(string, string, string)' and 'IKustoDataProvider.ExecuteQuery(string, string, string, string)'

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/AppServiceCertificateFilter.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.Attributes
+{
+    /// <summary>
+    /// Class representing Filter for App service certificate
+    /// </summary>
+    public class AppServiceCertificateFilter : ResourceFilterBase
+    {
+        public AppServiceCertificateFilter(bool internalOnly = true) : base(ResourceType.AppServiceCertificate, internalOnly)
```
stevenernst131: Are these constructors necessary?

ShekharGupta1988: Ya, it is. The base constructor has the Resource Type

stevenernst131: True, not sure you need internalOnly as a parameter

PR #51, opened 2018-06-30
Fix logging for Different Resources

ShekharGupta1988:


src/Diagnostics.Logger/ApiMetricsLogger.cs
```diff
@@ -48,27 +48,16 @@ private void ParseFromAddress(HttpContext httpContext)
                     return;
                 }
 
-                Regex webAppRegEx = new Regex("(.*)subscriptions/(.*)/resourcegroups/(.*)/providers/microsoft.web/sites/(.*)/diagnostics/(.*)");
-                Regex hostingEnvRegEx = new Regex("(.*)subscriptions/(.*)/resourcegroups/(.*)/providers/microsoft.web/hostingenvironments/(.*)/diagnostics/(.*)");
+                Regex csmResourceRegEx = new Regex("(.*)subscriptions/(.*)/resourcegroups/(.*)/providers/(.*)/(.*)/(.*)/diagnostics/(.*)");
```
puneetg1983: Do we need capture the provider type somewhere (sites, hostingEnvironments or domains, certs etc.) to be logged to easily identify the resource ?

ShekharGupta1988: Its already present in Address field and can be parsed like:

| parse Address with * "/providers/" provider "/" *

PR #54, opened 2018-07-07
Adding dataProvidersMetadata for onboarding flow as well

puneetg1983:
This will allow applens to show KUSTO Query information even in the on-boarding flow

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -110,10 +110,7 @@ namespace Diagnostics.RuntimeHost.Controllers
                     invocationResponse.UpdateDetectorStatusFromInsights();
 
                     List<DataProviderMetadata> dataProvidersMetadata = null;
-                    if (cxt.IsInternalCall)
-                    {
-                        dataProvidersMetadata = GetDataProvidersMetadata(dataProviders);
-                    }
+                    dataProvidersMetadata = GetDataProvidersMetadata(dataProviders);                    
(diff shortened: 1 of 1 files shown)


src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -108,7 +108,10 @@ public DiagnosticControllerBase(IStampService stampService, ICompilerHostClient
                     var responseInput = new Response() { Metadata = RemovePIIFromDefinition(invoker.EntryPointDefinitionAttribute, cxt.IsInternalCall) };
                     var invocationResponse = (Response)await invoker.Invoke(new object[] { dataProviders, cxt, responseInput });
                     invocationResponse.UpdateDetectorStatusFromInsights();
-                    queryRes.InvocationOutput = DiagnosticApiResponse.FromCsxResponse(invocationResponse);
+
+                    List<DataProviderMetadata> dataProvidersMetadata = null;
+                    dataProvidersMetadata = GetDataProvidersMetadata(dataProviders);
```
ShekharGupta1988: Does this method checks for internal vs external call? You dont want to add the metadata for external calls

puneetg1983: So earlier I was checking cut.IsInternal call and I found it was coming as false. Then I thought it made sense because the onboarding flow is not exposed externally so I removed that check

puneetg1983: Never mind, I was ealrier using DiagCmd to make this onboarding call and that wasnt passing the right internal header. I have updated the PR bringing back the check for internal call.

However , are we allowing external calls to on-boarding flow ?

PR #61, opened 2018-07-27
Add ApplicationInsights data provider

cindylovescoding:
Add ApplicationInsights data provider for diagnostic role.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
+++ b/src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
@@ -1,13 +1,13 @@
-﻿using System;
+﻿using Diagnostics.DataProviders.Interfaces;
+using Diagnostics.Logger;
+using Diagnostics.ModelsAndUtils.Models;
+using Newtonsoft.Json.Linq;
+using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
-using Diagnostics.DataProviders.Interfaces;
-using Diagnostics.Logger;
-using Diagnostics.ModelsAndUtils.Models;
-using Newtonsoft.Json.Linq;
 
 namespace Diagnostics.DataProviders
 {
--- a/src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
+++ b/src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
@@ -4,17 +4,19 @@ using System.Data;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
+using Diagnostics.DataProviders.Interfaces;
 using Diagnostics.Logger;
 using Diagnostics.ModelsAndUtils.Models;
 using Newtonsoft.Json.Linq;
 
 namespace Diagnostics.DataProviders
 {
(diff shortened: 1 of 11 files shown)


ShekharGupta1988:
@cindylovescoding  - this looks good. You can also add few tests using mock data provider. You can follow '\tests\DataProviderTests\DataProviderTests.cs'

src/Diagnostics.DataProviders/DataProviders/AppInsightsDataProvider.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        private readonly IAppInsightsClient _appInsightsClient;
+        private AppInsightsDataProviderConfiguration _configuration;
+
+        public AppInsightsDataProvider(OperationDataCache cache, AppInsightsDataProviderConfiguration configuration) : base(cache)
+        {
+            _configuration = configuration;
+            _appInsightsClient = new AppInsightsClient(_configuration);
+        }
+        public Task<bool> SetAppInsightsKey(string appId, string apiKey)
```
stevenernst131: Does this have to return a Task?

cindylovescoding: Yes seems so. We expose all the functions of dataproviders from DataProviderLogDecorator.cs, and the wrapper MakeDependencyCall<T>(string requestId, Task<T> dataProviderTask, [CallerMemberName]string dataProviderOperation = "") inside only accepts a Task.

hforeste: Any method to a dataprovider were always though of as some kind of IO operation which would require awaiting (Executing Kusto Query, Observer API calls, GeoMaster API calls). When creating the decorator class, I didn't foresee anyone setting properties with it. But this works.

hforeste:
@ShekharGupta1988 , I remember that we are showing customer app exceptions from AppInsights in our app service diagnostics blade. How did you get the APIKey?

src/Diagnostics.DataProviders/DataProviders/AppInsightsDataProvider.cs
ShekharGupta1988: There is one more problem we found in this design. You cannot have same method names in different Data Providers. Since Kusto Data Provider has an ExecuteQuery , so App Insights Data Provider cannot have method called ExecuteQuery, since at the end all methods needs to be redeclared in the decorator class..

ShekharGupta1988:
@hforeste - That was just a portal specific operation using auth token. Not the backend.. To start with, This Provider will be used only by monitoring API where she will specifically set the api key.

src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
```diff
@@ -1,20 +1,22 @@
-﻿using System;
+﻿using Diagnostics.DataProviders.Interfaces;
+using Diagnostics.Logger;
+using Diagnostics.ModelsAndUtils.Models;
+using Newtonsoft.Json.Linq;
```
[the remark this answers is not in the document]
cindylovescoding: Sure thanks Hawk. I am changing that:)

hforeste:
I see. The reason for the ask is to see if its possible to get customer appinsights data through this data provider. I know without the id and appkey we can't although it may be possible to get such information through Puneet's GeoMasterDataProvider AppSettings API.

PR #62, opened 2018-07-27
Support for Dropdown Rendering Type

ShekharGupta1988:
Adding Support for Dropdown View. The View Model is a (key: string, value: Response) pair where:
- "key" goes in the dropdown list
- "value" is the Response (insights, graphs, markdown etc.) that gets displayed when a key is selected.


ShekharGupta1988:
Adding Unit Tests

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Dropdown.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// First Item represents item in Dropdown, Second Item represents whether selected by default in dropdwon, Third Item represents the body after dropdown selection
+        /// </summary>
+        public List<Tuple<string, bool, Response>> Data;
+
+        /// <summary>
+        /// Creates an instance of Dropdown Class
+        /// </summary>
+        /// <param name="label">Dropdown Label</param>
+        /// <param name="data">Dropdown Data</param>
+        public Dropdown(string label, List<Tuple<string, bool, Response>> data)
```
stevenernst131: What would happen if multiple are selected? Might be a question for the UI

stevenernst131:
This looks good, you may want to do UI changes first, so that you don't have to go back and make updates here.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Dropdown.cs
ShekharGupta1988: the last one gets selected :)

ShekharGupta1988:
Ya, I have done UI changes also and validated end-to-end.

stevenernst131:
Awesome!

PR #64, opened 2018-08-02
Use only one Status enum and update event Id

stevenernst131:
Updating event id in order to make the data flow properly

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorStatus.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorStatus.cs
@@ -19,11 +19,11 @@ namespace Diagnostics.ModelsAndUtils.Models.ResponseExtensions
         /// <code>
         /// public async static Task<![CDATA[<Response>]]> Run(DataProviders dp, OperationContext cxt, Response res)
         /// {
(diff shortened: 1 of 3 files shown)


src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorStatus.cs
```diff
@@ -19,11 +19,11 @@ public static class ResponseDetectorStatusExtensions
         /// <code>
         /// public async static Task<![CDATA[<Response>]]> Run(DataProviders dp, OperationContext cxt, Response res)
         /// {
-        ///     res.SetDetectorStatus(DetectorStatus.Critical, "This is an error message that will come from the detector");
+        ///     res.SetDetectorStatus(InsightStatus.Critical, "This is an error message that will come from the detector");
         /// }
         /// </code>
         /// </example>
-        public static void SetDetectorStatus(this Response response, DetectorStatus status, string message = null)
+        public static void SetDetectorStatus(this Response response, InsightStatus status, string message = null)
```
ShekharGupta1988: I guess, this method is not getting used in developers? it'll be a breaking change

stevenernst131: It is not being used anywhere.

PR #65, opened 2018-08-03
Enable Insights for All Resource Types

stevenernst131:
@ShekharGupta1988 Let me know what you think of the way I handled having multiple resource types. Open to other suggestions

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/Enumerations.cs
+++ b/src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/Enumerations.cs
@@ -4,6 +4,60 @@ using System.Text;
 
 namespace Diagnostics.ModelsAndUtils.Attributes
 {
+    public class ArmResourceType
+    {
+        public ResourceType ResourceType { get; set; }
+
+        public string Provider { get; set; }
+
+        public string ResourceTypeName { get; set; }
+    }
+
+    public static class ArmResourceTypes
+    {
+        public static readonly ArmResourceType App = new ArmResourceType()
+        {
+            ResourceType = ResourceType.App,
+            Provider = "Microsoft.Web",
+            ResourceTypeName = "sites"
+        };
+
+        public static readonly ArmResourceType HostingEnvironment = new ArmResourceType()
+        {
+            ResourceType = ResourceType.HostingEnvironment,
+            Provider = "Microsoft.Web",
+            ResourceTypeName = "hostingEnvironments"
+        };
+
+        public static readonly ArmResourceType AppServiceCertificate = new ArmResourceType()
+        {
+            ResourceType = ResourceType.AppServiceCertificate,
+            Provider = "Microsoft.CertificateRegistration",
+            ResourceTypeName = "certificateOrders"
+        };
+
+        public static readonly ArmResourceType AppServiceDomain = new ArmResourceType()
+        {
+            ResourceType = ResourceType.AppServiceDomain,
+            Provider = "Microsoft.DomainRegistration",
+            ResourceTypeName = "domains"
+        };
+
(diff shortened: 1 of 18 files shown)


src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/Enumerations.cs
```diff
@@ -4,6 +4,60 @@
 
 namespace Diagnostics.ModelsAndUtils.Attributes
 {
+    public class ArmResourceType
```
ShekharGupta1988: One Suggestion -These fields can be part of IResource Interface and we can define the values in Inherited Resource Objects. 
In the Controller Base Class, you have access to TResource : IResource. So you'll be able to read these values.

Then, when adding a new Resource, we dont have to define this ArmResourceType.

src/Diagnostics.RuntimeHost/Controllers/ApiManagement/ApiManagementController.cs
```diff
@@ -12,6 +13,14 @@ namespace Diagnostics.RuntimeHost.Controllers
     [Route(UriElements.ApiManagementServiceResource)]
     public sealed class ApiManagementController : DiagnosticControllerBase<ApiManagementService>
     {
+        protected override ArmResourceType ResourceType
```
ShekharGupta1988: Look at my suggestion above. If we go with that approach, you dont have to define this override in the Controller class.
The Resource Object will have fields.

src/Diagnostics.RuntimeHost/Utilities/AzureSupportCenterInsightUtilities.cs
```diff
@@ -22,15 +22,11 @@ internal static class AzureSupportCenterInsightUtilites
...
             where TResource : IResource
         {
-
-            // Logic in line before would have to be updated if we added more resource types. 
-            // TODO: May make sense to have resource type enums contain the resource type string
-            var resourceTypeString = context.Resource.ResourceType == ResourceType.App ? "sites" : "hostingEnvironments";
-            var issueCategory = context.Resource.ResourceType == ResourceType.App ? "Web App" : "App Service Environment";
-            var applensPath = $"subscriptions/{context.Resource.SubscriptionId}/resourceGroups/{context.Resource.ResourceGroup}/providers/Microsoft.Web/{resourceTypeString}/{context.Resource.Name}/detectors/{detector.Id}?startTime={context.StartTime}&endTime={context.EndTime}";
+            var applensPath = $"subscriptions/{context.Resource.SubscriptionId}/resourceGroups/{context.Resource.ResourceGroup}/providers/{armResourceType.Provider}/{armResourceType.ResourceTypeName}/{context.Resource.Name}/detectors/{detector.Id}?startTime={context.StartTime}&endTime={context.EndTime}";
+            var category = detector.Name.Length > 32 ? armResourceType.ResourceTypeName : detector.Name;
```
ShekharGupta1988: Is there a limitation on the category length at ADS side?

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/Enumerations.cs
stevenernst131: Good suggestion, let me try that.

src/Diagnostics.RuntimeHost/Utilities/AzureSupportCenterInsightUtilities.cs
stevenernst131: You bet there is! Before I just used "Web App" for every insight because I was worrying about the length but the guy complained so I figured using the detector name when possible would be better. Also just makes it clear what the insight is for.

stevenernst131:
@ShekharGupta1988 Let me know what you think of the update!

PR #67, opened 2018-08-16
Move Publish Detector API in runtime host

ShekharGupta1988:


src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -189,6 +189,30 @@ protected async Task<IActionResult> ExecuteQuery<TPostBodyResource>(TResource re
             return Ok(queryRes);
         }
 
+        protected async Task<IActionResult> PublishDetector(DetectorPackage pkg)
+        {
+            if (pkg == null || string.IsNullOrWhiteSpace(pkg.Id) || string.IsNullOrWhiteSpace(pkg.DllBytes))
```
puneetg1983: >pkg.Id [](start = 57, length = 6)

if pkg is NULL, wouldnt this generate a NRE ?

puneetg1983: >pkg.DllBytes [](start = 94, length = 12)

Same here

src/Diagnostics.RuntimeHost/Controllers/AppServiceDomain/AppServiceDomainController.cs
```diff
@@ -52,5 +52,11 @@ public async Task<IActionResult> GetInsights(string subscriptionId, string resou
         {
             return await base.GetInsights(GetResource(subscriptionId, resourceGroupName, domainName), supportTopicId, minimumSeverity, startTime, endTime, timeGrain);
         }
+
+        [HttpPost(UriElements.Publish)]
+        public async Task<IActionResult> PublishDetector(string subscriptionId, string resourceGroupName, string domainName, [FromBody] DetectorPackage pkg)
```
stevenernst131: Do we need a resource specific endpoint for this operation?

ShekharGupta1988: Not Really. Just to stay consistent.

src/Diagnostics.RuntimeHost/Services/GithubClient/GithubClient.cs
```diff
@@ -87,13 +94,33 @@ public async Task DownloadFile(string fileUrl, string destinationPath)
...
+        public async Task CreateOrUpdateFile(string filePath, string content, string commitMessage, bool convertContentToBase64 = true)
+        {
+            try
+            {
+                // try to get the file (and with the file the last commit sha)
+                var existingFile = await _octokitClient.Repository.Content.GetAllContentsByRef(_userName, _repoName, filePath, _branch);
+
+                // update the existing file
+                var updateChangeSet = await _octokitClient.Repository.Content.UpdateFile(_userName, _repoName, filePath,
+                   new UpdateFileRequest(commitMessage, content, existingFile.First().Sha, _branch, convertContentToBase64));
```
cindylovescoding: Shekhar what will happen if there's  file existing but no actual changes happened? Will it be update too?

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/GitHubWatcher.cs
```diff
@@ -48,6 +49,37 @@ public override void Start()
...
+            }
+
+            Tuple<bool, Exception> output = new Tuple<bool, Exception>(true, null);
+
+            try
+            {
+                string detectorFilePath = $"{pkg.Id.ToLower()}/{pkg.Id.ToLower()}";
+
+                string csxFilePath = $"{detectorFilePath}.csx";
+                string dllFilePath = $"{detectorFilePath}.dll";
```
cindylovescoding: Since we have this information right now. Maybe we can leverage this to build change history for each detector?

src/Diagnostics.RuntimeHost/Services/GithubClient/GithubClient.cs
ShekharGupta1988: Ideally, that call shouldn't come through from applens. But lets say such call comes through. Then there would be a call to update the file with same content.
However, the file SHA will not be modified (since content is same) and we will not sync that file from github (as github will respond with no changes detected)

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/GitHubWatcher.cs
ShekharGupta1988: Exactly. We should be able to do it right now also.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
puneetg1983: Silly me :)

PR #68, opened 2018-08-16
Fix site level for ase

hforeste:
Set the resource name for hosting environment to the stamp name
This may break some ASE detectors but there are few so its easy to edit them.

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -280,6 +280,7 @@ namespace Diagnostics.RuntimeHost.Controllers
 
             HostingEnvironment hostingEnv = new HostingEnvironment(subscriptionId, resourceGroup, name)
             {
+                Name = stampPostBody.InternalName,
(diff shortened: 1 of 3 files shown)


src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -304,6 +304,7 @@ protected async Task<HostingEnvironment> GetHostingEnvironment(string subscripti
 
             HostingEnvironment hostingEnv = new HostingEnvironment(subscriptionId, resourceGroup, name)
             {
+                Name = stampPostBody.InternalName,
```
cindylovescoding: FriendlyName?

hforeste: So in the world of ASE there are two names the internal name and friendly name. The friendly name is an easier to remember ASE name which is the name that customer set and the ugly name aka internal name is the waws-prod- prefixed name.

cindylovescoding: Thanks Hawk for the explanation:)

PR #70, opened 2018-08-29
Adding ASC Insight Extension Method

stevenernst131:


src/Diagnostics.ModelsAndUtils/Diagnostics.ModelsAndUtils.csproj
```diff
@@ -9,6 +9,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="CommonMark.NET" Version="0.15.1" />
```
ShekharGupta1988: make sure this dll is part of the build package. If its not, we would need to update dotnet feed to incliude this package in dotnet package cache. (Few referenced packages are part of dotnet package cache)

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -130,13 +130,14 @@ public void LogRuntimeHostHandledException(string RequestId, string Source, stri
...
-        public void LogRuntimeHostDetectorAscInsight(string RequestId, string Source, string Message)
+        [Event(2007, Level = EventLevel.Informational, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogRuntimeHostDetectorAscInsight, Version = 2)]
+        public void LogRuntimeHostDetectorAscInsight(string RequestId, string Source, string Message, string Content)
         {
             WriteDiagnosticsEvent(2007,
                 RequestId,
                 Source,
-                Message);
+                Message,
+                Content);
```
ShekharGupta1988: New Column? maybe call it 'Details'

src/Diagnostics.ModelsAndUtils/Models/Response.cs
```diff
@@ -33,6 +33,11 @@ public class Response
         /// </summary>
         public List<Insight> Insights { get; set; }
 
+        /// <summary>
+        /// ASC Insight List
+        /// </summary>
+        public List<AzureSupportCenterInsight> AscInsights { get; set; }
```
ShekharGupta1988: having this outside DiagnosticData means it will always show up on top irrespective where developer is writing it?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/AscInsight.cs
```diff
@@ -0,0 +1,103 @@
...
+        ///                    Operation failed because a current scale operation is ongoing.
+        ///                    ```
+        ///         ", true), // This is markdown
+        ///         new Text(@"Copy and paste the Customer Ready Content and send to the customer"), // This is not markdown
+        ///         new Text(@$"Your scale operation for site ***{cxt.Resource.Name}*** failed because there was a current scale operation underway. 
+        ///                    
+        ///                    Please wait for the current operation to finish and then try again to scale your app service plan. ", true),
+        ///         cxt,
+        ///         new Dictionary<string, string>() {
+        ///             { "Header", "<markdown>This is extra insight members that you want to be shown in ***Applens*** and ***App Service Diagnostics***.</markdown>"
```
ShekharGupta1988: I am guessing this will break document generation. you would need to use CDATA to preserve html tags.
Just test this out locally:
1. Install docfx : https://github.com/dotnet/docfx/releases (save the folder and update PATH)
2. Go to 'data\documentation' folder in our project root
3. run 'docfx'  **(this is the step that needs to pass)**
4. run 'docfx serve _site'

stevenernst131: Perfect I was going to ask you how to do this. I will figure out how to make it work.

src/Diagnostics.ModelsAndUtils/Models/Response.cs
stevenernst131: No, these are not actually returned from the API. For insights calls I return them, otherwise I add an insight to diagnostic data.

src/Diagnostics.ModelsAndUtils/Diagnostics.ModelsAndUtils.csproj
stevenernst131: How do I check that?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/AscInsight.cs
stevenernst131: I have fixed this by removing that dictionary section with the markdown tags. Not the example looks like this, though the formatting is off because of the strange markdown strings, not sure what to do about that. 

image

stevenernst131: Also do I check in the .manifest file?

src/Diagnostics.ModelsAndUtils/Diagnostics.ModelsAndUtils.csproj
ShekharGupta1988: just run 'build' from root and check the runtime host package folder whether it has this dll or not. If its there, then all's good.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/AscInsight.cs
ShekharGupta1988: ya, you can add that too.

src/Diagnostics.ModelsAndUtils/Diagnostics.ModelsAndUtils.csproj
stevenernst131: It is there :)

PR #71, opened 2018-08-30
Runtime site name map utility

hforeste:
This utility will allow detector authors to filter site events to get relevant slot events


src/Diagnostics.ModelsAndUtils/Models/Resource/App.cs
```diff
@@ -76,7 +81,16 @@ public App(string subscriptionId, string resourceGroup, string appName) : base()
         {
             this.SubscriptionId = subscriptionId;
             this.ResourceGroup = resourceGroup;
-            this.Name = appName;
+            var appNameAndSlot = appName.Split(new char[] { '(', ')' }, StringSplitOptions.RemoveEmptyEntries);
+            this.Name = appNameAndSlot[0];
```
cindylovescoding: Do we need to check appNameAndSlot[0] exists or not as well:)

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Utilities.cs
```diff
@@ -121,5 +122,64 @@ public static string HostNamesFilterQuery(IEnumerable<string> hostnames, string
...
+                        if (siteName.Equals(slotTimeRangeInfo.RuntimeSitename, StringComparison.CurrentCultureIgnoreCase) && timeStamp >= slotTimeRangeInfo.StartTime && timeStamp <= slotTimeRangeInfo.EndTime)
+                        {
+                            dt.Rows.Add(row.ItemArray);
+                            break;
+                        }
+                    }
+                }
+            }catch(Exception ex)
+            {
+                throw new Exception("Failed to get runtime site events table.", ex);
```
ShekharGupta1988: maybe include 'runtime slotmap' in the message as now that is a well-known term in our team and we can identify quickly which method is throwing this exception.

hforeste:
Hey @stevenernst131 , my change broke your TestDataTableToDataTableResponseObjectAndBack when running on local machine. Your test was expecting a local time but my utility was converting each time to UTC. I made the fix, take a look and leave any comments or feedback if necessary

PR #72, opened 2018-08-30
Make Email Buttons Disabled By Default

stevenernst131:
Back end change so that the detector owner can decide if they want to show the email buttons or not in applens


ShekharGupta1988:
oh nice...
Should we also do this for title blue(or purple) box.

Lot of times I just wanted to put some sub-headings and cant use markdown as it wraps that in box?

stevenernst131:
I haven't done the UI change yet, if you don't specify a title or to show the buttons, I was going to get rid of the blue box header

PR #74, opened 2018-09-06
Internal/External View

stevenernst131:
Also removed unused insight inputs, and added logic to call a default insight for no coverage cases.

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using Diagnostics.ModelsAndUtils.Attributes;
+using System;
 using System.Collections.Generic;
 using System.Text;
 
@@ -26,10 +27,16 @@ namespace Diagnostics.ModelsAndUtils.Models
         public string EndTime { get; private set; }
 
         /// <summary>
+        /// sets to false when detector should run as it is being called from external source (Azure portal, CLI ...)
+        /// sets to true when detector should run as it is being called from internal source (Applens ..)
+        /// </summary>
+        public bool IsInternalCall { get; private set; }
+
+        /// <summary>
         /// sets to false when detector is called from external source (Azure portal, CLI ...)
         /// sets to true when detector is called from internal source (Applens ..)
         /// </summary>
-        public bool IsInternalCall { get; private set; }
+        public bool IsInternalClient { get; private set; }
 
         /// <summary>
         /// Request Id
@@ -37,18 +44,26 @@ namespace Diagnostics.ModelsAndUtils.Models
         public string RequestId { get; private set; }
 
         /// <summary>
+        /// If this detector is being run in the context of a support topic, i.e. called from Azure Support Center
+        /// then this value will be populated with the selected support topic
+        /// </summary>
+        public SupportTopic SupportTopic { get; private set; }
+
+        /// <summary>
         /// TimeGrain in minutes for aggregating data.
         /// </summary>
         public string TimeGrain { get; private set; }
 
-        public OperationContext(TResource resource, string startTimeStr, string endTimeStr, bool isInternalCall, string requestId, string timeGrain = "5")
+        public OperationContext(TResource resource, string startTimeStr, string endTimeStr, bool isInternalCall, bool isInternalClient, string requestId, string timeGrain = "5", SupportTopic supportTopic = null)
         {
             Resource = resource;
             StartTime = startTimeStr;
             EndTime = endTimeStr;
+            IsInternalClient = isInternalClient;
             IsInternalCall = isInternalCall;
             RequestId = requestId;
             TimeGrain = timeGrain;
+            SupportTopic = supportTopic;
         }
     }
 }
--- a/src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
@@ -33,12 +33,6 @@ namespace Diagnostics.ModelsAndUtils.Models
         public bool IsInternalCall { get; private set; }
 
         /// <summary>
-        /// sets to false when detector is called from external source (Azure portal, CLI ...)
-        /// sets to true when detector is called from internal source (Applens ..)
-        /// </summary>
-        public bool IsInternalClient { get; private set; }
-
-        /// <summary>
         /// Request Id
         /// </summary>
         public string RequestId { get; private set; }
@@ -54,12 +48,11 @@ namespace Diagnostics.ModelsAndUtils.Models
         /// </summary>
         public string TimeGrain { get; private set; }
 
-        public OperationContext(TResource resource, string startTimeStr, string endTimeStr, bool isInternalCall, bool isInternalClient, string requestId, string timeGrain = "5", SupportTopic supportTopic = null)
+        public OperationContext(TResource resource, string startTimeStr, string endTimeStr, bool isInternalCall, string requestId, string timeGrain = "5", SupportTopic supportTopic = null)
         {
             Resource = resource;
             StartTime = startTimeStr;
             EndTime = endTimeStr;
-            IsInternalClient = isInternalClient;
             IsInternalCall = isInternalCall;
             RequestId = requestId;
             TimeGrain = timeGrain;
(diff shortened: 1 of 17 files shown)


src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
```diff
@@ -25,30 +26,44 @@ public class OperationContext<TResource> where TResource : IResource
...
+        /// sets to true when detector should run as it is being called from internal source (Applens ..)
+        /// </summary>
+        public bool IsInternalCall { get; private set; }
+
         /// <summary>
         /// sets to false when detector is called from external source (Azure portal, CLI ...)
         /// sets to true when detector is called from internal source (Applens ..)
         /// </summary>
-        public bool IsInternalCall { get; private set; }
+        public bool IsInternalClient { get; private set; }
```
ShekharGupta1988: what is InternalClient? applens or asc?

src/Diagnostics.ModelsAndUtils/Utilities/AzureSupportCenterInsightUtilities.cs
```diff
@@ -11,6 +11,9 @@ namespace Diagnostics.ModelsAndUtils.Utilities
 {
     public static class AzureSupportCenterInsightUtilites
     {
+        
+
```
ShekharGupta1988: empty lines :)

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -152,10 +152,10 @@ protected async Task<IActionResult> ExecuteQuery<TPostBodyResource>(TResource re
                             if (!VerifyEntity(invoker, ref queryRes)) return Ok(queryRes);
                             OperationContext<TResource> cxt = PrepareContext(resource, startTimeUtc, endTimeUtc);
 
-                            var responseInput = new Response() { Metadata = RemovePIIFromDefinition(invoker.EntryPointDefinitionAttribute, cxt.IsInternalCall) };
+                            var responseInput = new Response() { Metadata = RemovePIIFromDefinition(invoker.EntryPointDefinitionAttribute, cxt.IsInternalClient) };
                             invocationResponse = (Response)await invoker.Invoke(new object[] { dataProviders, cxt, responseInput });
                             invocationResponse.UpdateDetectorStatusFromInsights();
-                            isInternalCall = cxt.IsInternalCall;
+                            isInternalCall = cxt.IsInternalClient;
```
ShekharGupta1988: I guess I dont uderstand what is InternalClient and how is it different from InternalCall?

src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
stevenernst131: IsInternalClient means Applens really. ASC I added that 'forceInternal' flag for insights API, so it works a little differently, though I could probably improve that logic.

ShekharGupta1988: Thats same as IsInternalCall, no?

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -220,25 +220,27 @@ private DiagnosticApiResponse CreateQueryExceptionResponse(Exception ex, Definit
...
+            IEnumerable<Definition> allDetectors = null;
             try
             {
                 supportTopicId = ParseCorrectSupportTopicId(supportTopicId);
-                var allDetectors = (await ListDetectorsInternal(cxt)).Select(detectorResponse => detectorResponse.Metadata);
+                allDetectors = (await ListDetectorsInternal(cxt)).Select(detectorResponse => detectorResponse.Metadata);
 
                 var applicableDetectors = allDetectors
-                    .Where(detector => string.IsNullOrWhiteSpace(supportTopicId) || detector.SupportTopicList.FirstOrDefault(supportTopic => supportTopic.Id == supportTopicId) != null);
+                    .Where(detector => string.IsNullOrWhiteSpace(supportTopicId) || detector.SupportTopicList.FirstOrDefault(st => st.Id == supportTopicId) != null);
```
ShekharGupta1988: you probably need to match PesId too also now for applicableDetectors

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
stevenernst131: InternalCall determines whether the detector runs as if it is being called by internal or external. I didn't change this because it is already being used and I didn't want to break anyone. 

InternalClient means that it is being called from Applens, or any other internal source in the future. 

Reason is that lets say you want to run a detector in applens, but see the output as a customer would. You want to run a detector that is actually internalOnly. If you try to run that detector before with InternalCall = false, you would get a 404 because it would get filtered out. That is why you need the logic of internal vs external client and internal vs. external view separate.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
stevenernst131: Our filtering was taking care of this right? Right now I don't have pesId piped up completely. Nothing should have changed in this part of the logic.

src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
stevenernst131: see my below comment and let me know if that makes more sense.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
stevenernst131: You can have an internal client that wants an internal view or an internal client that wants an external (customer) view. If it is an external client, then internal call will always be false.

ShekharGupta1988: This seems too much confusing :) 

Instead of this, can we add a field like user-agent or caller (of string type or enum type
) which can be populated by clients and gets populated in detectors context.

Applens calls ==> 'applens'
ASC calls ==> 'asc'
Any other ==> 'other'

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
ShekharGupta1988: Filtering will take care it in general. 
Rare scenario : if you have a detector marked for windows and linux , but support topic is only for windows, you dont want to trigger this detector in ASC for linux support topic (with same id)

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
ShekharGupta1988: or maybe rename the field ? bool type is easy in the scenario you described. However, I wonder if folks starts using InternalClient flag to make a decision for internal call vs external call.

src/Diagnostics.RuntimeHost/Utilities/DateTimeHelper.cs
```diff
@@ -117,7 +117,7 @@ internal static bool PrepareStartEndTimeUtc(string startTime, string endTime, ou
                 return false;
             }
 
-            if (endTimeUtc - startTimeUtc > TimeSpan.FromHours(24))
+            if (endTimeUtc - startTimeUtc > TimeSpan.FromHours(72))
             {
                 errorMessage = "Invalid Time Range. Time Range cannot be more than 24 hours.";
```
ShekharGupta1988: fix the error message to 72 hrs :)

stevenernst131:
@ShekharGupta1988 I ended up adding a RuntimeContext object because it was really the only way to go after looking at the code. Take a look and let me know what you think. It is working very nicely.

src/Diagnostics.ModelsAndUtils/Models/RuntimeContext.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.Models
+{
+    public class RuntimeContext<TResource>
```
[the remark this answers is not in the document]
stevenernst131: You are right, thought of that as I was creating it and thought it might be being used in Scripts project, I see now that it is not. I will put it in RuntimeHost project.

PR #75, opened 2018-09-07
Modify source watcher logging

hforeste:
Show source watcher type of exceptions in diagnostic dashboard.

### DIFF
--- a/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/SourceWatcherBase.cs
+++ b/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/SourceWatcherBase.cs
@@ -17,6 +17,8 @@ namespace Diagnostics.RuntimeHost.Services.SourceWatcher
         protected IInvokerCacheService _invokerCache;
         protected string _eventSource;
 
+        protected abstract string SourceName { get; }
+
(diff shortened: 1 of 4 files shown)


src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/SourceWatcherBase.cs
```diff
@@ -45,10 +47,8 @@ protected void LogWarning(string message)
 
         protected void LogException(string message, Exception ex)
         {
-            string exceptionType = ex != null ? ex.GetType().ToString() : string.Empty;
-            string exceptionDetails = ex != null ? ex.ToString() : string.Empty;
-
-            DiagnosticsETWProvider.Instance.LogSourceWatcherException(_eventSource, message, exceptionType, exceptionDetails);
+            var exception = new SourceWatcherException(SourceName, message, ex);
```
ShekharGupta1988: do you think this class is unnecessary? 
the source watcher base class already has the Source Name and all this exception is doing is formatting a message.

hforeste: Honestly I did it just for the exception type field so that we can group by on that when we look at the dashboard

PR #76, opened 2018-09-10
Adding GetContainerLogs method to GeoMasterProvider

puneetg1983:


src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
```diff
@@ -184,6 +184,11 @@ public Task<T> MakeHttpGetRequestWithFullPath<T>(string fullPath, string querySt
             return MakeDependencyCall(null, _geomasterDataProvider.MakeHttpGetRequestWithFullPath<T>(fullPath, queryString, apiVersion));
         }
 
+        public Task<string> GetContainerLogs(string subscriptionId, string resourceGroupName, string name)
```
ShekharGupta1988: I am guessing its linux container (docker or ACR).  Remember there is support for Windows Containers now also in antares. 
If this method is only targeting Linux, maybe call it 'GetLinuxContainerLogs' ?

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -364,6 +364,36 @@ public async Task<T> MakeHttpGetRequestWithFullPath<T>(string fullPath, string q
...
+        ///     var name = cxt.Resource.Name;
+        ///     
+        ///     string containerLogs = await dp.GeoMaster.GetContainerLogs(subId, rg, name);
+        ///         
+        ///     // do any processing on the string variable containerLogs
+        /// }
+        /// </code>
+        /// </example>
+        /// <returns></returns>
+        public async Task<string> GetContainerLogs(string subscriptionId, string resourceGroupName, string name)
```
ShekharGupta1988: (This is probably true of all methods).
Recently, @hforeste did a change to runtime to separate out site name with slot name and pass it in Context resource.

Does this method "SitePathUtility.GetSitePath()" take into consideration slotName also? or parsing 'app(slot)' perhaps

src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
puneetg1983: good point, changed it to 'GetLinuxContainerLogs'

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
puneetg1983: Yup it does. It allows passing site+slot as siteName(slot)

PR #77, opened 2018-09-12
Updating GeoMasterDataProvider to work with slots

puneetg1983:


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -67,9 +69,9 @@ private IGeoMasterClient InitClient()
         /// }
         /// </code>
         /// </example>
-        public async Task<IDictionary<string,string>> GetAppSettings(string subscriptionId, string resourceGroupName, string name)
+        public async Task<IDictionary<string,string>> GetAppSettings(string subscriptionId, string resourceGroupName, string name, string slotName)
```
hforeste: >string slotName [](start = 131, length = 15)

You can make this an optional parameter by setting this equal to a default value such as GeoMasterConstants.ProductionSlot. By doing that you don't have to create another GetAppSettings method.
And the same for the other new methods you added.
What you did is fine though, this is just a preference and something I typically see that C# developers do.

ShekharGupta1988: Hawk - this is the correct thing to do in order to maintain backward compatibility. If you change the signature of the existing method (even with optional param), then detectors which are using tht method would need to recompiled.

hforeste: Oh i see. Kool. Thanks for that clarification man 👍

puneetg1983: Exactly what @ShekharGupta1988 said :-) 

I had it initially with Optional parameters but I found that existing DLLs were breaking so did it this way. 

Thanks for the review!

PR #79, opened 2018-09-17
 LSI Fix : Handle Exceptions in Kusto Token Service

ShekharGupta1988:
For whatever reasons, if Kusto token refresh fails, the next token refresh will not happen right now as the service will stop. The fix is to handle exceptions and log. 
There is already alerts now covering this scenario.

### DIFF
--- a/src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
+++ b/src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
@@ -221,12 +221,11 @@ namespace Diagnostics.Logger
         }
 
         [Event(3003, Level = EventLevel.Warning, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogKustoTokenRefreshSummary)]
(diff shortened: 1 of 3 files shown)


src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -219,6 +219,21 @@ public void LogDataProviderOperationSummary(string RequestId, string Source, str
                 EndTime,
                 LatencyInMilliseconds);
         }
+
+        [Event(3003, Level = EventLevel.Warning, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogKustoTokenRefreshSummary)]
```
stevenernst131: Should this event just be info level? If it is going to be healthy most of the time.

stevenernst131:
Thanks for making this change @ShekharGupta1988, I think it was assigned to me but I had trouble getting to it.

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
ShekharGupta1988: yes. good catch. (copy paste error)

PR #81, opened 2018-09-24
Create postbody for site and ase

cindylovescoding:
Create postbody for site and ASE controller, so we don't need to construct ASE and site specific postbody from client side each time, which will also make all the resource types requests in consistency.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
@@ -127,15 +127,15 @@ namespace Diagnostics.DataProviders
                     .FirstOrDefault(
                         j => (j.ContainsKey("Subscription") && j["Subscription"].ToString() == subscriptionId
                             && j.ContainsKey("ResourceGroupName") && j["ResourceGroupName"].ToString() == resourceGroupName
-                            && j.ContainsKey("stampName")));
+                            && j.ContainsKey("StampName")));
 
-            string stampName = obj2?["stampName"]?.ToString();
+            string stampName = obj2?["StampName"]?.ToString();
             return stampName;
         }
 
-        public override async Task<dynamic> GetHostNames(string siteName)
+        public override async Task<dynamic> GetHostNames(string stampName, string siteName)
         {
-            var response = await Get($"sites/{siteName}/hostnames");
+            var response = await Get($"stamps/{stampName}/sites/{siteName}/hostnames");
             var hostNames = JsonConvert.DeserializeObject(response);
             return hostNames;
         }
@@ -147,7 +147,7 @@ namespace Diagnostics.DataProviders
 
             if (sitePostBody["HostNames"] == null)
             {
-                var hostNames = await GetHostNames(siteName);
+                var hostNames = await GetHostNames(stampName, siteName);
                 List<dynamic> hostNamesList = new List<dynamic>();
                 foreach (var hostName in hostNames)
                 {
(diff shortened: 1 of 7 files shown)


src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -118,6 +119,54 @@ public override async Task<dynamic> GetSite(string stampName, string siteName)
...
+        public override async Task<dynamic> GetHostNames(string siteName)
+        {
+            var response = await Get($"sites/{siteName}/hostnames");
+            var hostNames = JsonConvert.DeserializeObject(response);
+            return hostNames;
+        }
+
+        public override async Task<dynamic> GetSitePostBody(string stampName, string siteName)
+        {
+            var response = await Get($"stamps/{stampName}/sites/{siteName}/postbody");
```
ShekharGupta1988: Can we make the PostBody call and Hostname call in parallel instead of a sequential await.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -118,6 +119,54 @@ public override async Task<dynamic> GetSite(string stampName, string siteName)
...
+                            && j.ContainsKey("ResourceGroupName") && j["ResourceGroupName"].ToString() == resourceGroupName
+                            && j.ContainsKey("stampName")));
+
+            string stampName = obj2?["stampName"]?.ToString();
+            return stampName;
+        }
+
+        public override async Task<dynamic> GetHostNames(string siteName)
+        {
+            var response = await Get($"sites/{siteName}/hostnames");
```
hforeste: >$"sites/{siteName}/hostnames") [](start = 37, length = 30)

Use this path: stamps/{stamp}/sites/{site}/hostnames.
SiteNames are not unique so sometimes we can have a sitename in a public stamp and that same sitename in a ministamp. This call will return hostnames for both sites.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -118,6 +119,54 @@ public override async Task<dynamic> GetSite(string stampName, string siteName)
...
+            var hostNames = JsonConvert.DeserializeObject(response);
+            return hostNames;
+        }
+
+        public override async Task<dynamic> GetSitePostBody(string stampName, string siteName)
+        {
+            var response = await Get($"stamps/{stampName}/sites/{siteName}/postbody");
+            dynamic sitePostBody = JsonConvert.DeserializeObject(response);
+
+            if (sitePostBody["HostNames"] == null)
```
hforeste: >if (sitePostBody["HostNames"] == null) [](start = 12, length = 38)

The postBody response will always not have hostnames filled because the API was rushed when I created it lol. We can make that change in the SupportBayApi if you like so that you don't need to make an API call to get hostnames.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
cindylovescoding: Oh I see, Good Point. Let me change this.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -131,7 +167,7 @@ public override HttpClient GetObserverClient()
         /// <returns></returns>
         private async Task<string> Get(string path)
         {
-            var request = new HttpRequestMessage(HttpMethod.Get, $"https://support-bay-api.azurewebsites.net/observer/{path}?api-version=2.0");
```
hforeste: >}?api-version=2.0" [](start = 123, length = 18)

Hey Cindy, I'm pretty sure you changed this because the GET postbody calls are failing with a 400 when adding the api-version. I can fix this part on my side so that you don't need to do this.

cindylovescoding: Thanks a lot Hawk! That will be great! I will revert this back after your changes:)

PR #84, opened 2018-10-08
Improve logging

hforeste:
- Log request id in data provider logging. Events 3001-3003
- Add 60 second timeout to request

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -80,7 +80,7 @@ namespace Diagnostics.RuntimeHost.Controllers
             Dictionary<string, dynamic> systemContext = PrepareSystemContext(resource, detectorId, dataSource, timeRange);
 
             await this._sourceWatcherService.Watcher.WaitForFirstCompletion();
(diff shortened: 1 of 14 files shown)


src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
```diff
@@ -29,6 +33,12 @@ public async Task Invoke(HttpContext httpContext)
             {
                 await _next(httpContext);
             }
+            catch (TimeoutException ex)
+            {
+                httpContext.Response.Clear();
```
ShekharGupta1988: You'll get an exception here if the Response has Stated. Something like "it cannot be modified at this time". So you should check if the response has not started and then update it.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -420,6 +418,8 @@ private RuntimeContext<TResource> PrepareContext(TResource resource, DateTime st
                 supportTopic: supportTopic
             );
 
+            var tokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(60 * 1000));
```
ShekharGupta1988: shouldnt this be read from Host Constants?

PR #86, opened 2018-10-25
Add entry point definition to the context

cindylovescoding:


src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -374,18 +374,25 @@ protected async Task<HostingEnvironment> GetHostingEnvironment(string subscripti
 
             var invoker = this._invokerCache.GetDetectorInvoker<TResource>(detectorId, runtimeContext);
             IEnumerable<SupportTopic> supportTopicList = null;
-            if (invoker != null && invoker.EntryPointDefinitionAttribute != null && invoker.EntryPointDefinitionAttribute.SupportTopicList != null && invoker.EntryPointDefinitionAttribute.SupportTopicList.Any())
+            Definition definition = null;
+            if (invoker != null && invoker.EntryPointDefinitionAttribute != null)
             {
-                supportTopicList = invoker.EntryPointDefinitionAttribute.SupportTopicList;
+                if (invoker.EntryPointDefinitionAttribute.SupportTopicList != null && invoker.EntryPointDefinitionAttribute.SupportTopicList.Any())
```
stevenernst131: is there really a need for this second if check? seems like you could just say 

```C#
supportTopicList = invoker.EntryPointDefinitionAttribute.SupportTopicList;
```

without the nested if, if it is null then it will be null, just like it would be anyway by its current design.

cindylovescoding: I am checking invoker.EntryPointDefinitionAttribute.SupportTopicList.Any() to avoid the empty array which is the default value to be assigned.

hforeste: You can put this within one if body. That way it would be more readable.

stevenernst131: Could make default value null also. Not sure if there was an intention there.

PR #87, opened 2018-10-26
ASC Insight Updates

stevenernst131:


src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -140,6 +140,16 @@ public void LogRuntimeHostDetectorAscInsight(string RequestId, string Source, st
                 Details);
         }
 
+        [Event(2008, Level = EventLevel.Informational, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogRuntimeHostDetectorAscInsight)]
```
cindylovescoding: Should this be LogRuntimeHostSupportTopicAscInsight ?

stevenernst131: yes good catch, though I am not even sure these are really being used.

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -140,6 +140,16 @@ public void LogRuntimeHostDetectorAscInsight(string RequestId, string Source, st
...
+        public void LogFullAscInsight(string RequestId, string Source, string Message, string Details)
+        {
+            WriteDiagnosticsEvent(2007,
```
hforeste: Change to correct event id

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -140,6 +140,16 @@ public void LogRuntimeHostDetectorAscInsight(string RequestId, string Source, st
                 Details);
         }
```
hforeste: Do you need this anymore since your new event is essentially a superset of this event?

stevenernst131: Don't really need it, it was not working, so I figured easiest thing to do was just to create a new event

PR #89, opened 2018-10-30
Kusto and geochanges - Method to call DAAS and detectors to get Kusto Query 

puneetg1983:
With this PR, there are two methods that I am adding

1.  **dp.Kusto.GetKustoQuery** - This is for internal detectors that just need KUSTO query information and don't want to execute the query (Basically just give a link for further diagnostics)

   ```
   var k = await dp.Kusto.GetKustoQuery(GetQuery(cxt), cxt.Resource.Stamp.Name);
   var kustoDesktop = k.KustoDesktopUrl;
```

2. **Ability to call DAAS API's (GET only)**. This will allow us to get information about site properties in future (like ASP.NET Core versions , Diagnosers supported etc.)

   ```
   var resp = await dp.GeoMaster.InvokeDaasExtension<object>(cxt.Resource.SubscriptionId, cxt.Resource.ResourceGroup, cxt.Resource.Name, cxt.Resource.Slot, "api/diagnosers");
   ```

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
+++ b/src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
@@ -68,6 +68,11 @@ namespace Diagnostics.DataProviders
             return ExecuteQuery(query, DataProviderConstants.FakeStampForAnalyticsCluster, requestId, operationName);
         }
 
+        public Task<KustoQuery> GetKustoQuery(string query, string stampName)
+        {
+            return MakeDependencyCall(_kustoDataProvider.GetKustoQuery(query, stampName));
+        }
+
         public Task<JObject> GetAdminSitesByHostNameAsync(string stampName, string[] hostNames)
         {
             return MakeDependencyCall(_observerDataProvider.GetAdminSitesByHostNameAsync(stampName, hostNames));
@@ -228,6 +233,11 @@ namespace Diagnostics.DataProviders
         public Task<string> GetLinuxContainerLogs(string subscriptionId, string resourceGroupName, string name, string slotName)
         {
             return MakeDependencyCall(_geomasterDataProvider.GetLinuxContainerLogs(subscriptionId, resourceGroupName, name, slotName));
+        }
+
+        public Task<T> InvokeDaasExtension<T>(string subscriptionId, string resourceGroupName, string name, string slotName, string daasApiPath, string apiVersion = GeoMasterConstants.August2016Version, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            return MakeDependencyCall(_geomasterDataProvider.InvokeDaasExtension<T>(subscriptionId, resourceGroupName, name, slotName, daasApiPath, apiVersion, cancellationToken));
         }
 
         public Task<VnetValidationRespone> VerifyHostingEnvironmentVnet(string subscriptionId, string vnetResourceGroup, string vnetName, string vnetSubnetName, CancellationToken cancellationToken = default(CancellationToken))
--- a/src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
+++ b/src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
@@ -72,6 +72,10 @@ namespace Diagnostics.DataProviders
         {
             return MakeDependencyCall(_kustoDataProvider.GetKustoQuery(query, stampName));
         }
+        public Task<KustoQuery> GetKustoClusterQuery(string query)
+        {
+            return MakeDependencyCall(_kustoDataProvider.GetKustoClusterQuery(query));
+        }
 
         public Task<JObject> GetAdminSitesByHostNameAsync(string stampName, string[] hostNames)
         {
@@ -294,5 +298,7 @@ namespace Diagnostics.DataProviders
                 }
             }
         }
+
+       
     }
 }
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/Diagnostics.DataProviders/DataProviderLogDecorator.cs [hunk 1]
```diff
@@ -68,6 +68,11 @@ public Task<DataTable> ExecuteClusterQuery(string query, string requestId = null
             return ExecuteQuery(query, DataProviderConstants.FakeStampForAnalyticsCluster, requestId, operationName);
         }
 
+        public Task<KustoQuery> GetKustoQuery(string query, string stampName)
+        {
+            return MakeDependencyCall(_kustoDataProvider.GetKustoQuery(query, stampName));
+        }
```
hforeste: Is it intentional that you want detector authors to be able to call this?

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -459,6 +459,54 @@ private async Task<T> InvokeSiteExtension<T>(string subscriptionId, string resou
...
+        ///                     "api/diagnosers");
+        /// 
+        ///     // do something with the response object
+        ///     var responseFromDaaS = resp.ToString();
+        ///     return res;
+        /// }
+        /// </code>
+        /// </example>
+        /// <returns></returns>
+        public async Task<T> InvokeDaasExtension<T>(string subscriptionId, string resourceGroupName, string name, string slotName, string daasApiPath, string apiVersion = GeoMasterConstants.August2016Version, CancellationToken cancellationToken = default(CancellationToken))
```
hforeste: >InvokeDaasExtension [](start = 29, length = 19)

I think this is really kool. So essentially this is a step closer to the point where we can have buttons on detectors and the button can just call this to start for example profiler or something.

hforeste: >(string subscriptionId, string resourceGroupName, string name, string slotName, [](start = 51, length = 79)

This is a nitpik. You don't have to do this. The Diagnostics.DataProvider project is already referencing the Diagnostics.ModelsAndUtils project. Where I'm going with this is that you can create an overloaded method that just takes an IResource. IResource already has subscription, resourcegroup, and resource name information. 

This way its easier to use this method. Detector authors will just have to say `dp.GeoMaster.InvokeDaasExtension(cxt.Resource, "dasApiPath")` and thats it

src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
hforeste: Nevermind I just read your note on the top :)

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
puneetg1983: ooh, we are not going there yet. This might get a bit dangerous if we allow people to do it. This is only for DAAS GET calls where DAAS can help us get more information about customer's app without including PII. Things like node version, .net core version and other things that we might get from web.config in future...

src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
puneetg1983: :-)

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
puneetg1983: This is a really cool suggestion! let me implement this and push a commit

puneetg1983: I actually want to keep *both*, detectors passing cxt.Resource and all params...This would be in case if you want to do something on a slot or say a different app (I don't know maybe that scenario will arise at some point?)

puneetg1983: I take it back, I don't want to over publicize this method specially but this is good suggestion for any future function that I add in GeoMasterDataProvider and will keep it in mind :-)

src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
```diff
@@ -68,6 +68,11 @@ public Task<DataTable> ExecuteClusterQuery(string query, string requestId = null
             return ExecuteQuery(query, DataProviderConstants.FakeStampForAnalyticsCluster, requestId, operationName);
         }
 
+        public Task<KustoQuery> GetKustoQuery(string query, string stampName)
```
ShekharGupta1988: should we add one more for GetKustoClusterQuery(string query) , so no stamp name as input

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -459,6 +459,54 @@ private async Task<T> InvokeSiteExtension<T>(string subscriptionId, string resou
...
+        {
+            if (string.IsNullOrWhiteSpace(daasApiPath))
+            {
+                throw new ArgumentNullException(daasApiPath);
```
ShekharGupta1988: I guess you forgot to stringify the argument. It should be "daasApiPath"

src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
puneetg1983: I didn't even knew that we had a method call ExecuteClusterQuery. Your ask maskes sense so pushing a commit

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
puneetg1983: good catch and silly me. Thanks for spotting this

PR #90, opened 2018-11-14
Only Cache TenantIds for Public Stamps

ShekharGupta1988:
In V2, I was not caching TenantIds for ASE because they can change if ASE gets recreated.
However, somehow missed that fact while coding up for V3. Hence Updating the TenantId code to only cache the value for Public Stamps (as they will not change)

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -344,7 +344,7 @@ namespace Diagnostics.RuntimeHost.Controllers
 
             string stampName = !string.IsNullOrWhiteSpace(hostingEnv.InternalName) ? hostingEnv.InternalName : hostingEnv.Name;
 
-            var result = await this._stampService.GetTenantIdForStamp(stampName, startTime, endTime, (DataProviderContext)HttpContext.Items[HostConstants.DataProviderContextKey]);
+            var result = await this._stampService.GetTenantIdForStamp(stampName, hostingEnv.HostingEnvironmentType == HostingEnvironmentType.None,  startTime, endTime, (DataProviderContext)HttpContext.Items[HostConstants.DataProviderContextKey]);
             hostingEnv.TenantIdList = result.Item1;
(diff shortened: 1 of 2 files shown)


stevenernst131:
I understand the reasoning behind this change, however I would like to just bring up another point, and that is that for ASE sites this query at the 50th percentile takes 700ms and at the 90th percentile takes 3 seconds. Source: (I have filtered out the 0s since that means it hit the cache)

DiagnosticRole
| where TIMESTAMP  > ago(7d)
| where EventId == 63005 and Source == "GetTenantIdForStamp"
| where LatencyInMilliseconds > 0
| summarize percentiles(LatencyInMilliseconds, 50, 75, 90, 99)

  |   |   |   |  
-- | -- | -- | -- | --
  | percentile_LatencyInMilliseconds_50 | percentile_LatencyInMilliseconds_75 | percentile_LatencyInMilliseconds_90 | percentile_LatencyInMilliseconds_99
  | 704 | 1210 | 3683 | 59913

This is a blocking operation which means that it makes every single call to an ASE site that much longer. If we are fine with that then we can move forward I just want to make sure we know this is an expensive change to make to solve a problem that seems to be somewhat rare?

One other thing I could think of is that we use the cache always and continue with the request, then still run the task to get tenant ids, and later await it and throw an exception if there is a miss in the cache, and then the user could retry and the cache would have the updated value.

hforeste:
I didn't realize the 90th percentile was almost 4 seconds. In that case, let's cache tenant id for ASE. Because the stamp name that were adding to the cache is the ugly name, when the ASE get's recreated then the ugly name will change and we'll properly fetch the correct tenant id for that stamp name. 

I think the biggest thing is that we should protect against caching empty tenant ids.

stevenernst131:
Can you expand on the empty tenant ids? Was this related to Puneets email where he saw failures because the tenantId is empty? What caused this?

hforeste:
The logic that is susceptible to caching empty tenant id is a factor in Puneet's experience with empty tenant id. I'm not sure what may have been the cause in Puneet's examples but what could be a cause for ASE/WebApps in ASE is that the ASE itself might be suspended/deleted.

PR #93, opened 2018-11-21
Fix Deserialization Error + Parent detector view

stevenernst131:
Example exception that we will now avoid:

System.AggregateException: One or more errors occurred. (Unexpected character encountered while parsing value: [. Path 'Tables[0].Rows[0][3]', line 1, position 608.) ---> Newtonsoft.Json.JsonReaderException: Unexpected character encountered while parsing value: [. Path 'Tables[0].Rows[0][3]', line 1, position 608.
   at Newtonsoft.Json.JsonTextReader.ReadStringValue(ReadType readType)
   at Newtonsoft.Json.JsonTextReader.ReadAsString()
   at Newtonsoft.Json.JsonReader.ReadForType(JsonContract contract...

Also fix parent detector view when a child is showing red but actually doesn't have any insights.

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/DataTableResponseObject.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/DataTableResponseObject.cs
@@ -17,7 +17,7 @@ namespace Diagnostics.ModelsAndUtils.Models
 
         public IEnumerable<DataTableResponseColumn> Columns { get; set; }
 
-        public string[][] Rows { get; set; }
+        public dynamic[,] Rows { get; set; }
     }
 
     public class DataTableResponseColumn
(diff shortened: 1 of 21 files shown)


src/Diagnostics.ModelsAndUtils/Models/DataTableResponseObject.cs
```diff
@@ -17,7 +17,7 @@ public class DataTableResponseObject
 
         public IEnumerable<DataTableResponseColumn> Columns { get; set; }
 
-        public string[][] Rows { get; set; }
+        public dynamic[,] Rows { get; set; }
```
ShekharGupta1988: breaking change?

hforeste: >dynamic[,] Rows [](start = 15, length = 15)

Just to make sure I'm understanding this correctly, did the deserialization error occur because arrays or objects from kusto response string could not be deserialized to strings?

stevenernst131: no it should not be. Everything was working end to end. In JSON the two would be the same. This object is not accessed by the detector code, as we give them the data table and then convert it to this.

stevenernst131: Correct. There was something there in the value that looked like 

`['value','value',['array','from','makeset']]`

and it choked on the inner array when it tried to read it as string.

PR #94, opened 2018-12-17
Adding Kusto logging to mirror logging done for old detectors

stevenernst131:
Also uncoupling kusto client from any logic that is specific to web apps/app service. Move logic to get cluster to kusto data provider.

### DIFF
--- a/src/Diagnostics.DataProviders/MockKustoClient.cs
+++ b/src/Diagnostics.DataProviders/MockKustoClient.cs
@@ -10,11 +10,16 @@ namespace Diagnostics.DataProviders
 {
     class MockKustoClient: IKustoClient
     {
-        public async Task<DataTable> ExecuteQueryAsync(string query, string stampName, string requestId = null, string operationName = null)
+        public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, string database, string requestId = null, string operationName = null)
         {
-            if (string.IsNullOrWhiteSpace(stampName))
+            if (string.IsNullOrWhiteSpace(cluster))
             {
                 throw new ArgumentNullException("stampName");
+            }
+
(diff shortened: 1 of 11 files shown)


src/Diagnostics.DataProviders/MockKustoClient.cs
```diff
@@ -81,12 +86,17 @@ private Task<DataTable> GetLatestDeployment()
             return Task.FromResult(table);
         }
 
-        public Task<KustoQuery> GetKustoQueryAsync(string query, string stampName)
+        public Task<KustoQuery> GetKustoQueryAsync(string query, string cluster, string database)
         {
-            if (string.IsNullOrWhiteSpace(stampName))
+            if (string.IsNullOrWhiteSpace(cluster))
             {
                 throw new ArgumentNullException("stampName");
```
ShekharGupta1988: same herde

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -233,6 +233,19 @@ public void LogKustoTokenRefreshSummary(string Source, string Message, long Late
                 ExceptionDetails);
         }
 
+        [Event(3004, Level = EventLevel.Informational, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogKustoQueryInformation)]
+        public void LogKustoQueryInformation(string Source, string RequestId, string Message, long LatencyInMilliseconds, string Details, string Content, string ExceptionType, string ExceptionDetails)
```
ShekharGupta1988: QQ : will this add a new column for Content in Kusto or is it already there

stevenernst131: It is already there

PR #95, opened 2018-12-20
Adding card as a response extension

puneetg1983:


src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Card.cs
```diff
@@ -0,0 +1,138 @@
...
+            table.Columns.Add(new DataColumn("ActionType", typeof(string)));
+            table.Columns.Add(new DataColumn("ActionValue", typeof(string)));
+
+            cards.ForEach(card =>
+            {
+                table.Rows.Add(new object[] {
+                    card.Title,
+                    card.Icon,
+                    card.ShortDescription,
+                    JsonConvert.SerializeObject(card.Descriptions),
```
stevenernst131: You may not have to serialize these values, because the rows is dynamic[,] now which means they would be allowed. When you do a makeset in kusto for example, it returns an array, not a string serialized array. Can you try and see what the response looks like without serializing them?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Card.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// <returns></returns>
+        /// <example> 
+        /// This sample shows how to use <see cref="AddCards"/> method.
+        /// <code>
+        /// public async static Task<![CDATA[<Response>]]> Run(DataProviders dp, OperationContext cxt, Response res)
+        /// {
+        ///     var cards = new List<![CDATA[<Card>]]>();
+        ///     cards.Add(new Card(
+        ///         {
+        ///             Title = "HTTP 4XX Errors",
```
stevenernst131: You actually don't have a 0 parameter constructor so this does not work. Either add one or use the full constructor with all the params

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Card.cs
stevenernst131: Ignore this comment, I was wrong, it did not serialize correctly on its own.

puneetg1983: Yea that's what I thought that serialization is needed ...

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Card.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// <param name="response">Response</param>
+        /// <param name="cards">List<![CDATA[<Card>]]></param>
+        /// <returns></returns>
+        /// <example> 
+        /// This sample shows how to use <see cref="AddCards"/> method.
+        /// <code>
+        /// public async static Task<![CDATA[<Response>]]> Run(DataProviders dp, OperationContext cxt, Response res)
+        /// {
+        ///     var cards = new List<![CDATA[<Card>]]>();
+        ///     cards.Add(new Card(
```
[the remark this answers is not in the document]
puneetg1983: My bad . Should have been more careful. Will correct it after creating a small snippet...

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Card.cs
puneetg1983: Will correct the whole code comment block...

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Card.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        /// <summary>
+        /// Title of the Card
+        /// </summary>
+        public string Title;
+
+        /// <summary>
+        /// A list of descriptions for this card
+        /// </summary>
+        public string ShortDescription;
```
stevenernst131: Did we decide to keep or remove the short description? I think I have only mocked it up as having a list of items, which is the descriptions below.

puneetg1983: To be very frank, I forgot what we discussed....it's been a while now. If that's what we agreed , I can remove but see if you recall. 😊

stevenernst131: Yes lets just keep description, and remove that. If you push that change, I will sign off and i will try to get this stuff deployed ASAP.

puneetg1983: done!

PR #98, opened 2019-01-12
Remove the wawsobserver path validation

hforeste:
I removed logic that checked that a path in observer exist. Recently I added new APIs for Mahatab and he couldn't access them through the observer data provider because the path did not exist in the static list of paths written in the SupportObserverDataProviderBase class. Unfortunately I did some over engineering there.

In this pull request, anytime someone sends a request to observer with dp.Observer.GetResource, all this class will do is validate the host is in fact WawsObsever and then it will just send the request via HttpClient. If it doesn't exist, then the detector will get a 404 as expected.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
@@ -15,7 +15,6 @@ namespace Diagnostics.DataProviders
     {
         protected readonly SupportObserverDataProviderConfiguration _configuration;
         private readonly HttpClient _httpClient;
-        private List<string> _routeTemplates;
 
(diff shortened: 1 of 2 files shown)


src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -77,28 +75,8 @@ public Task<dynamic> GetResource(string resourceUrl)
...
-                }
-            }
-
-            if (routeParametersAndValues == null)
-            {
-                throw new ArgumentException("Obsever may not have an API for your observerUrl");
-            }
-
-            var apiPath = CreateObserverQueryAndPath(routeParametersAndValues, routeTemplate);
+            var apiPath = uri.PathAndQuery.Substring(1);
```
stevenernst131: Does this just take off the initial '/'?

PR #100, opened 2019-01-16
Account for null row values and throw exception in kusto client

stevenernst131:
Fixing exceptions that looked like this:

```
System.AggregateException: One or more errors occurred. (Cannot set Column 'S_siteid' to be null. Please use DBNull instead.) ---> System.ArgumentException: Cannot set Column 'S_siteid' to be null. Please use DBNull instead.
```

Adding user agent to kusto client in kestrel process. 

Throwing exception on failed kusto query.


stevenernst131:
There were 4 or 5 public detectors failing. A few more internal. I patched the queries for the public detectors to prevent failures but need to have a permanent fix for this.

PR #101, opened 2019-01-17
MDM data provider

crazyxy:


stevenernst131:
I know you are working on the nuget problem, but I don't think we should be checking in exe's, unless they are for some reason needed.

src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        /// <summary>
+        /// Gets the time series.
+        /// </summary>
+        /// <param name="startTimeUtc">The start time UTC.</param>
+        /// <param name="endTimeUtc">The end time UTC.</param>
+        /// <param name="sampling">The sampling type.</param>
+        /// <param name="definition">The time series definition.</param>
+        /// <returns>The time series for the given definition.</returns>
+        public async Task<IEnumerable<DataTable>> GetTimeSeriesAsync(DateTime startTimeUtc, DateTime endTimeUtc, Sampling sampling, Tuple<string, string, IEnumerable<KeyValuePair<string, string>>> definition)
```
hforeste: I think detector writers will find it more of a hard time to create the tuple definition for `Tuple<string, string, IEnumerable<KeyValuePair<string, string>>> definition`. Instead consider creating separate parameters for the namespace, metricname, and the combination of mdm dimensions.

src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        /// <summary>
+        /// Gets the dimension values for dimensionName satifying the dimensionFilters and
+        /// </summary>
+        /// <param name="metricNamespace">Metric namespace</param>
+        /// <param name="metricName">Metric name</param>
+        /// <param name="filter">The dimension filters representing the pre-aggregate dimensions. Create an emtpy include filter for dimension with no filter values. Requested dimension should also be part of this and should be empty.</param>
+        /// <param name="dimensionName">Name of the dimension for which values are requested.</param>
+        /// <param name="startTimeUtc">Start time for evaluating dimension values.</param>
```
hforeste: Don't make the startTime and endTime change default just yet, I didn't realize that we have to make changes in other parts of the project before each data provider can inherit the starttime and endtime from the UI. Let's just get this out and then I can talk to you afterwards on how to inherit the startTime and endTime

src/Diagnostics.DataProviders/MdmClient/MdmClient.cs
```diff
@@ -0,0 +1,451 @@
...
+                }
+
+                stage = "ValidateStatus";
+                response.EnsureSuccessStatusCode();
+
+                return Tuple.Create(responseString, response);
+            }
+            catch (Exception e)
+            {
+                var message = $"Failed to get a response from the server. TraceId:{traceId.Value.ToString("B")}, Url:{request.RequestUri}, HandlingServer:{handlingServer} Stage:{stage}, "
```
hforeste: Every time a data provider makes an outbound call and fails we should log the request URL so that we can look at the logs and repro it. So log this message in the finally clause and generalize the message for both success and failures. 

You don't have to log the response string. Unless you know the response string has something very useful for failures.

src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
crazyxy: Okay, I will keep the change.

src/Diagnostics.DataProviders/MdmClient/MdmClient.cs
```diff
@@ -353,6 +353,7 @@ private static HttpClient CreateHttpClient(string thumbprint)
                         (e.ResponseStatusCode != null && e.ResponseStatusCode != HttpStatusCode.ServiceUnavailable) ||
                         i == numAttempts)
                     {
+                        DiagnosticsETWProvider.Instance.LogDataProviderMessage(RequestId, "MdmClient", $"GetResponse Failure. Latency: {stopWatch.Elapsed}; Number of attempts: {i}; Exception: {e.ToString()}");
```
hforeste: If you log it here then we won't be able to track _all_ the retries from MDM. Consider logging this within a finally clause.

src/Diagnostics.DataProviders/MdmClient/MdmClient.cs
```diff
@@ -427,17 +429,20 @@ private static HttpClient CreateHttpClient(string thumbprint)
...
 
+                message = $"Succeeded to get a response from the server. Url: {request.RequestUri}, HandlingServer:{handlingServer}";
+
                 return Tuple.Create(responseString, response);
             }
             catch (Exception e)
             {
-                var message = $"Failed to get a response from the server. TraceId:{traceId.Value.ToString("B")}, Url:{request.RequestUri}, HandlingServer:{handlingServer} Stage:{stage}, "
-                    + $"LatencyMs:{requestLatency.ElapsedMilliseconds}, ResponseStatus:{response?.StatusCode.ToString() ?? "<none>"}, Response:{responseString}";
+                message = $"Failed to get a response from the server. Url:{request.RequestUri}, HandlingServer:{handlingServer} Stage:{stage}, "
```
hforeste: 👍

src/Diagnostics.DataProviders/MdmClient/MdmClient.cs
```diff
@@ -427,17 +429,20 @@ private static HttpClient CreateHttpClient(string thumbprint)
                 stage = "ValidateStatus";
                 response.EnsureSuccessStatusCode();
 
+                message = $"Succeeded to get a response from the server. Url: {request.RequestUri}, HandlingServer:{handlingServer}";
```
hforeste: 👍

src/Diagnostics.DataProviders/Utility/CertificateHelper.cs
```diff
@@ -18,31 +18,31 @@ internal static X509Certificate2 FindAndValidateCertificate(string certificateTh
             var cert = FindX509Certificate(certificateThumbprint, certificateStoreLocation);
             if (!cert.HasPrivateKey)
             {
-                throw new MetricsClientException(string.Format("Cert with Thumbprint [{0}] doesn't have a private key", cert.Thumbprint));
+                throw new InvalidOperationException(string.Format("Cert with Thumbprint [{0}] doesn't have a private key", cert.Thumbprint));
```
hforeste: This would not be considered an InvalidOperationException. You can just throw `Exception`. Please revisit the kind of exceptions you are throwing for this class. Either throw just a regular `Exception` type or find the appropriate type.

tests/DataProviderTests/DataProviderTests.cs
```diff
@@ -151,8 +151,7 @@ public async void TestMdmGetTimeSeriesValuesAsync()
             metadata.ScriptText = @"
                 public async static Task<IEnumerable<DataTable>> Run(DataProviders dataProviders) {
                     var dimensions = new Dictionary<string, string> { { ""StampName"", ""kudu1"" } };
-                    var definition = Tuple.Create<string, string, IEnumerable<KeyValuePair<string, string>>>(""Microsoft/Web/WebApps"", ""CpuTime"", dimensions);
-                    return await dataProviders.Mdm.GetTimeSeriesAsync(DateTime.UtcNow.AddMinutes(-10), DateTime.UtcNow, Sampling.Average | Sampling.Max | Sampling.Count, definition);
+                    return await dataProviders.Mdm.GetTimeSeriesAsync(DateTime.UtcNow.AddMinutes(-10), DateTime.UtcNow, Sampling.Average | Sampling.Max | Sampling.Count, ""Microsoft/Web/WebApps"", ""CpuTime"", dimensions);
```
hforeste: Your changes to the parameters now allows you to run this in just one line of code as opposed to two 🥇

src/Diagnostics.DataProviders/Utility/CertificateHelper.cs
crazyxy: I'd like to throw a regular `Exception` type exception.

PR #102, opened 2019-02-01
Default to all for mc

hforeste:
Azure China does not have an analytics cluster where we can fetch stack type. By default, return all detectors regardless of stack type if the runtime fails to get stack type


src/Diagnostics.RuntimeHost/Services/ResourceService/SiteService.cs
```diff
@@ -26,13 +27,21 @@ public async Task<StackType> GetApplicationStack(string subscriptionId, string r
...
                 | project sitestack";
 
-            DataTable stackTable = await dp.Kusto.ExecuteQuery(queryTemplate, DataProviderConstants.FakeStampForAnalyticsCluster, operationName: "GetApplicationStack");
+            DataTable stackTable = null;
+            
+            try{
+                stackTable = await dp.Kusto.ExecuteQuery(queryTemplate, DataProviderConstants.FakeStampForAnalyticsCluster, operationName: "GetApplicationStack");
+            }catch(Exception ex){
+                //swallow the exception. Since Mooncake does not have an analytics cluster
+                DiagnosticsETWProvider.Instance.LogRuntimeHostHandledException(dataProviderContext.RequestId, "GetApplicationStack", subscriptionId,
```
ShekharGupta1988: Is there a way to know which cloud we are running in? like Public or Non-Public.
There is a dependency of stack in Filters as well as we pass the data into Context for detector authors to use.

Another orthogonal question - what if there is an exception here. What is getting Passed to context? Is it StackType.None

hforeste: Yep StackType.None would be in the context. There might be a way to find out what cloud we're in but I didn't pursue that route as I think this is the fastest route.

src/Diagnostics.RuntimeHost/Services/ResourceService/SiteService.cs
```diff
@@ -36,7 +36,10 @@ public async Task<StackType> GetApplicationStack(string subscriptionId, string r
...
             
             try{
-                stackTable = await dp.Kusto.ExecuteQuery(queryTemplate, DataProviderConstants.FakeStampForAnalyticsCluster, operationName: "GetApplicationStack");
+                if (dataProviderContext.Configuration.KustoConfiguration.CloudDomain == KustoDataProviderConfiguration.AzureCloud)
+                {
+                    stackTable = await dp.Kusto.ExecuteQuery(queryTemplate, DataProviderConstants.FakeStampForAnalyticsCluster, operationName: "GetApplicationStack");
+                }
             }catch(Exception ex){
                 //swallow the exception. Since Mooncake does not have an analytics cluster
                 DiagnosticsETWProvider.Instance.LogRuntimeHostHandledException(dataProviderContext.RequestId, "GetApplicationStack", subscriptionId,
```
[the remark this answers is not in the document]
hforeste: Yea I think a warning would be better. Perhaps we can change the event level for this event since this event is logged only when we swallow the exception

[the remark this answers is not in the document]
ShekharGupta1988: I dont think you would be able to change the Level of the same event to warning in this code and error (when its used in Middleware)

I think I added a WarningEvent for runtime host that we can use. not sure though.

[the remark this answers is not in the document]
hforeste: No this is not the same one used in the Middleware. Its really subtle but this one is the Handled exception event while the one in the middleware is the unhanded exception event

src/Diagnostics.DataProviders/DataProviderConfigurations/KustoDataProviderConfiguration.cs
```diff
@@ -53,6 +57,25 @@ public class KustoDataProviderConfiguration : IDataProviderConfiguration
...
+            {
+                if (AADKustoResource.Contains("windows.net"))
+                {
+                    return AzureCloud;
+                }
+                else if (AADKustoResource.Contains("chinacloudapi.cn"))
+                {
+                    return AzureChinaCloud;
+                }
+                else
```
puneetg1983: >else [](start = 16, length = 4)

Do you want to check the check for Blackforest (.cloudapi.de) or is black forest not in scope yet ?

src/Diagnostics.RuntimeHost/Services/ResourceService/SiteService.cs
```diff
@@ -26,13 +27,24 @@ public async Task<StackType> GetApplicationStack(string subscriptionId, string r
...
-            DataTable stackTable = await dp.Kusto.ExecuteQuery(queryTemplate, DataProviderConstants.FakeStampForAnalyticsCluster, operationName: "GetApplicationStack");
+            DataTable stackTable = null;
+            
+            try{
+                if (dataProviderContext.Configuration.KustoConfiguration.CloudDomain == KustoDataProviderConfiguration.AzureCloud)
+                {
+                    stackTable = await dp.Kusto.ExecuteQuery(queryTemplate, DataProviderConstants.FakeStampForAnalyticsCluster, operationName: "GetApplicationStack");
+                }
+            }catch(Exception ex){
+                //swallow the exception. Since Mooncake does not have an analytics cluster
```
puneetg1983: >//swallow the exception. Since Mooncake does not have an analytics cluster [](start = 16, length = 74)

Is this comment outdated as you are already checking for CloudDomain == AzureCloud ?

As Shekhar was suggesting, would also be lovely to have a property added to Context that tells us what cloud we are hitting (if not too much work to do)...

src/Diagnostics.DataProviders/DataProviderConfigurations/KustoDataProviderConfiguration.cs
hforeste: Correct, not within the scope

PR #103, opened 2019-02-04
AKS and Guage Rendering

nmallick1:
Adding support for AKS resource type and Guage rendering

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/Resource/AzureKubernetesService.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/Resource/AzureKubernetesService.cs
@@ -1,0 +1,61 @@
+﻿using Diagnostics.ModelsAndUtils.Attributes;
+
+namespace Diagnostics.ModelsAndUtils.Models
+{
+    /// <summary>
+    /// Resource representing Azure Kubernetes Service
+    /// </summary>
+    public class AzureKubernetesService : AzureKubernetesServiceFilter, IResource
+    {
+        /// <summary>
+        /// Name of the Azure Kubernetes Service
+        /// </summary>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Subscription Id(Guid)
+        /// </summary>
+        public string SubscriptionId { get; set; }
+
+        /// <summary>
+        /// Resource Group Name
+        /// </summary>
+        public string ResourceGroup { get; set; }
+
+        public string Provider
+        {
+            get
+            {
+                return "Microsoft.ContainerService";
+            }
+        }
+
+        /// <summary>
+        /// Name of Resource Type as defined by ARM resource id. Examples: 'sites', 'hostingEnvironments'
+        /// </summary>
+        public string ResourceTypeName
+        {
+            get
+            {
+                return "managedClusters";
+            }
+        }
+
+        /// <summary>
+        /// Determines whether the Azure Kubernetes Service resource is applicable after filtering.
+        /// </summary>
+        /// <param name="filter">Resource Filter</param>
+        /// <returns>True, if resource passes the filter. False otherwise</returns>
+        public bool IsApplicable(IResourceFilter filter)
+        {
+            return filter is AzureKubernetesServiceFilter;
+        }
+
+        public AzureKubernetesService(string subscriptionId, string resourceGroup, string appName) : base()
+        {
+            this.SubscriptionId = subscriptionId;
+            this.ResourceGroup = resourceGroup;
+            this.Name = appName;
+        }
+    }
+}
--- a/src/Diagnostics.ModelsAndUtils/Models/Resource/AzureKubernetesService.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/Resource/AzureKubernetesService.cs
@@ -51,11 +51,11 @@ namespace Diagnostics.ModelsAndUtils.Models
             return filter is AzureKubernetesServiceFilter;
         }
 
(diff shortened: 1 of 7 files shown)


src/Diagnostics.ModelsAndUtils/Models/Resource/AzureKubernetesService.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// Determines whether the Azure Kubernetes Service resource is applicable after filtering.
+        /// </summary>
+        /// <param name="filter">Resource Filter</param>
+        /// <returns>True, if resource passes the filter. False otherwise</returns>
+        public bool IsApplicable(IResourceFilter filter)
+        {
+            return filter is AzureKubernetesServiceFilter;
+        }
+
+        public AzureKubernetesService(string subscriptionId, string resourceGroup, string appName) : base()
```
ShekharGupta1988: lets call it 'resourceName' or 'name' instead of 'appName'

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Guage.cs
```diff
@@ -0,0 +1,231 @@
...
+
+                var diagData = new DiagnosticData()
+                {
+                    Table = table,
+                    RenderingProperties = new Rendering(RenderingType.Guage)
+                };
+                response.Dataset.Add(diagData);
+                return diagData;
+            }
+            catch (Exception ex)
```
ShekharGupta1988: What a special catch if you are just throwing the exception

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Guage.cs
```diff
@@ -0,0 +1,231 @@
...
+        /// </summary>
+        public Guage(InsightStatus status, double percentFilled, string displayValue, string label, GuageSize size, string description = "")
+        {
+            try
+            {
+                this.Status = status;
+                this.PercentFilled = percentFilled;
+                this.DisplayValue = displayValue;
+                this.Label = label;
+                if (size == GuageSize.Inherit) throw new Exception("Guage elements must have a size. Allowed size's are GuageSize.Small, GuageSize.Medium and GuageSize.Large. GuageSize.Inherit is allowed to use only with the Response.AddGuages extension method.");
```
ShekharGupta1988: Does this mean It will throw exception every time I create a Guage object with Inherit Size?

nmallick1: Yes, the inherit size is not supposed to be used with Guage object, it is supposed to be passed as a parameter to the AddGuages Response Extension method. I tried to mention its purpose in the comment. Is there a better way for me to achieve this ? Have a different enum perhaps ?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Guage.cs
nmallick1: I wasn't sure if the exception needs to be bubbled up and what the calling code is doing with it. I saw other extensions and they do not have a try in them. There is no specific action that I can take on an exception. Should I remove the try catch ?

nmallick1: removed the try catch block

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Guage.cs
```diff
@@ -0,0 +1,224 @@
...
+using System.Text;
+using System.Linq;
+using System.Data;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace Diagnostics.ModelsAndUtils.Models.ResponseExtensions
+{
+    
+
```
ShekharGupta1988: nitpick : remove empty line..

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Guage.cs
```diff
@@ -0,0 +1,224 @@
...
+        }
+
+
+        /// <summary>
+        /// Insight Level for the Guage. Decides the color of the Guage
+        /// </summary>
+        [JsonConverter(typeof(StringEnumConverter))]
+        public InsightStatus Status { get; set; }
+
+
```
ShekharGupta1988: same here. Remove empty line

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Guage.cs
```diff
@@ -0,0 +1,224 @@
...
+        ///         }
+        ///     }
+        ///     res.AddGuages(guages, GuageRenderDirection.Horizontal, GuageSize.Inherit);
+        ///     res.AddInsight(InsightStatus.Info, "More detailed info below");
+        ///     res.AddGuages(guages, GuageRenderDirection.Vertical, GuageSize.Large);
+        ///     return res;
+        /// }
+        /// </code>
+        /// </example>
+        public static DiagnosticData AddGuages(this Response response, List<Guage> guages, GuageRenderDirection renderDirection, GuageSize guageSize)
```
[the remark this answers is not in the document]
nmallick1: I was thinking of a scenario where someone would want the Warning / Critical gauge display bigger to grab attention whereas the rest remain small. An example is as follows

image

PR #105, opened 2019-02-07
Gitignore all appsettings files to avoid any risk of committing secrets

dagrooms52:
- Ignored all appsettings.{environment}.json files
- Ignored appsettings.json to eliminate all room for error (within reason) - this file is still tempting to drop keys into, and easy to forget or overlook in a commit


dagrooms52:
I pictured a self-documenting configuration file for the issue of knowing what settings the system expects. The only difference is that each module that pulls from configuration right now handles it in its own class, so the settings are spread out. If we pulled all configuration management into a central module, it would be clear what settings would be necessary to run the system.

The "Access a Secret" section of [this article](https://docs.microsoft.com/en-us/aspnet/core/security/app-secrets?view=aspnetcore-2.2&tabs=windows) shows how that would be accomplished in `Startup.ConfigureServices`

The appsettings.json is supposed to be the settings template, but it seems that some people are using it as their main configuration file anyway, which risks pushing keys to the remote repo.

dagrooms52:
@ShekharGupta1988 and I agreed to the following practice for protecting secrets in appsettings:
- appsettings.json will not be used by the app for configuration, and will remain blank as a template
- appsettings.{Environment}.json will be used to load configuration. This can contain secrets for the near future
- secrets.json will be utilized in the future to move all secrets outside of the repo, lowering the risk of committing secrets

src/Diagnostics.RuntimeHost/Diagnostics.RuntimeHost.csproj
```diff
@@ -1,7 +1,8 @@
-<Project Sdk="Microsoft.NET.Sdk.Web">
+﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp2.0</TargetFramework>
+    <UserSecretsId>79a3edd0-2092-40a2-a04d-dcb46d5ca9ed</UserSecretsId>
```
ShekharGupta1988: Do we need this UserSecretsID?

src/Diagnostics.DataProviders/Diagnostics.DataProviders.csproj
```diff
@@ -15,6 +15,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.2.0" />
```
ShekharGupta1988: Any reason for using this additional Package?

dagrooms52: This is to pass the IHostingEnvironment into the ConfigurationFactory constructor.

src/Diagnostics.RuntimeHost/Diagnostics.RuntimeHost.csproj
dagrooms52: I left this in from trying to configure secrets.json, it won't hurt to leave it or remove it. Do you want me to take it out?

PR #107, opened 2019-02-11
Fix/Disable MDM

hforeste:
- Disable MDM data provider for now
- Fix Registry key value names


src/Diagnostics.DataProviders/DataProviders.cs
```diff
@@ -18,7 +18,6 @@ public DataProviders(DataProviderContext context)
             Observer = new DataProviderLogDecorator(context, SupportObserverDataProviderFactory.GetDataProvider(_cache, context.Configuration));
             GeoMaster = new DataProviderLogDecorator(context, new GeoMasterDataProvider(_cache, context.Configuration.GeoMasterConfiguration));
             AppInsights = new DataProviderLogDecorator(context, new AppInsightsDataProvider(_cache, context.Configuration.AppInsightsConfiguration));
-            Mdm = new DataProviderLogDecorator(context, new MdmDataProvider(_cache, context.Configuration.MdmConfiguration, context.RequestId));
```
ShekharGupta1988: we are not initializing MDM decorator?

hforeste: No, I think we should disable for now until we run better tests for this

PR #110, opened 2019-02-20
Fix MDM and test E2E

crazyxy:


src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
```diff
@@ -186,15 +188,13 @@ public async Task<IEnumerable<DataTable>> GetMultipleTimeSeriesAsync(DateTime st
                 foreach (var s in samplingType)
                 {
                     var table = new DataTable();
-                    table.Columns.Add("Metric", typeof(string));
                     table.Columns.Add("TimeStamp", typeof(DateTime));
-                    table.Columns.Add(s.ToString());
+                    table.Columns.Add("Metric", typeof(string));
+                    table.Columns.Add(s.ToString(), typeof(double));
```
hforeste: Instead of `s.ToString()` shouldn't this be parsed as Double

src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
```diff
@@ -61,215 +61,215 @@ public DataProviderMetadata GetMetadata()
 
         #region AppInsight_DataProvider
 
-        public Task<DataTable> ExecuteAppInsightsQuery(string query)
```
hforeste: Please revert this file

src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
crazyxy: No. The first parameter is the column name so it should be should. The second parameter is the type of the column.

src/Diagnostics.DataProviders/DataProviderLogDecorator.cs
crazyxy: If I revert this file, the loading of detectors will timeout.

hforeste: This is the same code in production and we don't run into this. Lets sync up.

PR #112, opened 2019-02-26
Shared code

crazyxy:
* Github Worker for handling Gist and Detector
    * Modify EntityInvoker for compiling detectors with source reference
    * REST API for getting gist(s) and publishing package


src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
```diff
@@ -116,15 +182,65 @@ public async Task<IActionResult> GetInsights(string subscriptionId, string resou
             return await base.GetInsights(app, pesId, supportTopicId, startTime, endTime, timeGrain);
         }
 
+        /// <summary>
+        /// Publish detector package.
+        /// </summary>
+        /// <param name="pkg">The package.</param>
+        /// <returns>Task for publishing package.</returns>
         [HttpPost(UriElements.Publish)]
-        public async Task<IActionResult> PublishDetector(string subscriptionId, string resourceGroupName, string siteName, [FromBody] DetectorPackage pkg)
```
hforeste: >string subscriptionId, string resourceGroupName, string siteName,  [](start = 57, length = 66)

You will need this as the URI path requires it

ShekharGupta1988:
@crazyx - Yan, do you think we need to have some sort of namespaces for gists to separate them out for multiple products.  Though I am not saying that a particular gist in  one namespace cannot be used by any other detector, but it will give a nice logical separation between multiple teams.

There are 4 more teams using appLens infrastructure apart from our org.  

Detectors right now doesnt have that problem as they are separated by product type

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
```diff
@@ -116,15 +182,65 @@ public async Task<IActionResult> GetInsights(string subscriptionId, string resou
...
         }
 
-        private bool IsPostBodyMissing(DiagnosticSiteData postBody)
+        /// <summary>
+        /// Publish gist package.
+        /// </summary>
+        /// <param name="pkg">The package.</param>
+        /// <returns>Task for publishing package.</returns>
+        [HttpPost(UriElements.PublishGist)]
+        public async Task<IActionResult> PublishGistPackage([FromBody] GistPackage pkg)
```
hforeste: Same here. You will need subscriptionId, resourceGroupName, and siteName because of the Route[SiteResource] at the controller level

ShekharGupta1988:
So Your Gist can have something like (an example):

GistDefinition(
Name = "KustoQueryHelper",
Author="shgup",
Namespace="Microsoft.Web",        // Or "Web App(Windows)" , "Web App(Linux)", "Function App", "AKS"
...
)
public static class KustoQueryHelper{
.....
}

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
hforeste: I wonder if it would make sense to create a separate controller, something like GistController. As far as I can understand, Gist do not rely on subscription, resourceGroup, sitename, or hosting environment input. What do you think @ShekharGupta1988 ?

hforeste:
@ShekharGupta1988 , as far as your namespace ask. Since the gist files are just csx files, maybe we can reuse the resource filter attributes. What do you think?

src/Diagnostics.CompilerHost/Controllers/CompilerHostController.cs
```diff
@@ -29,24 +28,32 @@ public async Task<IActionResult> Post([FromBody]JToken jsonBody)
...
-            CompilerResponse compilerResponse = new CompilerResponse();
-            using (var invoker = new EntityInvoker(metaData, ScriptHelper.GetFrameworkReferences(), ScriptHelper.GetFrameworkImports()))
+            var metaData = new EntityMetadata(script);
+            var compilerResponse = new CompilerResponse();
+            using (var invoker = new EntityInvoker(metaData, ScriptHelper.GetFrameworkReferences(), ScriptHelper.GetFrameworkImports(), references.ToImmutableDictionary()))
             {
                 await invoker.InitializeEntryPointAsync();
                 compilerResponse.CompilationOutput = invoker.CompilationOutput;
                 compilerResponse.CompilationSucceeded = invoker.IsCompilationSuccessful;
+                compilerResponse.References = invoker.References;
```
hforeste: This is more of curiosity and a question, but how is the client using the references in the compilation response?

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
ShekharGupta1988: Yup. Its a good idea to create a GistController.

Ideally, we want a separation atleast at team level for Gists so that LogicApp is not messing up thr Web App gists or AKS gits.

So we can create a GistController with NameSpace as input ("Microsoft.Web" perhaps)

ShekharGupta1988:
> @ShekharGupta1988 , as far as your namespace ask. Since the gist files are just csx files, maybe we can reuse the resource filter attributes. What do you think?

Just had a chat with Yan regarding this. The concern is if we have a common place for all teams to publish gist without any separation, it could become really difficult to maintain and assign ownership.


Instead of using Resource Filter, its probably better to use RP Name (like Microsoft.Web or Microsoft.Logic) as the namespace here.

Imagine a scenario where both ASE and Site needs a CPU query from a Gist.

src/Diagnostics.CompilerHost/Controllers/CompilerHostController.cs
crazyxy: The client will use the references to figure out which gists the detector(or gist) loads.

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/GitHubWatcher.cs
```diff
@@ -1,105 +1,146 @@
...
+        /// Create or update a package.
+        /// </summary>
+        /// <param name="pkg">Detector package.</param>
+        /// <returns>Task for creating or updateing detector.</returns>
+        public override async Task<Tuple<bool, Exception>> CreateOrUpdatePackage(Package pkg)
         {
             if (pkg == null)
             {
-                return new Tuple<bool, Exception>(false, new ArgumentNullException("pkg"));
+                return new Tuple<bool, Exception>(false, new ArgumentNullException(nameof(pkg)));
```
hforeste: Somehow I feel this is over engineering, you can just throw the ArgumentNullException and not have to wrap it inside of some Tuple response.

crazyxy: I try to reduce the chagne of code. I will remove the redundant code.

hforeste:
> > @ShekharGupta1988 , as far as your namespace ask. Since the gist files are just csx files, maybe we can reuse the resource filter attributes. What do you think?
> 
> Just had a chat with Yan regarding this. The concern is if we have a common place for all teams to publish gist without any separation, it could become really difficult to maintain and assign ownership.
> 
> Instead of using Resource Filter, its probably better to use RP Name (like Microsoft.Web or Microsoft.Logic) as the namespace here.
> 
> Imagine a scenario where both ASE and Site needs a CPU query from a Gist.

I agree, having a separation is important. I think your example with the GistDefinition can definitely work but you'd have to maintain a static list of strings for resource provider to do string matches. Its not bad but we can achieve the same thing with resource filter and it less hassle since its already there. But i think either approaches can work.

ShekharGupta1988:
So the resource provider string is already there in the resource object in ResourceType member.
Also when we add a GenericARMController, ResourceProvider string will automatically come as a parameter to the controller.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -62,6 +67,35 @@ protected async Task<IActionResult> GetDetector(TResource resource, string detec
...
         }
 
+        protected async Task<IActionResult> ListGists()
+        {
+            await _sourceWatcherService.Watcher.WaitForFirstCompletion();
+
+            return Ok(_gistCache.GetAll()
+                .Select(p => new Dictionary<string, string>
+                {
+                    { "Id", p.Id },
```
ShekharGupta1988: would recommend adding a GistDefinition Model in the Diagnostics.Model class and use that here as well as in Authoring Experience.

src/Diagnostics.RuntimeHost/Utilities/FileHelper.cs
```diff
@@ -74,7 +74,7 @@ internal static bool IsDllorPdb(FileInfo fileInfo)
 
         internal static string SanitizeScriptFile(string val)
         {
-            var lines = val.Split(new[] { '\n' }).Where(p => p.StartsWith("#r ") || p.StartsWith("#load "));
```
ShekharGupta1988: This was added to Sanitize the script file coming from Local Development.
For Auto-Complete and intellisense, we add a reference to our framework dlls in Local Development template using #load and remove the #load while compiling against the server.

crazyxy: This will also remove the `#load` which is added by the detector author who wants to loads gists. Since the compiling service supports `#load`, we may not need to remove `#load` if we can provide a `../Framework/References/_frameworkRef.csx` in compiling service. Or we can only remove `#load "../Framework/References/_frameworkRef.csx"`.

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
crazyxy: Discussed offline. Keep the change.

crazyxy:
> So Your Gist can have something like (an example):
> 
> GistDefinition(
> Name = "KustoQueryHelper",
> Author="shgup",
> Namespace="Microsoft.Web", // Or "Web App(Windows)" , "Web App(Linux)", "Function App", "AKS"
> ...
> )
> public static class KustoQueryHelper{
> .....
> }

@ShekharGupta1988 I think we can reuse the definition but add a new attribute for gist.

```
[GistFilter(ResourceType=ResourceType.App)]
[Definition(Id = "<YOUR_DETECTOR_ID>", Name = "", Author = "<YOUR_ALIAS>", Description = "")]
public static class Helper{
    public static string Name = "a";
}
```

We can add more filters in the future.

crazyxy:
> So Your Gist can have something like (an example):
> 
> GistDefinition(
> Name = "KustoQueryHelper",
> Author="shgup",
> Namespace="Microsoft.Web", // Or "Web App(Windows)" , "Web App(Linux)", "Function App", "AKS"
> ...
> )
> public static class KustoQueryHelper{
> .....
> }

@ShekharGupta1988 I think we can use the resource filter, like

```
[LogicAppFilter]
[AzureKubernetesServiceFilter]
[Definition(Id = "<YOUR_GIST_ID>", Name = "", Author = "<YOUR_ALIAS>", Description = "")]
public static class HELPER {
    public static string Name = "a";
}
```

src/Diagnostics.RuntimeHost/Utilities/FileHelper.cs
ShekharGupta1988: So for the first option, Cindy tried not removing the #load earlier but then the process looked for the exact path to find framework csx in the backend which was not present. So it has to be removed. I liked your idea of not needing to remove it but it might involve extra work right now which is not really in the scope of this PR. 

A quick fix can be the second option you suggested which is removing the #load which ends with _frameworkRef.csx

ShekharGupta1988:
> > So Your Gist can have something like (an example):
> > GistDefinition(
> > Name = "KustoQueryHelper",
> > Author="shgup",
> > Namespace="Microsoft.Web", // Or "Web App(Windows)" , "Web App(Linux)", "Function App", "AKS"
> > ...
> > )
> > public static class KustoQueryHelper{
> > .....
> > }
> 
> @ShekharGupta1988 I think we can use the resource filter, like
> 
> ```
> [AzureKubernetesServiceFilter]
> [Definition(Id = "<YOUR_GIST_ID>", Name = "", Author = "<YOUR_ALIAS>", Description = "")]
> public static class HELPER {
>     public static string Name = "a";
> }
> ```

Thats also fine but its difficult to differentiate a gist with a detector. Maybe that distinction is not needed perhaps.

I liked your original way which was :

[GistFilter(ResourceTypes = ResourceType.AKS)]
[Definition(Id="",....]
public static class Helper{
...
....
}

crazyxy:
> > > So Your Gist can have something like (an example):
> > > GistDefinition(
> > > Name = "KustoQueryHelper",
> > > Author="shgup",
> > > Namespace="Microsoft.Web", // Or "Web App(Windows)" , "Web App(Linux)", "Function App", "AKS"
> > > ...
> > > )
> > > public static class KustoQueryHelper{
> > > .....
> > > }
> > 
> > 
> > @ShekharGupta1988 I think we can use the resource filter, like
> > ```
> > [AzureKubernetesServiceFilter]
> > [Definition(Id = "<YOUR_GIST_ID>", Name = "", Author = "<YOUR_ALIAS>", Description = "")]
> > public static class HELPER {
> >     public static string Name = "a";
> > }
> > ```
> 
> Thats also fine but its difficult to differentiate a gist with a detector. Maybe that distinction is not needed perhaps.
> 
> I liked your original way which was :
> 
> [GistFilter(ResourceTypes = ResourceType.AKS)]
> [Definition(Id="",....]
> public static class Helper{
> ...
> ....
> }

I think we can use package.json to differentiate a gist with a detector. Ideally, the gist package is same as the detector package, except for package.json. For gist, the package.json is:
```
{
...
"type":"Gist"
}
``` 
and for detector, the field is "Detector". This type determines which github worker is used to download the package and load to the cache.

ShekharGupta1988:
> > > > So Your Gist can have something like (an example):
> > > > GistDefinition(
> > > > Name = "KustoQueryHelper",
> > > > Author="shgup",
> > > > Namespace="Microsoft.Web", // Or "Web App(Windows)" , "Web App(Linux)", "Function App", "AKS"
> > > > ...
> > > > )
> > > > public static class KustoQueryHelper{
> > > > .....
> > > > }
> > > 
> > > 
> > > @ShekharGupta1988 I think we can use the resource filter, like
> > > ```
> > > [AzureKubernetesServiceFilter]
> > > [Definition(Id = "<YOUR_GIST_ID>", Name = "", Author = "<YOUR_ALIAS>", Description = "")]
> > > public static class HELPER {
> > >     public static string Name = "a";
> > > }
> > > ```
> > 
> > 
> > Thats also fine but its difficult to differentiate a gist with a detector. Maybe that distinction is not needed perhaps.
> > I liked your original way which was :
> > [GistFilter(ResourceTypes = ResourceType.AKS)]
> > [Definition(Id="",....]
> > public static class Helper{
> > ...
> > ....
> > }
> 
> I think we can use package.json to differentiate a gist with a detector. Ideally, the gist package is same as the detector package, except for package.json. For gist, the package.json is:
> 
> ```
> {
> ...
> "type":"Gist"
> }
> ```
> and for detector, the field is "Detector". This type determines which github worker is used to download the package and load to the cache.

Ok.. That also sound good. Lets go with the type field then.

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
crazyxy: I would like to keep the gist action here. We can use the resource type to filter gists.

src/Diagnostics.CompilerHost/Controllers/CompilerHostController.cs
```diff
@@ -29,24 +46,33 @@ public async Task<IActionResult> Post([FromBody]JToken jsonBody)
...
 
             if (string.IsNullOrWhiteSpace(script))
             {
                 return BadRequest("Missing script from body");
             }
 
-            EntityMetadata metaData = new EntityMetadata(script);
-            CompilerResponse compilerResponse = new CompilerResponse();
-            using (var invoker = new EntityInvoker(metaData, ScriptHelper.GetFrameworkReferences(), ScriptHelper.GetFrameworkImports()))
+            var references = JsonConvert.DeserializeObject<Dictionary<string, string>>(jsonBody["reference"].ToString());
```
hforeste: `jsonBody[reference]` could trigger KeyNotFoundException. Check for this empywhitespace as you are with the script.

hforeste: Can you add an example json that has these references? Just want to know how it looks and what it is.

hforeste:
> > > > > So Your Gist can have something like (an example):
> > > > > GistDefinition(
> > > > > Name = "KustoQueryHelper",
> > > > > Author="shgup",
> > > > > Namespace="Microsoft.Web", // Or "Web App(Windows)" , "Web App(Linux)", "Function App", "AKS"
> > > > > ...
> > > > > )
> > > > > public static class KustoQueryHelper{
> > > > > .....
> > > > > }
> > > > 
> > > > 
> > > > @ShekharGupta1988 I think we can use the resource filter, like
> > > > ```
> > > > [AzureKubernetesServiceFilter]
> > > > [Definition(Id = "<YOUR_GIST_ID>", Name = "", Author = "<YOUR_ALIAS>", Description = "")]
> > > > public static class HELPER {
> > > >     public static string Name = "a";
> > > > }
> > > > ```
> > > 
> > > 
> > > Thats also fine but its difficult to differentiate a gist with a detector. Maybe that distinction is not needed perhaps.
> > > I liked your original way which was :
> > > [GistFilter(ResourceTypes = ResourceType.AKS)]
> > > [Definition(Id="",....]
> > > public static class Helper{
> > > ...
> > > ....
> > > }
> > 
> > 
> > I think we can use package.json to differentiate a gist with a detector. Ideally, the gist package is same as the detector package, except for package.json. For gist, the package.json is:
> > ```
> > {
> > ...
> > "type":"Gist"
> > }
> > ```
> > and for detector, the field is "Detector". This type determines which github worker is used to download the package and load to the cache.
> 
> Ok.. That also sound good. Lets go with the type field then.

Thinking about Shekhar's point with the `GistFilter`, I think that might be better compared to using the package.json. Its not that we can't use the package.json but I think we can just keep the package.json just for packages.

crazyxy:
> > > > > > So Your Gist can have something like (an example):
> > > > > > GistDefinition(
> > > > > > Name = "KustoQueryHelper",
> > > > > > Author="shgup",
> > > > > > Namespace="Microsoft.Web", // Or "Web App(Windows)" , "Web App(Linux)", "Function App", "AKS"
> > > > > > ...
> > > > > > )
> > > > > > public static class KustoQueryHelper{
> > > > > > .....
> > > > > > }
> > > > > 
> > > > > 
> > > > > @ShekharGupta1988 I think we can use the resource filter, like
> > > > > ```
> > > > > [AzureKubernetesServiceFilter]
> > > > > [Definition(Id = "<YOUR_GIST_ID>", Name = "", Author = "<YOUR_ALIAS>", Description = "")]
> > > > > public static class HELPER {
> > > > >     public static string Name = "a";
> > > > > }
> > > > > ```
> > > > 
> > > > 
> > > > Thats also fine but its difficult to differentiate a gist with a detector. Maybe that distinction is not needed perhaps.
> > > > I liked your original way which was :
> > > > [GistFilter(ResourceTypes = ResourceType.AKS)]
> > > > [Definition(Id="",....]
> > > > public static class Helper{
> > > > ...
> > > > ....
> > > > }
> > > 
> > > 
> > > I think we can use package.json to differentiate a gist with a detector. Ideally, the gist package is same as the detector package, except for package.json. For gist, the package.json is:
> > > ```
> > > {
> > > ...
> > > "type":"Gist"
> > > }
> > > ```
> > > and for detector, the field is "Detector". This type determines which github worker is used to download the package and load to the cache.
> > 
> > 
> > Ok.. That also sound good. Lets go with the type field then.
> 
> Thinking about Shekhar's point with the `GistFilter`, I think that might be better compared to using the package.json. Its not that we can't use the package.json but I think we can just keep the package.json just for packages.

We need package.json for Gist as well because Gist could also refer other gists files.

src/Diagnostics.CompilerHost/Controllers/CompilerHostController.cs
crazyxy: This will not trigger KeyNotFoundException since jsonBody is JObject. I will wrapper this code with try catch because it could fail when deserialize.

crazyxy: > Can you add an example json that has these references? Just want to know how it looks and what it is.

The references is actually a dictionary. Client will give init this value in `CompilerHostClient`.

ShekharGupta1988:
@crazyxy - Given the magnitude of the changes here, make sure you test this end to end with a private georegion service:

1. Basic compilation of a new detector
2. Publishing a new detector
3. Recompilation of old detector
4. Runtime loads with old detector (no breaking changes)

5. Also try out few detectors for different resources like Web App down for Linux and Windows, Function App Down for Function App, ASE detector.

@rekhaswaminathan did some compilation flow enhancements to not recompile an already complied script which is merged to master. Just merge that here and sync up with her to make sure those scenarios are also not broken :)

crazyxy:
> @crazyxy - Given the magnitude of the changes here, make sure you test this end to end with a private georegion service:
> 
> 1. Basic compilation of a new detector
> 2. Publishing a new detector
> 3. Recompilation of old detector
> 4. Runtime loads with old detector (no breaking changes)
> 5. Also try out few detectors for different resources like Web App down for Linux and Windows, Function App Down for Function App, ASE detector.
> 
> @rekhaswaminathan did some compilation flow enhancements to not recompile an already complied script which is merged to master. Just merge that here and sync up with her to make sure those scenarios are also not broken :)

@ShekharGupta1988 

 - [x] Basic compilation of a new detector
 - [x] Publishing a new detector
 - [x] Recompilation of old detector
 - [x] Runtime loads with old detector (no breaking changes)
 - [x] Try out few detectors for different resources like Web App down for Linux and Windows, Function App Down for Function App, ASE detector.
 - [ ] <s>Verify after merging @rekhaswaminathan 's change. To verify her change, we need to wait for her to check in the front-end change.</s>

PR #113, opened 2019-02-26
Forms

rekhaswaminathan:
- Changes to Controllers to accept FormContext query params

- ResponseExtension methods for authors to use Forms 

- Compilation optimization by checking for script-etag.


stevenernst131:
Can you add a screenshot of what this could look like? I know the forms are probably customizable but just any example works.

rekhaswaminathan:
Added screenshots in the UI project: 
https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/pull/237

src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
```diff
@@ -48,7 +48,12 @@ public class OperationContext<TResource> where TResource : IResource
         /// </summary>
         public string TimeGrain { get; private set; }
 
-        public OperationContext(TResource resource, string startTimeStr, string endTimeStr, bool isInternalCall, string requestId, string timeGrain = "5", SupportTopic supportTopic = null)
+        /// <summary>
+        /// ExecuteForm object
+        /// </summary>
+        public FormContext Form { get; private set; }
```
stevenernst131: If you can add multiple forms in one detector, would there be multiple form context objects that would need to be passed?

rekhaswaminathan: FormContext contains data of the form that is executing. A user can click only one button at at time. So at any point of time, there will be only one FormContext object in a detector.

src/Diagnostics.ModelsAndUtils/Models/CompilerResponse.cs
```diff
@@ -8,10 +8,14 @@ public class CompilerResponse
...
-        public IEnumerable<string> CompilationOutput;
+        public IEnumerable<string> CompilationTraces;
 
         public string AssemblyBytes;
 
         public string PdbBytes;
+
+        public string AssemblyName;
+       
+        public bool IsCompiled;
```
ShekharGupta1988: Should we call this CompilationStateChanged ?
Alternatively, do we need this member. We can also deduce this using script etag from response header

src/Diagnostics.ModelsAndUtils/Models/Response.cs
```diff
@@ -38,6 +38,11 @@ public class Response
         /// </summary>
         public List<AzureSupportCenterInsight> AscInsights { get; set; }
 
+        /// <summary>
+        /// List of detector forms - DO NOT ADD TO THIS DIRECTLY
+        /// </summary>
+        public List<Form> DetectorForms { get; set; }
```
ShekharGupta1988: Why do we need a different member object for Forms? This should be part of DataSet, no?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Form.cs
```diff
@@ -0,0 +1,271 @@
...
+
+        /// <summary>
+        /// Represents the label of the input 
+        /// </summary>
+        public string Label;
+
+        /// <summary>
+        /// Combines the parent form id and input id
+        /// </summary>
+        public string CombinedId;
```
ShekharGupta1988: should we call this as InternalId?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Form.cs
```diff
@@ -0,0 +1,271 @@
...
+    {
+        /// <summary>
+        /// Creates an instance of Textbox class
+        /// </summary>
+        /// <param name="id">Unique id for the input</param>
+        /// <param name="label">Label for the textbox</param>
+        /// <param name="isRequired">Indicates if it is a required input</param>
+        public Textbox(int id, string label = "", bool isRequired = false): base(id, InputTypes.TextBox, label, isRequired)
+        {
+
```
ShekharGupta1988: nit: get rid of the empty line

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Form.cs
```diff
@@ -0,0 +1,271 @@
...
+    /// </example>
+    public class Textbox: FormInput
+    {
+        /// <summary>
+        /// Creates an instance of Textbox class
+        /// </summary>
+        /// <param name="id">Unique id for the input</param>
+        /// <param name="label">Label for the textbox</param>
+        /// <param name="isRequired">Indicates if it is a required input</param>
+        public Textbox(int id, string label = "", bool isRequired = false): base(id, InputTypes.TextBox, label, isRequired)
```
ShekharGupta1988: isnt Label mandatory instead of optional?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Form.cs
```diff
@@ -0,0 +1,271 @@
...
+        public void AddFormInputs(List<FormInput> inputs)
+        {
+            inputs.ForEach(input =>
+            {
+                AddFormInput(input);
+            });
+        }
+    }
+
+    public abstract class FormInput
```
ShekharGupta1988: nit: Calling it FormInputBase explicitly calls out a abstract base class.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Form.cs
rekhaswaminathan: Yes true. Will change it.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Form.cs
```diff
@@ -0,0 +1,271 @@
...
+        /// <param name="label">Label to display on the button</param>
+        /// <param name="buttonStyle">Bootstrap button style for the button</param>
+        public Button(int id, string label = "", ButtonStyles buttonStyle = ButtonStyles.Primary): base(id, InputTypes.Button, label, false)
+        {
+            this.ButtonStyle = buttonStyle;
+        }
+
+    }
+
+    public enum InputTypes
```
ShekharGupta1988: Call it "FormInputTypes"

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Form.cs
```diff
@@ -0,0 +1,271 @@
...
+    {
+        TextBox = 0,
+        Checkbox,
+        RadioButton,
+        DropDown,
+        Button
+    }
+
+    public enum ButtonStyles
```
ShekharGupta1988: nitpick : Should we call this as StyleClass instead of ButtonStyles.
I see this getting used at multiple places and not only in just forms. So maybe taking this enum out into a separate file makes sense

src/Diagnostics.ModelsAndUtils/Models/Response.cs
rekhaswaminathan: So the main purpose of having a different member DetectorForms is to validate all the forms per detector in the response sent back to the UI (Refer to ValidateForms method in DiagnosticControllerBase.cs). Forms are added part of the DataSet by using res.AddForm or res.AddForms - here we can only validate individual forms but not all the forms per detector. So by having a different member DetectorForms - we can perform validations across Forms such as each form should have unique id

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Form.cs
ShekharGupta1988: As we discussed, this actually can be removed from the class and can be created on the UI side as its specific to the UI

src/Diagnostics.RuntimeHost/Models/FormModelBinder.cs
```diff
@@ -0,0 +1,81 @@
...
+                });
+            }
+            bindingContext.Result = ModelBindingResult.Success(result);
+            return Task.CompletedTask;
+        }
+
+        private bool IsInt(string num)
+        {
+            int id = 0;
+            return int.TryParse(num, out id);
```
ShekharGupta1988: you can also simply do return int.TryParse(num, out int id);

src/Diagnostics.ModelsAndUtils/Models/CompilerResponse.cs
rekhaswaminathan: As discussed, I will remove the IsCompiled property and the server will always return script etag.. The UI will check if it has changed by comparing the previous etag and the current etag and then update the assembly name.

src/Diagnostics.Scripts/CompilationService/Utilities/ScriptCompilation.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// Checks if the given assembly is already loaded
+        /// </summary>
+        public static bool IsAssemblyLoaded(string givenAssembly, out Assembly loadedAssembly)
+        {
+            loadedAssembly = null;
+            if(string.IsNullOrEmpty(givenAssembly))
+            {
+                return false;
+            }
+            Assembly[] assemblies = AppDomain.CurrentDomain.GetAssemblies();
```
ShekharGupta1988: Little bit concerned about this if we have to call this every time on compilation request from performance point of view.

Orthogonal to this, you can also create a in-process memory cache using dotnet core singleton service where you keep all the assemblies loaded in the cache (top 100 by time) and use that for optimization.

src/Diagnostics.Scripts/CompilationService/Utilities/ScriptCompilation.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </summary>
+        public static bool IsSameScript(string script, string scriptHash)
+        {
+            if(string.IsNullOrEmpty(script))
+            {
+                return true;
+            }
+            byte[] oldHash = Convert.FromBase64String(scriptHash);
+            byte[] newHash = GetHashFromScript(script);
+            bool areEqual = false;
```
ShekharGupta1988: you can simply do:

return oldHash.SequenceEqual(newHash);

src/Diagnostics.Scripts/CompilationService/Utilities/ScriptCompilation.cs
rekhaswaminathan: Trying to understand this so correct me if I am wrong. Is this something similar to InvokeCache which gets Configured in Startup.cs? So I'd be creating a new service will cache the assemblies loaded and use that service to check if the assembly is present is the cache.

ShekharGupta1988: Yes.. correct. A dotnet service runs in a singleton fashion and can be used by various controllers/components by simple dependency injection.
So You can create a service which hold latest 100 (or some number) of assemblies and check against that.

ShekharGupta1988:
@rekhaswaminathan  - do add unit tests for response extension that you have written also. Check the test project

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Form.cs
rekhaswaminathan: Shouldnt it be StyleClasses then? And also do you mean moving to a separate file under ResponseExtension namespace?

src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
```diff
@@ -51,9 +52,9 @@ public class OperationContext<TResource> where TResource : IResource
         /// <summary>
         /// ExecuteForm object
```
ShekharGupta1988: nitpick : fix the description

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Utilities.cs
```diff
@@ -198,5 +199,29 @@ public static ExecuteFormInput GetFormInput(FormContext form, int inputId)
...
+        /// Gets the button id of the button currently executing
+        /// </summary>
+        public static int GetExecutingButtonId(Form form)
+        {
+            try
+            {
+                var executingButtonId = form.FormInputs.Find(input => input.InputType == FormInputTypes.Button);
+                return executingButtonId.InputId;
+            }
+            catch(Exception e)
```
ShekharGupta1988: nitpick : you dont need the catch block if there is no exception handling.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Form.cs
ShekharGupta1988: Correct. 
You canuse them right now with button but the same set of bootstrap classes can be used with multiple other UI components that Bootstrap provides.

src/Diagnostics.RuntimeHost/Services/AssemblyCacheService.cs
```diff
@@ -0,0 +1,89 @@
...
+            AssemblyCache = new Dictionary<string, Assembly>();
+            InitializeCache();
+        }
+
+        /// <summary>
+        /// Initializes the cache with currently loaded assemblies
+        /// </summary>
+        public void InitializeCache()
+        {
+            Assembly[] currentAssemblies = AppDomain.CurrentDomain.GetAssemblies();
```
ShekharGupta1988: Why is this step needed? We dont want to add all the loaded assemblies in the cache. The Loaded assemblies will comprise of published detectors mostly.
I would say init the cache as empty to begin with

src/Diagnostics.RuntimeHost/Services/AssemblyCacheService.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// </summary>
+        bool IsAssemblyLoaded(string assemblyName, out Assembly loadedAssembly);
+    }
+
+    public class AssemblyCacheService: IAssemblyCacheService
+    {
+        /// <summary>
+        /// Queue to maintain top <see cref="MaxQueueSize"/> assemblies
+        /// </summary>
+        Queue<string> AssemblyQueue;
```
ShekharGupta1988: Use a ConcurrentQueue instead of a normal Queue as the later is not thread-safe.

ShekharGupta1988: Also qualify the members as private

PR #117, opened 2019-02-28
Basic Solutions Component

dagrooms52:
* Pre-built restart solution allows a full app restart over ARM API
* Solutions can only be added to Insights currently
* Solutions aren't very customizable currently, but I would like to add the pre-built Restart solution to a few detectors to get early feedback from customers


src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
```diff
@@ -0,0 +1,167 @@
...
+                $"{SolutionConstants.RestartInstructions}";
+
+            return new Solution("Restart Site", resourceUri, isInternal, ActionType.RestartSite, instructions,
+                new string[] { SolutionConstants.AppRestartDescription }, confirm: true, actionName: "Restart App");
+        }
+
+        /// <summary>
+        /// Creates a pre-defined Solution component capable of restarting an App instance
+        /// </summary>
+        public static Solution Restart(string resourceUri, OperationContext context, string detectorId)
```
dagrooms52: Does anyone know of a way to provide the OperationContext and Detector metadata to this component without relying on the detector author to provide them? They will be consistent in every Solution case, and should be taken care of without the detector author's help.

ShekharGupta1988: For Detector Metadata, it is actually part of the Response object that is passed to the script.
We populate the Definition object from the Entry Point (Run method) before invoking the csx.

image

So you should be ale to use that in the extension method.

However, this is not true for OperationContext. So it looks like you would have to pass that explicitly.

src/Diagnostics.ModelsAndUtils/Models/Resource/ApiManagementService.cs
```diff
@@ -22,6 +23,16 @@ public class ApiManagementService : ApiManagementServiceFilter, IResource
         /// </summary>
         public string ResourceGroup { get; set; }
 
+        /// <summary>
+        /// Resource URI
+        /// </summary>
+        public string ResourceUri {
```
ShekharGupta1988: Can you abstract these out in a base class for Resource which Implements an IResource.

ShekharGupta1988: actually never mind.  The resource class inherits members from filter class, so not sure if it would be possible

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -152,9 +152,9 @@ public override async Task<string> GetStampName(string subscriptionId, string re
             JToken obj2 = ((JArray)siteObjects)
                     .Select(i => (JObject)i)
                     .FirstOrDefault(
-                        j => (j.ContainsKey("subscription") && j["subscription"]["name"].ToString() == subscriptionId
-                            && j.ContainsKey("resource_group_name") && j["resource_group_name"].ToString() == resourceGroupName
-                            && j.ContainsKey("stamp")));
+                        j => j.ContainsKey("subscription") && j["subscription"]["name"].ToString().ToLower() == subscriptionId.ToLower()
```
crazyxy: nit: string.Equal("", "", StringComparison.OrdinalIgnoreCase)

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -152,9 +152,9 @@ public override async Task<string> GetStampName(string subscriptionId, string re
...
+                            && j.ContainsKey("resource_group_name") && j["resource_group_name"].ToString().ToLower() == resourceGroupName.ToLower()
```
crazyxy: same here

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
dagrooms52: Would it make sense to add OperationContext in the Response object? Or since all I need is the internal/external flag, would it make sense to add that to the Response?
I imagine this isn't the only component that changes its rendering depending on internal/external view, so it's possible we could cut out a lot of extra work for detector authors on other components as well.

src/Diagnostics.ModelsAndUtils/Utilities/UriUtilities.cs
```diff
@@ -0,0 +1,23 @@
...
+    public static class UriUtilities
+    {
+        public static string BuildResourceUri(string subscriptionId, string resourceGroup, string resourceName,
+            string provider = "Microsoft.Web", string resourceTypeName = "sites")
+        {
+            return $"/subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/" +
+                $"providers/{provider}/{resourceTypeName}/{resourceName}";
+        }
+        
+        public static string BuildDetectorLink(string resourceUri, string detectorId)
```
crazyxy: Do we need to use `Uri` here?

src/Diagnostics.ModelsAndUtils/Utilities/StringExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.Utilities
+{
+    public static class StringExtensions
+    {
+        public static string Truncate(this string value, int maxLength)
+        {
+            if (string.IsNullOrEmpty(value))
```
crazyxy: string.IsNullOrWhitespace?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
```diff
@@ -0,0 +1,167 @@
...
+            Dictionary<string, object> actionArgs)
+        {
+            return UpdateAppSettings(resourceUri, context.IsInternalCall, detectorId, actionArgs);
+        }
+    }
+
+    /// <summary>
+    /// Actions that a Solution can perform, such as ARM API requests.
+    /// </summary>
+    public enum ActionType
```
crazyxy: Private or internal? I don't find other references outside this file.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
dagrooms52: Oh right, forgot C# has a helper for everything

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
ShekharGupta1988: So the way we differentiate between UI Views for Internal/External is in the client code actually. 
I guess I didnt understand the current flow : The backend only communicates to the frontend that there is a restart solution and the UI has selected views (and descriptions) or the Backend makes the decision for internal/external and sends different description/data to UI

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
dagrooms52: Detector authors will have to be able to create ActionTypes

src/Diagnostics.ModelsAndUtils/Utilities/Constants.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.Utilities
+{
+    public static class SolutionConstants
+    {
+        public static readonly string AppRestartDescription = @"
+ #### App Restart
```
ShekharGupta1988: Do you think its better to have this content somewhere outside the backend project because any small change in this would need a full backend deployment in production. 

I can think of following options:
1) Having it be stored as Github doc and backend reads and caches the doc (based on etag changes). This is a longer route however will give you most flexibility for future changes.
2) Having them in the UI project : Its a shorter route right now and to change the description in future, it would need a UI deployment which is much cheaper to do.

src/Diagnostics.ModelsAndUtils/Utilities/StringExtensions.cs
dagrooms52: Whitespace can go over the length limit so it can be truncated, and I'd prefer not to mess with whitespace values if anyone wanted to use this function to truncate them. It would be up to the caller to trim their own whitespace.

src/Diagnostics.ModelsAndUtils/Utilities/Constants.cs
dagrooms52: Good point, I'll move them to the frontend for now.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
crazyxy: You need to move it to a separate file.

crazyxy: > Detector authors will have to be able to create ActionTypes

You need to move it to a separate file.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
stevenernst131: There is probably a way to remove "IsInternal" from the response object in general and then pass all the information you need to display either view (which it looks like you are doing) and have the UI know whether it is internal or not and display accordingly. The logic would need to be added in the UI but you could do something like the detector container level provides a service that has some information like this that is common for the whole detector, and then it can access it from the solution component. I also think detector-control service has access to this information already.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
dagrooms52: I originally did that, and the Detector editor was unable to find the Enum namespace. Is there a way of importing it that won't be confusing to Detector authors?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
dagrooms52: @stevenernst131 I would need some more specifics on how to go that route. It does sound better in the long term to leave the Response object as simple as possible. Do you think the current implementation is fine to get this PR through?

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
crazyxy: You can add the reference [here](https://github.com/Azure/Azure-AppServices-Diagnostics/blob/97977ec0643ca86b6ac40a300ade563c78ef88e9/src/Diagnostics.ModelsAndUtils/ScriptUtilities/ScriptHelper.cs#L25).

ShekharGupta1988:
@dagrooms52  - Make sure to test these changes with existing pre-compiled detectors to ensure no breaking changes are there.

PR #119, opened 2019-03-01
Add headers to start with 'diag-'

rekhaswaminathan:
Changed headers to start with diag-

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -173,15 +173,15 @@ namespace Diagnostics.RuntimeHost.Controllers
             };
 
             string scriptETag = string.Empty;
-            if (Request.Headers.ContainsKey("script-etag"))
+            if (Request.Headers.ContainsKey("diag-script-etag"))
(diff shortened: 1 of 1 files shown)


src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -212,7 +212,7 @@ protected async Task<IActionResult> ExecuteQuery<TPostBodyResource>(TResource re
                         _assemblyCacheService.AddAssemblyToCache(tempAsm.FullName, tempAsm);
                     }
 
-                    Request.HttpContext.Response.Headers.Add("script-etag", Convert.ToBase64String(ScriptCompilation.GetHashFromScript(jsonBody.Script)));
+                    Request.HttpContext.Response.Headers.Add("diag-script-etag", Convert.ToBase64String(ScriptCompilation.GetHashFromScript(jsonBody.Script)));
```
ShekharGupta1988: Are you not passing assembly name in response header ?

rekhaswaminathan: It is passed in the response of compilation output.

PR #129, opened 2019-03-20
Insight friendly name update for asc

nmallick1:
Updated InsightFriendlyName for ASC Insight to reflect the name of a detector instead of an insight title.


src/Diagnostics.ModelsAndUtils/Utilities/AzureSupportCenterInsightUtilities.cs
```diff
@@ -49,7 +49,7 @@ internal static AzureSupportCenterInsight CreateInsight<TResource>(string title,
                 Id = GetDetectorGuid(title),
                 Title = title,
                 ImportanceLevel = (ImportanceLevel)Enum.Parse(typeof(ImportanceLevel), ((int)status).ToString()),
-                InsightFriendlyName = title,
+                InsightFriendlyName = category,
```
ShekharGupta1988: is this our categories like "Unknown", "Availability" etc or this is something else? and How is it used

nmallick1: Category is being computed as follows.

var category = detector.Name.Length > 32 ? context.Resource.ResourceTypeName : detector.Name

So most of the times it should be the detector name. I am not sure under what scenarios is the length actually > 32.

ShekharGupta1988: Got it. How is the InsightFriendlyName used? Will it break any logging expectations at ASC side

stevenernst131: What is the purpose of this change?

nmallick1: ASC was complaining that the Insight Friendly name is random and that was skewing their reporting. Ideally, the insight title should be constant and not change. Since that is difficult to enforce, Praveen came up with a middle ground with them where we show the detector name in insight friendly name so that their reporting does not go for a toss and they get consistent names. They still log an insight's title in other fields so there is no data loss either.

The change is to help satisfy ASC logging requirements.

ShekharGupta1988: Assuming you/Praveen has already talked to ASC folks about this, I am signing off :)
Thanks Nazeef for the explanation.

nmallick1: Thanks Shekhar :)

PR #133, opened 2019-03-26
 Log x-ms headers received to Kusto

rekhaswaminathan:
Screenshot of headers content logged:

image

### DIFF
--- a/src/Diagnostics.Logger/ApiMetricsLogger.cs
+++ b/src/Diagnostics.Logger/ApiMetricsLogger.cs
@@ -49,6 +49,16 @@ namespace Diagnostics.Logger
 
             endTime = DateTime.UtcNow;
             latencyInMs = Convert.ToInt64((endTime - startTime).TotalMilliseconds);
+            var headers = context.Request.Headers;
+            var headersContent = "Headers received: ";
+            foreach (var header in headers)
+            {
+                if (header.Key.StartsWith("x-ms", StringComparison.OrdinalIgnoreCase))
+                {
+                    headersContent += $"{header.Key} - ";
+                    headersContent += string.IsNullOrWhiteSpace(header.Value.FirstOrDefault()) ? string.Empty : "****";
+                }
+            }
 
             DiagnosticsETWProvider.Instance.LogRuntimeHostAPISummary(
                 requestId ?? string.Empty,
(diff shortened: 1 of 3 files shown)


src/Diagnostics.Logger/ApiMetricsLogger.cs
```diff
@@ -49,6 +49,16 @@ public void LogRuntimeHostAPIMetrics(HttpContext context)
...
             endTime = DateTime.UtcNow;
             latencyInMs = Convert.ToInt64((endTime - startTime).TotalMilliseconds);
+            var headers = context.Request.Headers;
+            var headersContent = "Headers received: ";
+            foreach (var header in headers)
+            {
+                if (header.Key.StartsWith("x-ms", StringComparison.OrdinalIgnoreCase))
+                {
+                    headersContent += $"{header.Key} - ";
+                    headersContent += string.IsNullOrWhiteSpace(header.Value.FirstOrDefault()) ? string.Empty : "****";
```
ShekharGupta1988: QQ - Have you considered the ease of writing an alert if we don't get x-ms-client-id header passed? 
I think the format in which we are writing this log message can be improved.

Should you log them as a JSON string like this which might make it easier to parse and extensible for future?
{
"requestHeaders" : [{"name": "x-ms-client-id", "value": "*****"}, {}, ...]
"someOtherFieldInFuture": "some Value"
}

rekhaswaminathan: I agree, I will make changes to log it a JSON string. 
To check if RunTimeHost received the x-ms-client-id header, we can check against the incoming Request.Headers. I was planning to follow up (while PR is in review) regarding how we configure alerts.

rekhaswaminathan: Changed to log headers as JSON string.
image

hforeste:
Hey @rekhaswaminathan, what is the context of this change?

rekhaswaminathan:
> Hey @rekhaswaminathan, what is the context of this change?

@hforeste For the change tracking project, we are making API calls to changeanalysis team's endpoint. One of the requirement is to send x-ms-client-object-id to changeanalysis endpoint. This header is coming to DiagRole from GeoMaster. So we are logging the headers received by DiagRole in Kusto and also planning to set up alerts if we do not receive x-ms-client-object-id header.

ShekharGupta1988:
Looks good @rekhaswaminathan 
One improvement I was thinking was to have a list of headers for which we can log their real values instead of '****'  (In case we want to see real values for few headers).

Nothing urgent though. We can handle that later.

PR #134, opened 2019-03-26
Solution Gist Integration

dagrooms52:
- Solutions can be written and imported to a Detector via Gists


src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
```diff
@@ -8,123 +8,96 @@ namespace Diagnostics.ModelsAndUtils.Models.ResponseExtensions
...
-        /// Creates a Solution component. Prefer a pre-defined Solution component such as <see cref="Solution.Restart()"/>.
+        /// Set the ResourceUri and optionally set the ActionOptions for the Solution to use.
         /// </summary>
-        /// <param name="title">Solution title.</param>
-        /// <param name="resourceUri">The URI of the target resource on which an action will be performed.</param>
-        /// <param name="isInternal">Reads <see cref="OperationContext"/> to determine if an internal view should be rendered.</param>
-        /// <param name="action">Denotes which action will be performed, such as a specific API call or navigation link.</param>
-        public Solution(string title, string resourceUri, ActionType action)
+        // TODO: actionOptions has to merge with existing options in case detector author adds to solution author's options
+        public Solution Using(string resourceUri, Dictionary<string, object> actionOptions = null)
```
rekhaswaminathan: have you added sample code documentation anywhere on how authors can write Solution?

dagrooms52: Not yet, I'm wondering where the best place would be for that since it will be entirely Gist-based.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
```diff
@@ -8,123 +8,96 @@ namespace Diagnostics.ModelsAndUtils.Models.ResponseExtensions
...
+        [JsonConverter(typeof(StringEnumConverter))]
+        public ActionType Action { get; set; }
 
         /// <summary>
-        /// If the solution requires confirmation, it will signify a potentially dangerous action in the UI.
+        /// Options pertaining to <see cref="ActionType.ArmApi"/>.
+        /// This is ignored if <see cref="Action"/> is not <see cref="ActionType.ArmApi"/>.
         /// </summary>
-        public bool RequiresConfirmation { get; set; }
+        public ArmApiOptions ApiOptions { get; set; }
```
puneetg1983: >ArmApiOptions [](start = 15, length = 13)

Do you think it is better to have just one Options class  as base class and have all these ArmOptions, BladeOptions, TabOptions derive from it? That way based on ActionType the right options is loaded. It also becomes easier for detector author to just pass Options and they wont have to remember each type.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
```diff
@@ -8,123 +8,96 @@ namespace Diagnostics.ModelsAndUtils.Models.ResponseExtensions
...
-        public ActionType Action { get; set; }
+        public string ResourceUri { get; set; }
 
         /// <summary>
-        /// Free-form JSON-serializable arguments that can be sent as the body of the action request if a request
-        /// body is applicable.
+        /// Instructions that will be sent to the customer by support staff. Rendered as Markdown.
         /// </summary>
-        public Dictionary<string, object> ActionArgs { get; set; }
+        public string InternalInstructions { get; set; }
```
puneetg1983: >InternalInstructions [](start = 22, length = 20)

Calling this InternalMarkdown makes it explicit that you can pass Markdown to this property

src/Diagnostics.ModelsAndUtils/ScriptUtilities/SolutionTypeTag.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.ScriptUtilities
+{
+    public enum SolutionTypeTag
```
puneetg1983: >SolutionTypeTag [](start = 16, length = 15)

SolutionType? More readable...

puneetg1983:
>         public static string BuildAzureResourceUri(string subscriptionId, string resourceGroup, string resourceName,

Can you add one more Override for this method that takes context object ? That way detector authors just have to pass the context and makes code cleaner for the author

---
Refers to: src/Diagnostics.ModelsAndUtils/ScriptUtilities/UriUtilities.cs:13 in 146df57. [](commit_id = 146df57236b3721df7dcf327d84f3e21c3e637e3, deletion_comment = False)

src/Diagnostics.ModelsAndUtils/ScriptUtilities/SolutionTypeTag.cs
dagrooms52: This is just to match `enum SolutionTypeTag` from the Angular project

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
dagrooms52: I don't think this fits into a polymorphic model, none of the objects share similar attributes or methods and they aren't acted on similarly.

dagrooms52:
> > ```
> >     public static string BuildAzureResourceUri(string subscriptionId, string resourceGroup, string resourceName,
> > ```
> 
> Can you add one more Override for this method that takes context object ? That way detector authors just have to pass the context and makes code cleaner for the author
> 
> Refers to: src/Diagnostics.ModelsAndUtils/ScriptUtilities/UriUtilities.cs:13 in [146df57](https://github.com/Azure/Azure-AppServices-Diagnostics/commit/146df57236b3721df7dcf327d84f3e21c3e637e3). [](commit_id = [146df57](https://github.com/Azure/Azure-AppServices-Diagnostics/commit/146df57236b3721df7dcf327d84f3e21c3e637e3), deletion_comment = False)

@puneetg1983 Github messed up your comment and I wasn't able to directly reply.

This method is already used on the `IResource` of `OperationContext`; each resource provides its resource URI as follows:

```
public string ResourceUri
{
    get
    {
        return UriUtilities.BuildAzureResourceUri(SubscriptionId, ResourceGroup, Name, Provider, ResourceTypeName);
    }
}
```

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Solution.cs
dagrooms52: It's possible we could have an "example Gist" that shows Solution documentation, once @crazyxy adds sub-folders for Gists and we make a Solutions folder.
For now I'll just be incorporating simple Solutions into existing Detectors, so nobody should have to write their own Solutions for a short while

PR #135, opened 2019-04-02
Add the flag to remove markdown border

cindylovescoding:
Add the flag to remove markdown border

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Markdown.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Markdown.cs
@@ -26,7 +26,33 @@ namespace Diagnostics.ModelsAndUtils.Models.ResponseExtensions
         /// }
         /// </code>
         /// </example>
-        public static DiagnosticData AddMarkdownView(this Response response, string markdown, string title = null, bool isContainerNeeded = true)
+        public static DiagnosticData AddMarkdownView(this Response response, string markdown, string title = null)
+        {
+            return response.AddMarkdownView(markdown, false, title);
+        }
+
+        /// <summary>
+        /// Adds markdown section to the response
+        /// </summary>
+        /// <param name="response">Response object</param>
+        /// <param name="isContainerNeeded">If true, will keep the container of markdown view</param>
+        /// <param name="markdown">String that will be translated to markdown in UI</param>
+        /// <param name="title">Title of markdown section</param>
+        /// <example> 
+        /// This sample shows how to use <see cref="AddMarkdownView"/> method.
+        /// <code>
+        /// public async static Task<![CDATA[<Response>]]> Run(DataProviders dp, OperationContext cxt, Response res)
+        /// {
+        ///     var markdown = @"
+        ///     ## This is header
+        ///     ";
+        ///     
+        ///     res.AddMarkdownView(markdown, "Title of markdown section");
+        /// }
+        /// </code>
+        /// </example>
+
+        public static DiagnosticData AddMarkdownView(this Response response, bool isContainerNeeded, string markdown, string title = null)
         {
             return response.AddMarkdownView(markdown, false, title, isContainerNeeded);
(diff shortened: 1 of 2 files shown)


src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Markdown.cs
```diff
@@ -51,7 +78,7 @@ public static DiagnosticData AddMarkdownView(this Response response, string mark
         /// }
         /// </code>
         /// </example>
-        public static DiagnosticData AddMarkdownView(this Response response, string markdown, bool enableEmailButtons, string title = null)
+        public static DiagnosticData AddMarkdownView(this Response response, string markdown, bool enableEmailButtons, string title = null, bool isContainerNeeded = true)
```
ShekharGupta1988: This is a breaking change as you are modifying the existing definition.

hforeste: Correct me if I'm wrong but adding the optional parameter does not break detectors calling this method.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Markdown.cs
```diff
@@ -31,13 +31,40 @@ public static DiagnosticData AddMarkdownView(this Response response, string mark
...
+        ///     var markdown = @"
+        ///     ## This is header
+        ///     ";
+        ///     
+        ///     res.AddMarkdownView(markdown, "Title of markdown section");
+        /// }
+        /// </code>
+        /// </example>
+
+        public static DiagnosticData AddMarkdownView(this Response response, bool isContainerNeeded, string markdown, string title = null)
```
hforeste: nitpik: Please make the parameter order consistent with the other overloaded method. `AddMarkdownView(this Response response, string markdown, bool isContainerNeeded, string title = null)`

hforeste: Maybe you can make `isContainerNeeded` an optional parameter: `isContainerNeeded = true`. That way people can just focus on the markdown content but if they want to remove the container they can specify `false`

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Markdown.cs
ShekharGupta1988: it will break as method signature has changed. Pre-compiled versions will break.

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Markdown.cs
cindylovescoding: This will throw an ambiguous build error : "error CS0121: The call is ambiguous between the following methods or properties: 'ResponseMarkdownExtension.AddMarkdownView(Response, string, string)' and 'ResponseMarkdownExtension.AddMarkdownView(Response, string, string, bool)'" if I make that optional as well..

hforeste: Oh okay then I'd recommend removing the parameter and if people want to remove the container they can use the other overloaded method. It would also be a good idea because changing this signature would break detectors.

hforeste: Oh okay then I'd recommend removing the parameter and if people want to remove the container they can use the other overloaded method. It would also be a good idea because changing this signature would break detectors.

cindylovescoding: Updated PR...

ShekharGupta1988: Did we update this @cindylovescoding ?

PR #137, opened 2019-04-03
Generic arm resource

nmallick1:
Support for generic ARM resource. This will enable us to onboard new services without the need of deploying the backend.


data/templates/Detector_OpenShiftManagedClusters.csx
```diff
@@ -0,0 +1,30 @@
+using System;
+using System.Threading;
+
+private static string GetQuery(OperationContext<ArmResource> cxt)
+{
+    return
+    $@
+        let startTime = datetime({cxt.StartTime});
+        let endTime = datetime({cxt.EndTime});
+        cluster('Aks').database('AKSprod').YOUR_TABLE_NAME
```
ShekharGupta1988: cluster('ClusterName').database('DBName')

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/ArmResourceFilter.cs
```diff
@@ -0,0 +1,14 @@
+﻿namespace Diagnostics.ModelsAndUtils.Attributes
```
ShekharGupta1988: nitpick : For new files, lets do the documentation as mandated. :)

src/Diagnostics.ModelsAndUtils/Models/Resource/ArmResource.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            get
+            {
+                return UriUtilities.BuildAzureResourceUri(SubscriptionId, ResourceGroup, Name, Provider, ResourceTypeName);
+            }
+        }
+
+        public bool IsApplicable(IResourceFilter filter)
+        {
+            return filter is ArmResource;
```
ShekharGupta1988: you can do the Provider Type and Resource Type string matching also here.

data/templates/Detector_OpenShiftManagedClusters.csx
nmallick1: Updated

src/Diagnostics.ModelsAndUtils/Models/Resource/ArmResource.cs
nmallick1: Makes sense, updated

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/ArmResourceFilter.cs
nmallick1: Updated documentation for the files that I am adding. :-)

src/Diagnostics.ModelsAndUtils/Models/Resource/ArmResource.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// Determines whether the current arm resource is applicable after filtering.
+        /// </summary>
+        /// <param name="filter">App Resource Filter</param>
+        /// <returns>True, if app resource passes the filter. False otherwise</returns>
+        public bool IsApplicable(IResourceFilter filter)
+        {
+            if (filter is ArmResource)
+            {
+                ArmResourceFilter armFilter = filter as ArmResourceFilter;
+                return ((string.Compare(armFilter.Provider, this.Provider, true) == 0) && (string.Compare(armFilter.ResourceTypeName, this.ResourceTypeName) == 0));
```
ShekharGupta1988: You might want to ignore case sensitive match, otherwise we will see "Microsoft.Sql" and "Microsoft.SQL" as two different resource types

PR #139, opened 2019-04-09
DataProvider for ChangeAnalysis RP

rekhaswaminathan:
- This PR only has changes related to DataProvider that queries for change data. I will not be merging  this to master immediately as it will change KustoService. I will merge only after I have completed the backend fully so it doesnt affect master branch.

-  Created TokenServiceBase class that can be used by both Kusto and ChangeAnalysis. In future, if we have another Service that requires AAD auth token, it can extend this class by providing the required AAD Config.

- Added methods to gather dependent resources from Kusto, get resource URIs for hostnames, get changesets for given ARM Resource URI and changes for given ChangeSetId and ResourceUri.


rekhaswaminathan:
@hforeste Added you here as I have changes to KustoTokenService. :)

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
```diff
@@ -0,0 +1,144 @@
...
+                HttpResponseMessage responseMessage = await httpClient.SendAsync(requestMessage, cancellationTokenSource.Token);
+                string content = await responseMessage.Content.ReadAsStringAsync();
+                if (!responseMessage.IsSuccessStatusCode)
+                {
+                    throw new Exception(content);
+                }
+
+                return content;
+            }
+            catch (Exception ex)
```
hforeste: I don't think you need a try..catch.. statement here since you are not doing anything with the exception

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// x-ms-client-object-id header to pass to Change Analysis endpoint.
+        /// </summary>
+        private string clientObjectIdHeader;
+
+        /// <summary>
+        /// For detectors loaded from Diagnose and Solve, pass x-ms-client-principal-name to Change Analysis endpoint.
+        /// </summary>
+        private string clientPrincipalNameHeader;
+
+        private readonly string changeAnalysisEndPoint = "https://changeanalysis-dataplane-dev.azurewebsites.net/providers/microsoft.changeanalysis/";
```
hforeste: The pattern we adopted for this project is load values like these as configuration values from DataProviderConfigurations. Add another property for endpoint in `ChangeAnalysisDataProviderConfiguration` and that value will be loaded from registry in production or the appsettings when run in other environments.

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
rekhaswaminathan: You are right. I think I should be logging the request in Kusto instead of doing nothing with it.

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
rekhaswaminathan: Sure thats a valid point.

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
hforeste: There is logging already happening at `DataProviderLogDecorator`. Unless you want to add any special logging here

src/Diagnostics.DataProviders/DataProviders/ChangeAnalysisDataProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        /// <summary>
+        /// Get all change sets for the given arm resource uri.
+        /// </summary>
+        /// <param name="armResourceUri">ARM Resource URI.</param>
+        /// <param name="startTime">Start time of time range.</param>
+        /// <param name="endTime">End time of the time range.</param>
+        /// <returns>List of changesets.</returns>
+        public async Task<List<ChangeSetResponseModel>> GetChangeSetsForResource(string armResourceUri, DateTime startTime, DateTime endTime)
+        {
```
hforeste: Detector owners will be calling this method, I suggest checking for null or empty with `isNullOrWhitespace` for the `armResourceUri`.

hforeste: Does the change analysis API have any limitations on starttime and endtime. Like do they only keep changelog information for the last 30 days? If we don't know that yet then its fine we can just move forward.

src/Diagnostics.DataProviders/DataProviders/ChangeAnalysisDataProvider.cs
```diff
@@ -0,0 +1,113 @@
...
+            return await changeAnalysisClient.GetChangeSetsAsync(request);
+        }
+
+        /// <summary>
+        /// Get all Changes for a given changesetId.
+        /// </summary>
+        /// <param name="changeSetId">ChangeSetId retrieved from <see cref="GetChangeSetsForResource(string, DateTime, DateTime)"/>.</param>
+        /// <param name="resourceUri">ARM Resource Uri.</param>
+        /// <returns>List of changes.</returns>
+        public async Task<List<ResourceChangesResponseModel>> GetChangesByChangeSetId(string changeSetId, string resourceUri)
```
hforeste: Can you have overloaded method that just takes `ChangeSetResponseModel`?

rekhaswaminathan: The idea is like this:
1. We get all change sets and display in UI in a way that would prompt user to click on each one of them
2. When the user clicks on a particular changeset, we fetch changes related to it. More like on-demand fetching changes. We have to keep in mind that each "changeset" can return 9-10 changes. 

Reasons behind this approach is:
Performance impact: If we query all changesets at once, it will cause throttling because ChangeAnalysis API needs to make 3-4 API calls at their end to fetch changes. The experience will be slow. 

So the idea is to get changesets separately and then fetch changes.

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
rekhaswaminathan: I just saw what the `DataProviderLogDecorator` does. I was not aware of this class before. I will add it

src/Diagnostics.DataProviders/DataProviders/ChangeAnalysisDataProvider.cs
rekhaswaminathan: There is a limitation I believe. I will check again with their team to find out what the range is and update the method to validate for that.

src/Diagnostics.DataProviders/TokenService/TokenServiceBase.cs
```diff
@@ -0,0 +1,113 @@
+﻿using Microsoft.IdentityModel.Clients.ActiveDirectory;
+using System.Threading.Tasks;
+using System;
+using System.Linq;
+using Diagnostics.Logger;
+
+namespace Diagnostics.DataProviders.TokenService
+{
+    public class TokenServiceBase
```
hforeste: I suggest making the class an `abstract` class and making all the properties abstract too to force any inheritors to provide values

src/Diagnostics.DataProviders/TokenService/TokenServiceBase.cs
```diff
@@ -0,0 +1,113 @@
+﻿using Microsoft.IdentityModel.Clients.ActiveDirectory;
```
hforeste: The abstraction looks good 👍

src/Diagnostics.DataProviders/TokenService/TokenServiceBase.cs
rekhaswaminathan: I am not sure if I understood this fully, so correct me if I'm wrong. 

>  making all the properties abstract too to force any inheritors to provide values

Which properties are you saying should be abstract? The idea is that derived classes should provide values that are needed for AAD such as `ClientCredential` , correct? 

I made the `AuthorizationToken` to have private set so that only the `StartTokenRefresh` can set the value and derived classes cannot modify this.

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
```diff
@@ -0,0 +1,144 @@
...
+            };
+            string jsonString = await PrepareAndSendRequest(requestUri, postBody);
+            List<ChangeSetResponseModel> changeSetsResponse = JsonConvert.DeserializeObject<List<ChangeSetResponseModel>>(jsonString);
+            return changeSetsResponse;
+        }
+
+        /// <inheritdoc/>
+        public async Task<ResourceIdResponseModel> GetResourceIdAsync(List<string> hostnames, string subscription)
+        {
+            string requestUri = changeAnalysisEndPoint + "resourceId?api-version=2019-04-01-preview";
```
ShekharGupta1988: Have the api-version also defined in Config so that when change analysis team goes GA, we can simply update the api-version config.

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// </summary>
+        private string clientPrincipalNameHeader;
+
+        private readonly string changeAnalysisEndPoint = "https://changeanalysis-dataplane-dev.azurewebsites.net/providers/microsoft.changeanalysis/";
+
+        private readonly Lazy<HttpClient> client = new Lazy<HttpClient>(() =>
+        {
+            var client = new HttpClient();
+            client.DefaultRequestHeaders.Accept.Clear();
+            client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
```
ShekharGupta1988: Is there any default timeout here

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
rekhaswaminathan: Sure, will do that.

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
rekhaswaminathan: There isnt currently. But I can add that.

rekhaswaminathan: I just checked it again, there is a timeout and the value is taken from `DataProviderConstants.DefaultTimeoutInSeconds`

src/Diagnostics.DataProviders/DataProviderContext.cs
```diff
@@ -11,7 +11,16 @@ public class DataProviderContext
         public string RequestId { get; private set; }
         public CancellationToken DataSourcesCancellationToken { get; private set; }
 
-        public DataProviderContext(DataSourcesConfiguration dataSourceConfiguration, string requestId = null, CancellationToken dataSourceCancellationToken = default(CancellationToken))
+        /// <summary>
+        /// Value of x-ms-client-object-id header received for requests coming from 'Diagnose and Solve' and Applens AppId for requests from Applens.
+        /// </summary>
+        public string clientObjectId {get; private set; }
```
hforeste: Because this is specific to the ChangeAnalysisDataProvider, I would suggest putting these properties into `ChangeAnalysisDataProviderConfiguration`

rekhaswaminathan: `ChangeAnalysisDataProviderConfiguration` has configs for entire environment, right? These are not configs. These are detector specific. The values depend on where the detector is called from Applens vs Diagnose and Solve. I saw that `DataProviderContext` gets the value based on HttpContext. So I decided to keep them here

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
```diff
@@ -0,0 +1,186 @@
...
+        /// <param name="subscriptionId">Subscription Id.</param>
+        public async Task<SubscriptionOnboardingStatus> CheckSubscriptionOnboardingStatus(string subscriptionId)
+        {
+            string requestUri = changeAnalysisEndPoint + $"Subscription/{subscriptionId}";
+            try
+            {
+               string jsonString = await PrepareAndSendRequest(requestUri, httpMethod: HttpMethod.Get);
+               return JsonConvert.DeserializeObject<SubscriptionOnboardingStatus>(jsonString);
+            }
+            catch (Exception)
```
ShekharGupta1988: We should log the exception here.

ShekharGupta1988: Use the data provider events here

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
```diff
@@ -0,0 +1,186 @@
...
+            {
+               requestMessage.Content = new StringContent(JsonConvert.SerializeObject(postBody), Encoding.UTF8, "application/json");
+            }
+
+            CancellationTokenSource cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(DataProviderConstants.DefaultTimeoutInSeconds));
+            HttpResponseMessage responseMessage = await httpClient.SendAsync(requestMessage, cancellationTokenSource.Token);
+            string content = await responseMessage.Content.ReadAsStringAsync();
+            if (!responseMessage.IsSuccessStatusCode)
+            {
+                throw new Exception(responseMessage.ReasonPhrase);
```
ShekharGupta1988: If you are throwing an exception here, consider following:
- Use an HttpException
- Add Details in the exception like Status Code (Reason Phrase) , Content so that its easy to debug

src/Diagnostics.DataProviders/DataProviders/ChangeAnalysisDataProvider.cs
```diff
@@ -0,0 +1,173 @@
...
+            if (string.IsNullOrWhiteSpace(armResourceUri))
+            {
+                throw new ArgumentNullException(nameof(armResourceUri));
+            }
+
+            // Validation for date range as the data retention policy is 14 days.
+            DateTime maxDataRetentionDate = DateTime.Now.AddDays(-14);
+            if (startTime < maxDataRetentionDate)
+            {
+                throw new Exception("Changes beyond last 14 days are not available. Please provide dates within last 14 days");
```
ShekharGupta1988: ArgumentException

src/Diagnostics.DataProviders/DataProviders/ChangeAnalysisDataProvider.cs
```diff
@@ -0,0 +1,173 @@
...
+        {
+            // Query Kusto to get Dependencies
+            string query =
+                $@"
+                    let startTime = datetime({startTime});
+                    let endTime = datetime({endTime});
+                    let period = 30m;
+                    let sitename = ""{sitename}"";
+                    AntaresRuntimeWorkerSandboxEvents
+                    | where TIMESTAMP  < endTime and TIMESTAMP >= ago(20d) and Sandbox =~ sitename and EventId == 60011
```
ShekharGupta1988: Somewhere in the pipeline, do we ensure or throw the error that its not supported for Linux right now?

src/Diagnostics.DataProviders/DataProviders/ChangeAnalysisDataProvider.cs
```diff
@@ -0,0 +1,173 @@
...
+                    AntaresRuntimeWorkerSandboxEvents
+                    | where TIMESTAMP  < endTime and TIMESTAMP >= ago(20d) and Sandbox =~ sitename and EventId == 60011 
+                    | extend CustomProcessId = iff(ImagePath == 'w3wp.exe' , Pid , NewProcessId )
+                    | summarize by Role, RoleInstance, CustomProcessId , EventStampName, Tenant  
+                    | join (DNSQueryThirtyMinuteTable | extend CustomProcessId = Pid) 
+                     on Role, RoleInstance, EventStampName, CustomProcessId, Tenant
+                    | where TIMESTAMP > (startTime - period) and TIMESTAMP < endTime 
+                    | distinct QueryName 
+                  ";
+            DataTable kustoResultsTask = await kustoDataProvider.ExecuteQuery(query, stamp);
```
ShekharGupta1988: nitpick : its not a task. Consider renaming this to just kustoResults

ShekharGupta1988:
@rekhaswaminathan - consider adding unit tests for the data provider and response extension methods.

rekhaswaminathan:
> @rekhaswaminathan - consider adding unit tests for the data provider and response extension methods.

I had those created but didnt check those in yet .I will do soon

src/Diagnostics.DataProviders/DataProviders/ChangeAnalysisDataProvider.cs
rekhaswaminathan: We don't. I wasn't sure if we should really a throw an error because the query will run but won't give any results. So technically it doesnt "break" anything. I also thought when Linux team adds those tables in the future, we can tweak this method to check AppType. Let me know your thoughts.

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
rekhaswaminathan: Correct me if i'm wrong. The Dataprovider calls this method and DataProvider method is already wrapped inside `DataProviderLogDecorator`. Should I add additional logging here?

PR #141, opened 2019-04-16
Fix possible deadlock

hforeste:
- Add logging whenever we're missing swap history data. This will set us up for future alerting.
- Create Observer Token service like Kusto Token Service to account for the possible deadlock issue.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
+using Diagnostics.Logger;
 using Diagnostics.ModelsAndUtils.Models;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
@@ -14,7 +15,7 @@ namespace Diagnostics.DataProviders
     {
         private object _lockObject = new object();
 
-        public SupportObserverDataProvider(OperationDataCache cache, SupportObserverDataProviderConfiguration configuration) : base(cache, configuration)
+        public SupportObserverDataProvider(OperationDataCache cache, SupportObserverDataProviderConfiguration configuration, string requestId) : base(cache, configuration, requestId)
         {
         }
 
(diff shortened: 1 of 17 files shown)


src/Diagnostics.DataProviders/IWawsObserverTokenService.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Diagnostics.DataProviders
+{
+    public interface IWawsObserverTokenService
```
dagrooms52: Since this is defining the same interface as ISupportBayApiObserverTokenService, could we simplify to one interface or are they likely to diverge?

hforeste: So I did this because of the dependency injection in the [StartUp.cs](https://github.com/Azure/Azure-AppServices-Diagnostics/pull/141/files#diff-d567401b55ead5c2a4a3f4a20ed0fc7c). I have to specify some kind of type hence the two interfaces to delineate between two different token services that can be injected and used. See the [DiagnosticRequestMiddleware.cs](https://github.com/Azure/Azure-AppServices-Diagnostics/pull/141/files#diff-058205c71f6a0cc2f65ad7406829861a) on how I'm using it. If I had just specified `ObserverTokenService` as my service type when using `services.AddSingleton` in the StartUp.cs file then it will make it ambigious

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -61,14 +62,53 @@ public override async Task<dynamic> GetCertificatesInResourceGroupAsync(string s
...
-            return await GetRuntimeSiteSlotMapInternal(stampName, siteName);
+            return await GetRuntimeSiteSlotMapInternal(stampName, siteName, null);
         }
 
-        private async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMapInternal(string stampName, string siteName)
+        public override async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string stampName, string siteName, string slotName)
+        {
+            if (string.IsNullOrWhiteSpace(stampName))
+            {
+                throw new ArgumentNullException(stampName);
```
ShekharGupta1988: nameof(stampName) or "stampName".

Same for below exceptions also

src/Diagnostics.DataProviders/TokenService/TokenServiceBase.cs
```diff
@@ -1,44 +1,45 @@
-﻿using Microsoft.IdentityModel.Clients.ActiveDirectory;
-using System.Threading.Tasks;
+﻿// <copyright file="TokenServiceBase.cs" company="Microsoft Corporation">
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+// </copyright>
+
```
rekhaswaminathan: is this something we have to do in every new class?

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -200,7 +240,7 @@ public override HttpClient GetObserverClient()
         private async Task<string> Get(string path)
         {
             var request = new HttpRequestMessage(HttpMethod.Get, $"https://support-bay-api.azurewebsites.net/observer/{path}?api-version=2.0");
```
rekhaswaminathan: In appsettings.json we have a Support Observer endpoint, how this is one different from that?

src/Diagnostics.DataProviders/TokenService/TokenServiceBase.cs
hforeste: The styles are enforcing this. It will just create a warning I believe if we don't include it

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
hforeste: I can talk more about this with you offline but we have two services that we rely on and this is one of them

PR #146, opened 2019-04-19
Code cleanup and documentation

dagrooms52:


src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -92,56 +79,49 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
...
-                   kustoApiException != null ? kustoApiException.GetType().ToString() : string.Empty,
-                   kustoApiException != null ? kustoApiException.ToString() : string.Empty);
-            }
+            return dataSet?.Tables == null ? new DataTable() : dataSet.Tables.FirstOrDefault().ToDataTable();
+        }
 
-            
+        private void LogKustoQuery(string query, string cluster, string operationName, Stopwatch timeTakenStopWatch, string kustoClientId, Exception kustoApiException, DataTableResponseObjectCollection dataSet)
+        {
+            timeTakenStopWatch.Stop();
```
hforeste: I suggest just keeping this inside of the finally clause and let the `LogKustoQuery` function just be focused on the single responsibility of logging.

dagrooms52: Definitely, that slipped under my radar when I used the VS `refactor to function` operation

PR #150, opened 2019-04-30
Rekha/changeanalysis feature

rekhaswaminathan:
- API calls to submitscan request and checkscan progress
- ChangeSetId query param to detector to load changes and display in changes table


src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/ChangeAnalysis.cs
```diff
@@ -26,7 +26,7 @@ public static class ResponseChangeAnalysisExtension
         /// </example>
         public static DiagnosticData AddChangeSets(this Response response, List<ChangeSetResponseModel> changeSets)
         {
-            if (changeSets == null)
+            if (changeSets == null || changeSets.Count <= 0)
```
ShekharGupta1988: nit: you can also use .Any()

PR #152, opened 2019-05-01
Local host observer

hforeste:
Made some modifications to the project when targeting observer on local host
1. Disable token service for observer as the request to local host will just be HTTP requests
2. Add a feature flag behind local host observer so that we can turn on/off anytime. (This would require diagnostic role config update)

Note this also includes the changes from another pull requests https://github.com/Azure/Azure-AppServices-Diagnostics/pull/151.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviderConfigurations/SupportObserverDataProviderConfiguration.cs
+++ b/src/Diagnostics.DataProviders/DataProviderConfigurations/SupportObserverDataProviderConfiguration.cs
@@ -57,6 +57,9 @@ namespace Diagnostics.DataProviders
             }
         }
 
+        [ConfigurationName("ObserverLocalHostEnabled", DefaultValue = false)]
+        public bool ObserverLocalHostEnabled { get; set; }
+
         /// <summary>
         /// Gets or sets tenant to authenticate with.
         /// </summary>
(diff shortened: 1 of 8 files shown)


src/Diagnostics.DataProviders/DataProviderConfigurations/SupportObserverDataProviderConfiguration.cs
```diff
@@ -35,6 +35,12 @@ public SupportObserverDataProviderConfiguration()
         [ConfigurationName("IsMockConfigured", DefaultValue = false)]
         public bool IsMockConfigured { get; set; }
 
+        /// <summary>
+        /// Is the local host version of observer enabled.
+        /// </summary>
+        [ConfigurationName("ObserverLocalHostEnabled", DefaultValue = false)]
+        public bool ObserverLocalHostEnabled { get; set; }
```
rekhaswaminathan: does this require a change in appsettings.json to add the new config?

hforeste: Not really since the default will be false and for us that is okay but it wouldn't hurt to have it there in the appsettings though

rekhaswaminathan: I asked because sometimes when a property is missing in appsettings.json and we run runtimehost locally, we get a NullReferenceException for the missing config property.

hforeste: Yeah that's good thinking. Because this is bool and is a value type, the value will always be false. And that's what I want it to be for now. I made some changes to the appsettings to include this config. You will see it in the next commit

PR #156, opened 2019-05-07
Observer Logging

dagrooms52:
- Log information about Observer requests coming from RuntimeHost
- Code cleanup

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
@@ -1,10 +1,12 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
+using Diagnostics.Logger;
 using Diagnostics.ModelsAndUtils.Models;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
@@ -16,6 +18,7 @@ namespace Diagnostics.DataProviders
         protected readonly SupportObserverDataProviderConfiguration Configuration;
         protected readonly DataProviderContext DataProviderContext;
         protected readonly string RequestId;
+        protected readonly DiagnosticsETWProvider Logger;
         private readonly HttpClient _httpClient;
 
         public SupportObserverDataProviderBase(OperationDataCache cache, SupportObserverDataProviderConfiguration configuration, DataProviderContext dataProviderContext) : base(cache)
@@ -23,6 +26,7 @@ namespace Diagnostics.DataProviders
             Configuration = configuration;
             RequestId = dataProviderContext.RequestId;
             DataProviderContext = dataProviderContext;
+            Logger = DiagnosticsETWProvider.Instance;
             _httpClient = GetObserverClient();
             _httpClient.BaseAddress = new Uri($"{configuration.Endpoint}/api/");
             _httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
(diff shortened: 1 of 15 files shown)


src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -101,8 +105,14 @@ protected async Task<string> GetObserverResource(string url, string resourceId =
...
             }
 
-            var cancelToken = new CancellationToken();
+            var cancelTokenSource = new CancellationTokenSource();
+            var cancelToken = cancelTokenSource.Token;
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
             var response = await _httpClient.SendAsync(request, cancelToken);
+            stopwatch.Stop();
+            Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider", $"url:{url};resourceId:{resourceId};latency:{stopwatch.Elapsed};statusCode:{response.StatusCode}");
```
hforeste: - Remove the resourceId from the logging as that is only AAD related.
- The latency is tracked by `DataProviderLogDecorator`

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
```diff
@@ -67,11 +67,7 @@ private void BeginRequest_Handle(HttpContext httpContext)
 
             httpContext.RequestAborted = cTokenSource.Token;
             httpContext.Request.Headers.TryGetValue(HeaderConstants.RequestIdHeaderName, out StringValues values);
-            Dictionary<string, string> queryStringValues = new Dictionary<string, string>();
-            foreach (var query in httpContext.Request.Query)
-            {
-                queryStringValues.Add(query.Key.ToLower(), query.Value.FirstOrDefault());
-            }
+            var queryStringValues = httpContext.Request.Query.ToDictionary(query => query.Key.ToLower(), query => query.Value.FirstOrDefault());
```
hforeste: Can the `Query` part of Request be `null`? In that case `ToDictionary` will give `NullReferenceException`.

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -101,8 +105,14 @@ protected async Task<string> GetObserverResource(string url, string resourceId =
                 request.Headers.TryAddWithoutValidation("Authorization", await GetToken(resourceId));
             }
 
-            var cancelToken = new CancellationToken();
+            var cancelTokenSource = new CancellationTokenSource();
+            var cancelToken = cancelTokenSource.Token;
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
             var response = await _httpClient.SendAsync(request, cancelToken);
```
hforeste: I think I forgot to remove this but can you remove the `cancelToken` as the `DataProviderLogDecorator` class already handles cancellationToken. You can think of the `DataProviderLogDecorator` as the frontend to all data provider calls. It executes the data provider tasks and logs exceptions and helpful summary information such as latency

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
```diff
@@ -87,12 +83,11 @@ private void BeginRequest_Handle(HttpContext httpContext)
...
-            string clientObjId = string.Empty;
-            StringValues internalClientHeader;
-            bool isInternalClient = false;
+            var clientObjId = string.Empty;
+            var isInternalClient = false;
 
             // For requests coming Applens, populate client object id with Applens App Id.
-            httpContext.Request.Headers.TryGetValue("x-ms-internal-client", out internalClientHeader);
+            httpContext.Request.Headers.TryGetValue("x-ms-internal-client", out var internalClientHeader);
             if (internalClientHeader.Any())
```
rekhaswaminathan: How would this work if x-ms-internal-client header is not found?

dagrooms52: This wouldn't change the functionality at all because the variable is still declared, just in a more compact manner; that said, it could be an issue if this out var came out null, so I'm updating the if statement to take the boolean result of TryGetValue into account as well.

rekhaswaminathan: Ok. I will wait for the change.

PR #157, opened 2019-05-08
Create API to get site and hosting environment details from observer …

hforeste:


src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -101,9 +101,18 @@ protected async Task<string> GetObserverResource(string url, string resourceId =
...
-            response.EnsureSuccessStatusCode();
+            var response = await _httpClient.SendAsync(request);
+
+            try
+            {
+                response.EnsureSuccessStatusCode();
+            }
+            catch (HttpRequestException ex)
+            {
+                ex.Data.Add("StatusCode", response.StatusCode);
```
ShekharGupta1988: you should probably add response content string also if not already present in the exception.

ShekharGupta1988: I am not sure whether is for logging purposes also. I see below you use this status code.

hforeste: I'm just using this statuscode for the api response in the controller. Response content sounds like a good idea.

PR #158, opened 2019-05-13
Rekha/changeanalysis fixes

rekhaswaminathan:
Fixed following issues:

- User does not have to wait for 4 hours for AST changes to show up, Once subscription is onboarded to the service, user will be able to see the UI and scan for changes
- Without AST enabled, users can still see ARG data 
- User needs to be RBAC owner to see change details information. 
- Added logging 
- If change set id has characters such as +, it gets converted to space in backend. Added url decoding to handle this (Will add url encoding in UI once this is deployed so we dont break current flow)

### DIFF
--- a/src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
+++ b/src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
@@ -68,15 +68,27 @@ namespace Diagnostics.DataProviders
         /// <inheritdoc/>
         public async Task<List<ResourceChangesResponseModel>> GetChangesAsync(ChangeRequest changeRequest)
         {
-            string requestUri = changeAnalysisEndPoint + $"changes?api-version={apiVersion}";
-            object postBody = new
+            try
             {
-                changeRequest.ResourceId,
-                changeRequest.ChangeSetId
-            };
-            string jsonString = await PrepareAndSendRequest(requestUri, postBody, HttpMethod.Post);
-            List<ResourceChangesResponseModel> resourceChangesResponse = JsonConvert.DeserializeObject<List<ResourceChangesResponseModel>>(jsonString);
-            return resourceChangesResponse;
+                string requestUri = changeAnalysisEndPoint + $"changes?api-version={apiVersion}";
+                object postBody = new
+                {
+                    changeRequest.ResourceId,
+                    changeRequest.ChangeSetId
+                };
+                string jsonString = await PrepareAndSendRequest(requestUri, postBody, HttpMethod.Post);
+                List<ResourceChangesResponseModel> resourceChangesResponse = JsonConvert.DeserializeObject<List<ResourceChangesResponseModel>>(jsonString);
+                return resourceChangesResponse;
+            } catch (HttpRequestException httpException)
+            {
+                // Its possible that users dont have access to view the change details
+                if ((HttpStatusCode)httpException.Data["Status Code"] == HttpStatusCode.Forbidden)
+                {
+                    return new List<ResourceChangesResponseModel>();
+                }
+
+                throw httpException;
+            }
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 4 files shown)


src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
```diff
@@ -57,62 +61,106 @@ private HttpClient httpClient
...
+                    changeRequest.ChangeSetId
+                };
+                string jsonString = await PrepareAndSendRequest(requestUri, postBody, HttpMethod.Post);
+                List<ResourceChangesResponseModel> resourceChangesResponse = JsonConvert.DeserializeObject<List<ResourceChangesResponseModel>>(jsonString);
+                return resourceChangesResponse;
+            }
+            catch (HttpRequestException httpException)
+            {
+                // Its possible that users dont have access to view the change details, log the exception and send empty list of change details.
+                string message = $"HttpRequestException in GetChangesAsync, Message: {httpException.Message} ";
```
ShekharGupta1988: will a 500 also result in HttpRequestException?
Maybe we check for specific status code for "users not having access" before consuming the whole exception

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
```diff
@@ -57,62 +61,106 @@ private HttpClient httpClient
...
+                if (httpException.Data.Contains(ExceptionStatusCode))
+                {
+                    message += $"Status Code: {httpException.Data[ExceptionStatusCode]}";
+                }
+
+                DiagnosticsETWProvider.Instance.LogDataProviderMessage(requestId, "ChangeAnalysisClient", message);
```
ShekharGupta1988: I believe this should be warning instead of message?

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
rekhaswaminathan: That's what I attempted to do first. I initially handled it with 403 Forbidden Status code, but I wasn't sure if its status code will remain same or not. I can change it to check specifically for 403 - But if there is some other failure such as 500 on this endpoint, do we still want the users to see the initial view?

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
rekhaswaminathan: You mean `DataProviderException` instead of `DataProviderMessage`?

hforeste: We don't have an event with warning event level, perhaps Rekha you can create one.

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
hforeste: I haven't looked at other PRs you sent Rekha so you may have already answered this in a different PR but quick question, if you are enforcing RBAC, how are you getting the customers auth token and sending it via this HTTP call to possibly get back a 403?

rekhaswaminathan: @hforeste 403 is received when Applens makes a call to view change details of code scan changes. Since its same detector code running on Portal vs Backend, its possible that we get 403 when viewing change details

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
rekhaswaminathan: I can change it to log as exception for now. I don't think I will have enough time to make it warning level. I can make that change in a separate PR

ShekharGupta1988: Ok. File a backlog item to create a warning data provider event and for the time, use info event only. Using exception will not be the right thing there.

rekhaswaminathan: Ok sure

PR #159, opened 2019-05-14
Ajsharm/search code

SharmaAjay19:
1. Search Service to interact with python search and training api.
2. API code for Search API and Training API.
3. Modification in ListDetectors method to take a text argument and return detector results scored as similar to text.
4. Modification in ExecuteQuery (detectors) method to recommend customer problem search terms based on detector description - "SuggestedUtterances" attribute.
5. Modification in Publish method to accept a metadata attribute for detector to be saved in file metadata.json.
6. InternalAPIController in Runtime host for training needs, and enable python apis to log using runtime host event logger.


src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
```diff
@@ -68,16 +68,20 @@ public async Task<IActionResult> Post(string subscriptionId, string resourceGrou
         /// <param name="postBody">Request json body.</param>
         /// <returns>Task for listing detectors.</returns>
         [HttpPost(UriElements.Detectors)]
-        public async Task<IActionResult> ListDetectors(string subscriptionId, string resourceGroupName, string siteName, [FromBody] DiagnosticSiteData postBody)
+        public async Task<IActionResult> ListDetectors(string subscriptionId, string resourceGroupName, string siteName, [FromBody] DiagnosticSiteData postBody, [FromQuery(Name = "text")] string text = null)
         {
+            if (text == null)
+            {
+                text = Request.Query["text"].ToString();
```
rekhaswaminathan: I didnt understand why we are checking for null and getting the value from `Request.Query["text"]`.  Aren't `FromQuery` and `Request.Query` the same?

src/Diagnostics.RuntimeHost/Controllers/InternalAPIController.cs
```diff
@@ -0,0 +1,163 @@
...
+            return Ok("Model published Successfully");
+        }
+
+        [HttpPost(UriElements.RefreshModel)]
+        public async Task<IActionResult> RefreshModel(string productId)
+        {
+            Request.Headers.TryGetValue(HeaderConstants.RequestIdHeaderName, out StringValues requestIds);
+            var requestId = requestIds.FirstOrDefault();
+            var parameters = new Dictionary<string, string>();
+            parameters.Add("productId", productId);
```
rekhaswaminathan: What does `productId` represent?

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -536,12 +557,34 @@ private async Task<string> GetGistInternal(string gistId, RuntimeContext<TResour
...
         }
 
-        private async Task<IEnumerable<DiagnosticApiResponse>> ListDetectorsInternal(RuntimeContext<TResource> context)
+        private async Task<IEnumerable<DiagnosticApiResponse>> ListDetectorsInternal(RuntimeContext<TResource> context, string queryText=null)
         {
             await this._sourceWatcherService.Watcher.WaitForFirstCompletion();
+            var allDetectors = _invokerCache.GetEntityInvokerList<TResource>(context).ToList();
+
+            SearchResults searchResults = null;
+            if (queryText != null && queryText.Length>3)
```
rekhaswaminathan: `queryText.Length > 3` - Can I know why there is minimum length restriction of 3 above? What would happen if its less than 3?

src/Diagnostics.ModelsAndUtils/Models/Search/SearchModels.cs
```diff
@@ -0,0 +1,40 @@
+﻿using Diagnostics.ModelsAndUtils.Attributes;
+using Diagnostics.ModelsAndUtils.ScriptUtilities;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.Models
```
rekhaswaminathan: Add documentation for properties if possible.

src/Diagnostics.AIProjects/SearchAPI/RegistryReader.py
```diff
@@ -0,0 +1,16 @@
+from winreg import *
+githubFolderPath = ""
+kustoClientId = ""
+kustoClientSecret = ""
+try:
+	aReg = ConnectRegistry(None,HKEY_LOCAL_MACHINE)
+	aKey = OpenKey(aReg, r"SOFTWARE\Microsoft\IIS Extensions\Web Hosting Framework")
+	sourceWatcherKey = OpenKey(aKey, "SourceWatcher")
+	dataProvidersKey = OpenKey(aKey, "DiagnosticDataProviders")
+	githubKey = OpenKey(sourceWatcherKey,"Github")
```
ShekharGupta1988: Will this fail if GitHub key is not present.
We dont add Github Registry if the SourceWatcher Mode is set to Local instead of Github

src/Diagnostics.AIProjects/TrainingFactory/metadata/trainingConfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "14748": {
+    "sof-tags": [
+      "azure-web-app-service",
+      "azure-webapps",
+      "azure-app-service",
+      "azure-web-app-for-containers",
```
ShekharGupta1988: web app for containers in azure indicates linux containers. Will this create a problem?

src/Diagnostics.ModelsAndUtils/Models/Search/SearchModels.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    public class SearchResult
+    {
+        public string Detector { get; set; }
+
+        public float Score { get; set; }
+    }
+
+    public class SearchResults
```
ShekharGupta1988: All these model classes will be used by Detector Authors or just our Runtime Host API.
If there are not getting used by developer authors, then move these models into Runtime Host Project.

This project only contains models and utils that are exposed to Detector Authors

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -231,6 +238,20 @@ protected async Task<IActionResult> ExecuteQuery<TPostBodyResource>(TResource re
...
+                    // Get suggested utterances for the detector
+                    string[] utterances = null;
+                    if (jsonBody.DetectorUtterances != null && invoker.EntryPointDefinitionAttribute.Description.ToString().Length > 3)
+                    {
+                        utterances = JsonConvert.DeserializeObject<string[]>(jsonBody.DetectorUtterances);
+                        string description = invoker.EntryPointDefinitionAttribute.Description.ToString();
+                        var resourceParams = _internalApiHelper.GetResourceParams(invoker.ResourceFilter);
+                        var searchUtterances = await _searchService.SearchUtterances(runtimeContext.OperationContext.RequestId, description, utterances, resourceParams);
+                        string resultContent = await searchUtterances.Content.ReadAsStringAsync();
+                        utterancesResults = JsonConvert.DeserializeObject<QueryUtterancesResults>(resultContent);
```
ShekharGupta1988: This  casting will fail in case of response other than success  from Search API with JSON Deserialization error.

Should you include Success check here?

src/Diagnostics.RuntimeHost/Controllers/InternalAPIController.cs
ShekharGupta1988: its the PesIds like windows is 14748 etc..

src/Diagnostics.RuntimeHost/Models/InternalAPIModels.cs
```diff
@@ -0,0 +1,197 @@
+﻿using System.Collections.Generic;
+
+namespace Diagnostics.RuntimeHost.Models
+{
+    /// <summary>
+    /// Detector Metadata.
+    /// </summary>
+    public class DetectorMetadata
```
ShekharGupta1988: We already have these fields right in an existing model? Can we not reuse that or extend that

src/Diagnostics.AIProjects/SearchAPI/RegistryReader.py
ShekharGupta1988: its fine for now to not support local mode.. but add a backlog item for that for tracking

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -536,12 +557,34 @@ private async Task<string> GetGistInternal(string gistId, RuntimeContext<TResour
...
             await this._sourceWatcherService.Watcher.WaitForFirstCompletion();
+            var allDetectors = _invokerCache.GetEntityInvokerList<TResource>(context).ToList();
+
+            SearchResults searchResults = null;
+            if (queryText != null && queryText.Length>3)
+            {
+                var resourceParams = _internalApiHelper.GetResourceParams(context.OperationContext.Resource as IResourceFilter);
+                var res = await _searchService.SearchDetectors(context.OperationContext.RequestId, queryText, resourceParams);
+                string resultContent = await res.Content.ReadAsStringAsync();
+                searchResults = JsonConvert.DeserializeObject<SearchResults>(resultContent);
```
cindylovescoding: Same here maybe add successful check as Shekhar suggested

src/Diagnostics.AIProjects/SearchAPI/run.pyw
```diff
@@ -0,0 +1,334 @@
...
+        copyFolder(os.path.join(modelsPath, "resourceConfig"), os.path.join(os.path.dirname(os.path.abspath(__file__)), "resourceConfig"))
+    except Exception as e:
+        raise ResourceConfigDownloadFailed("Resource config can't be downloaded " + str(e))
+
+downloadResourceConfig()
+config = json.loads(open("resourceConfig/config.json", "r").read())
+resourceConfig = config["resourceConfig"]
+def getProductId(resourceObj):
+    productids = []
+    if resourceObj["ResourceType"] == "App":
```
cindylovescoding: Does this support other resource types other than App?

src/Diagnostics.ModelsAndUtils/Attributes/Definition.cs
```diff
@@ -83,6 +83,9 @@ public List<string> AnalysisTypes
         [DataMember]
         public DetectorType Type { get; set; }
 
+
+        public float Score { get; set; }
```
cindylovescoding: Comments for the property?

src/Diagnostics.AIProjects/TrainingFactory/metadata/trainingConfig.json
SharmaAjay19: This is only a placeholder training config for now. In ideal situations we will be providing trainingConfig whenever we train. So, this is just a sample file for how training config should look like.

src/Diagnostics.AIProjects/SearchAPI/run.pyw
SharmaAjay19: Yes, it will support all other resource types. This special handling has been done for "App" resource type since it has multiple variations.

src/Diagnostics.RuntimeHost/Models/InternalAPIModels.cs
SharmaAjay19: There are slight differences between the two models. We are representing ResourceFilter as a dictionary of strings here rather than Enum. Also we don't need attributes like CodeString, PdbBytes, DllBytes for this. This model is used to send request to the Internal API.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
SharmaAjay19: A query of length <= 3 does not make much sense as such words are mostly stopwords for our search. If the size is 3 or less we will treat it as no query i.e. we would return the list of all detectors for that resource as we would otherwise do.

src/Diagnostics.ModelsAndUtils/Models/Search/SearchModels.cs
SharmaAjay19: The model QueryUtterancesResults  is used by the DiagnosticApiResponse class of Diagnostics.ModelsAndUtilities as well as Diagnostics.RuntimeHost. Since we have project reference of ModelsAndUtilities in RuntimeHost (and not the other way round), we have to put this in ModelsAndUtilities.

src/Diagnostics.RuntimeHost/Controllers/Site/SitesController.cs
SharmaAjay19: Removed

src/Diagnostics.ModelsAndUtils/Models/Search/SearchModels.cs
SharmaAjay19: Done.

src/Diagnostics.AIProjects/SearchAPI/RegistryReader.py
SharmaAjay19: Done.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
SharmaAjay19: Done.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
SharmaAjay19: Done.

src/Diagnostics.ModelsAndUtils/Attributes/Definition.cs
SharmaAjay19: Done.

PR #163, opened 2019-05-16
Remove hardcoded string for kusto endpoint

hforeste:
Hardcoded string for Kusto query endpoint will fail in national clouds

### DIFF
--- a/src/Diagnostics.DataProviders/KustoClient.cs
+++ b/src/Diagnostics.DataProviders/KustoClient.cs
@@ -21,7 +21,7 @@ namespace Diagnostics.DataProviders
     public class KustoClient : IKustoClient
     {
         private string _requestId;
-        private const string KustoApiEndpoint = "https://{0}.kusto.windows.net:443/v1/rest/query";
+        private string KustoApiQueryEndpoint;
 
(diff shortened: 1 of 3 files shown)


src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -40,9 +40,10 @@ private HttpClient _httpClient
             }
         }
 
-        public KustoClient(string requestId)
+        public KustoClient(KustoDataProviderConfiguration config, string requestId)
         {
             _requestId = requestId;
+            KustoApiQueryEndpoint = config.KustoApiEndpoint + ":443/v1/rest/query";
```
rekhaswaminathan: Since config has kusto api end point now, does this require a change in appsettings.json?

hforeste: No as this is generated from the `AADKustoResource` which is already in the app.settings. See [Configuration](https://github.com/Azure/Azure-AppServices-Diagnostics/pull/163/files#diff-4bf3b7af0397fb900e551a8192d4b5ae)

rekhaswaminathan: I see. I understood now. Thanks Hawk

PR #166, opened 2019-05-22
Run SQL Queries Against Observer

dagrooms52:
- SQL Queries can be run from detectors using `dp.Observer.ExecuteSqlQueryAsync()`
- Queries return a DataTable for formatting in detectors
- Error messages and non-parsable JSON are handled very clearly
- Logged Issue: Observer SQL route expects `Content-Type="application/json"` but should expect `text/plain`, causes SQL query input to need to be wrapped in quotes to be accepted as JSON

### DIFF
--- a/src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
+++ b/src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
@@ -12,6 +12,7 @@ using Diagnostics.DataProviders.TokenService;
 using Diagnostics.Logger;
 using Diagnostics.ModelsAndUtils.Models.ChangeAnalysis;
 using Newtonsoft.Json;
+using static Diagnostics.Logger.HeaderConstants;
 
 namespace Diagnostics.DataProviders
 {
@@ -265,7 +266,7 @@ namespace Diagnostics.DataProviders
                     };
                 }
 
-                if(httpexception.Data.Contains(ExceptionStatusCode) && (HttpStatusCode)httpexception.Data[ExceptionStatusCode] == HttpStatusCode.Forbidden)
+                if (httpexception.Data.Contains(ExceptionStatusCode) && (HttpStatusCode)httpexception.Data[ExceptionStatusCode] == HttpStatusCode.Forbidden)
                 {
                     return new ChangeScanModel
                     {
@@ -279,7 +280,7 @@ namespace Diagnostics.DataProviders
                 return new ChangeScanModel
                 {
                     OperationId = string.Empty,
-                    State = "",
+                    State = string.Empty,
                     SubmissionTime = null,
                     CompletionTime = null
                 };
@@ -300,12 +301,12 @@ namespace Diagnostics.DataProviders
 
             // Add required headers.
             requestMessage.Headers.Add("Authorization", authToken);
-            requestMessage.Headers.Add("x-ms-client-object-id", clientObjectIdHeader);
+            requestMessage.Headers.Add(ClientObjectIdHeader, clientObjectIdHeader);
 
             // For requests coming from Diagnose and Solve, add x-ms-client-principal-name header.
             if (!string.IsNullOrWhiteSpace(clientPrincipalNameHeader))
             {
-               requestMessage.Headers.Add("x-ms-client-principal-name", clientPrincipalNameHeader);
+                requestMessage.Headers.Add(ClientPrincipalNameHeader, clientPrincipalNameHeader);
             }
 
             if (postBody != null)
--- a/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Data;
 using System.Diagnostics;
 using System.Linq;
 using System.Net.Http;
@@ -182,7 +183,7 @@ namespace Diagnostics.DataProviders
 
         public abstract Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string stampName, string siteName, string slotName);
 
-        public abstract Task<string> ExecuteSqlQueryAsync(string cloudServiceName, string query);
+        public abstract Task<DataTable> ExecuteSqlQueryAsync(string cloudServiceName, string query);
 
         public abstract HttpClient GetObserverClient();
 
--- a/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
@@ -93,16 +93,7 @@ namespace Diagnostics.DataProviders
         protected async Task<string> GetObserverResource(string url, string resourceId = null)
         {
             var request = new HttpRequestMessage(HttpMethod.Get, url);
-            request.Headers.TryAddWithoutValidation("x-ms-requestid", RequestId);
-            if (!Configuration.ObserverLocalHostEnabled)
-            {
-                request.Headers.TryAddWithoutValidation("Authorization", await GetToken(resourceId));
-            }
-
-            var response = await _httpClient.SendAsync(request);
-            Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider", $"url:{url};statusCode:{response.StatusCode}");
-
-            response.EnsureSuccessStatusCode();
+            var response = await SendObserverRequestAsync(request, resourceId);
             var result = await response.Content.ReadAsStringAsync();
 
             try
@@ -117,6 +108,20 @@ namespace Diagnostics.DataProviders
             }
 
             return result;
+        }
+
+        protected async Task<HttpResponseMessage> SendObserverRequestAsync(HttpRequestMessage request, string resourceId = null)
+        {
+            request.Headers.TryAddWithoutValidation("x-ms-requestid", RequestId);
+            if (!Configuration.ObserverLocalHostEnabled)
+            {
+                request.Headers.TryAddWithoutValidation("Authorization", await GetToken(resourceId));
+            }
+
+            var response = await _httpClient.SendAsync(request);
+            Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider", $"url:{new Uri(_httpClient.BaseAddress, request.RequestUri)}, statusCode:{response.StatusCode}");
+
+            return response;
         }
 
         private async Task<string> GetToken(string resourceId)
--- a/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
@@ -113,7 +113,7 @@ namespace Diagnostics.DataProviders
 
         protected async Task<HttpResponseMessage> SendObserverRequestAsync(HttpRequestMessage request, string resourceId = null)
         {
-            request.Headers.TryAddWithoutValidation("x-ms-requestid", RequestId);
+            request.Headers.TryAddWithoutValidation(HeaderConstants.RequestIdHeaderName, RequestId);
             if (!Configuration.ObserverLocalHostEnabled)
             {
                 request.Headers.TryAddWithoutValidation("Authorization", await GetToken(resourceId));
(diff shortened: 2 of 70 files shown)


src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -226,16 +228,48 @@ public override async Task<dynamic> GetHostingEnvironmentPostBody(string hosting
...
+                // BUG: Passing as JSON requires query to be wrapped in quotes, fix API expected content type
+                query = $"\"{query}\"";
+            }
+
+            var request = new HttpRequestMessage(HttpMethod.Post, $"/api/service/{cloudServiceName}/invokesql?api-version=2.0")
+            {
+                Content = new StringContent(query, Encoding.Default, "application/json")
+            };
+
+            var response = await SendObserverRequestAsync(request);
```
rekhaswaminathan: I remember sometime ago during our end of scrum demo we had discussions about parameterized queries. Will this PR take care of it or do we have any plans for it.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -226,16 +228,48 @@ public override async Task<dynamic> GetHostingEnvironmentPostBody(string hosting
...
+            response.EnsureSuccessStatusCode();
+
+            var result = await response.Content.ReadAsStringAsync();
+
+            DataTable datatable;
+            try
+            {
```
rekhaswaminathan: should we put try catch around the API call itself in case server returns errors like 5xx?

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
dagrooms52: I would like parameterized queries, but this PR doesn't address them at all. We can add parameters as needed, but for now this solves the main issue of too much data coming back from Observer.

rekhaswaminathan: Okay

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
dagrooms52: The `response.EnsureSuccessStatusCode()` call handles the API returning non-2xx responses.
This try/catch just handles unexpected data being returned even though a 2xx response was received, such as a message string.

rekhaswaminathan: I see. Ok

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -118,6 +111,20 @@ protected async Task<string> GetObserverResource(string url, string resourceId =
...
+        protected async Task<HttpResponseMessage> SendObserverRequestAsync(HttpRequestMessage request, string resourceId = null)
+        {
+            request.Headers.TryAddWithoutValidation("x-ms-requestid", RequestId);
+            if (!Configuration.ObserverLocalHostEnabled)
+            {
+                request.Headers.TryAddWithoutValidation("Authorization", await GetToken(resourceId));
+            }
+
+            var response = await _httpClient.SendAsync(request);
+            Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider", $"url:{new Uri(_httpClient.BaseAddress, request.RequestUri)}, statusCode:{response.StatusCode}");
```
hforeste: 👍

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -226,16 +228,48 @@ public override async Task<dynamic> GetHostingEnvironmentPostBody(string hosting
...
+                query = $"\"{query}\"";
+            }
+
+            var request = new HttpRequestMessage(HttpMethod.Post, $"/api/service/{cloudServiceName}/invokesql?api-version=2.0")
+            {
+                Content = new StringContent(query, Encoding.Default, "application/json")
+            };
+
+            var response = await SendObserverRequestAsync(request);
+            response.EnsureSuccessStatusCode();
```
hforeste: In the case we get anything other than 200, lets also log the query that way we can backtrack to the logs to look for invalid SQL Queries.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -242,18 +241,32 @@ public override async Task<DataTable> ExecuteSqlQueryAsync(string cloudServiceNa
...
             var response = await SendObserverRequestAsync(request);
-            response.EnsureSuccessStatusCode();
+
+            try
+            {
+                response.EnsureSuccessStatusCode();
+            }
+            catch (HttpRequestException ex)
+            {
+                Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider", $"message:Observer SQL query request failed, query:{query}, error:{ex.Message}");
```
hforeste: My only recommendation here is to log instead the response code and the response body in the case the body has content. Since your throwing the exception, the DataLogProviderDecorator will already log that exception

src/Diagnostics.ModelsAndUtils/Diagnostics.ModelsAndUtils.csproj
```diff
@@ -18,7 +18,7 @@
   
   <ItemGroup>
     <PackageReference Include="CommonMark.NET" Version="0.15.1" />
-    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
```
hforeste: @ShekharGupta1988 do you foresee a problem here?

hforeste: @dagrooms52 do we need to bump up the Newtonsoft version?

ShekharGupta1988: Try building the solution and verify that:
a) There are no warnings related to newtonsoft assembly version mismatch between our projects
b) in the runtimehost build output, newtonsoft dll is present (it should)
c) latest version is there.

@hforeste - In source depot, the problem was there were many dependent projects using different versions of newton soft which use to create problem.

In our case, if a) and b) are there, then it should be perfectly fine.

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -118,6 +111,20 @@ protected async Task<string> GetObserverResource(string url, string resourceId =
             return result;
         }
 
+        protected async Task<HttpResponseMessage> SendObserverRequestAsync(HttpRequestMessage request, string resourceId = null)
+        {
+            request.Headers.TryAddWithoutValidation("x-ms-requestid", RequestId);
```
ShekharGupta1988: nit: just from consistency point of view, the correct header name is "x-ms-request-id"

dagrooms52: Oops, typo 🙂

src/Diagnostics.ModelsAndUtils/Diagnostics.ModelsAndUtils.csproj
dagrooms52: This was just a consolidation step, so the 12.0.2 dll should be present as it was before.

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
```diff
@@ -300,12 +301,12 @@ private async Task<string> PrepareAndSendRequest(string requestUri, object postB
...
             // Add required headers.
             requestMessage.Headers.Add("Authorization", authToken);
-            requestMessage.Headers.Add("x-ms-client-object-id", clientObjectIdHeader);
+            requestMessage.Headers.Add(ClientObjectIdHeader, clientObjectIdHeader);
 
             // For requests coming from Diagnose and Solve, add x-ms-client-principal-name header.
             if (!string.IsNullOrWhiteSpace(clientPrincipalNameHeader))
             {
-               requestMessage.Headers.Add("x-ms-client-principal-name", clientPrincipalNameHeader);
+                requestMessage.Headers.Add(ClientPrincipalNameHeader, clientPrincipalNameHeader);
```
ShekharGupta1988: @rekhaswaminathan  - Just wanted to make sure that we have fixed this header name in the variable also? This header is used by Azure Tools team.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -226,16 +227,62 @@ public override async Task<dynamic> GetHostingEnvironmentPostBody(string hosting
...
+            var response = await SendObserverRequestAsync(request);
+            var result = await response.Content.ReadAsStringAsync();
+
+            try
+            {
+                response.EnsureSuccessStatusCode();
+            }
+            catch (HttpRequestException)
+            {
+                Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider",
```
ShekharGupta1988: Shouldn't this be logged as an error instead of info level?

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -226,16 +227,62 @@ public override async Task<dynamic> GetHostingEnvironmentPostBody(string hosting
...
+            {
+                throw new Exception($"SQL Query did not return parsable JSON; response: \"{json}\"", ex);
+            }
+
+            return datatable;
+        }
+
         public override HttpClient GetObserverClient()
         {
+            // Instantiating an HttpClient class for every request will exhaust the number of sockets available under heavy loads
```
ShekharGupta1988: I believe we should fix this as this will cause LSI by leaking sockets.

src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
ShekharGupta1988: nvmind. I found the class below while reviewing the PR:

/// <summary>
        /// Client Principal Name header.
        /// </summary>
        public static readonly string ClientPrincipalNameHeader = "x-ms-client-principal-name";

rekhaswaminathan: Yeah I checked it, its correct in variable.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
dagrooms52: I have a bug task for it. It will probably be a larger refactor throughout the codebase.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
dagrooms52: The exception will be logged by the dataProviderDecorator up the stack; this is just supplemental logging for reproducing the error.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
ShekharGupta1988: I think this is the only place we might be creating http client per request. Rest all has a singleton instance or static.

dagrooms52: True but it is still advised not to directly create an HttpClient. There are a few plugins with Polly that Hawk and I are looking at integrating as well that will automate the exponential backoff and circuit breaker patterns on our requests that will improve the usability of the services as well, so it looks like it will be a pretty involved refactor.

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
hforeste: If you change this here then also make the change in the LocalHostObserver project so we don't miss out on the logging. Right now LocalHostObserver is expecting a header with name `x-ms-requestid`

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
hforeste: Singletons are fine as far as creating an HttpClient, the documentation was saying that in the case of creating our own HttpClient even if it is singleton we need to be weary of DNS changes which honestly we're not to worried about. But I would love to utilize the Polly library for resiliency. It sounds really useful. Good find @dagrooms52

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
dagrooms52: Ah, that's where I must have copied from

PR #173, opened 2019-05-28
Add Logging for Slow SQL Queries

dagrooms52:
- SQL queries taking over 1 second will be logged for investigation purposes

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Diagnostics;
 using System.Linq;
 using System.Net.Http;
 using System.Text;
@@ -235,12 +236,17 @@ namespace Diagnostics.DataProviders
                 query = $"\"{query}\"";
             }
 
-            var request = new HttpRequestMessage(HttpMethod.Post, $"/api/service/{cloudServiceName}/invokesql?api-version=2.0")
+            var route = $"/api/service/{cloudServiceName}/invokesql?api-version=2.0";
+
+            var request = new HttpRequestMessage(HttpMethod.Post, route)
             {
                 Content = new StringContent(query, Encoding.Default, "application/json")
             };
 
+            var timer = new Stopwatch();
+            timer.Start();
             var response = await SendObserverRequestAsync(request);
+            timer.Stop();
             var result = await response.Content.ReadAsStringAsync();
 
             try
@@ -249,9 +255,18 @@ namespace Diagnostics.DataProviders
             }
             catch (HttpRequestException)
             {
-                Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider",
-                    $"message:Observer SQL query request failed, query:{query}, statusCode:{response.StatusCode}, response:{result}");
+                var logMessage = $"message:Observer SQL query request failed, route:{route}, query:{query}, " +
+                    $"statusCode:{response.StatusCode}, response:{result}, latencyMs:{timer.ElapsedMilliseconds}";
+                Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider", logMessage);
                 throw;
+            }
+
+            if (timer.ElapsedMilliseconds > 1000)
+            {
+                var logMessage = $"message:Observer SQL query ran for longer than one second, route:{route}, " +
+                    $"query:{query}, statusCode:{response.StatusCode}, response:{result}, " +
+                    $"latencyMs:{timer.ElapsedMilliseconds}";
+                Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider", logMessage);
             }
 
             return TryDeserializeDataTable(result);
--- a/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
@@ -243,10 +243,7 @@ namespace Diagnostics.DataProviders
                 Content = new StringContent(query, Encoding.Default, "application/json")
             };
 
-            var timer = new Stopwatch();
-            timer.Start();
             var response = await SendObserverRequestAsync(request);
-            timer.Stop();
             var result = await response.Content.ReadAsStringAsync();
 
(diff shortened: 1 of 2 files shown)


src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -249,11 +255,20 @@ public override async Task<DataTable> ExecuteSqlQueryAsync(string cloudServiceNa
...
             {
-                Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider",
-                    $"message:Observer SQL query request failed, query:{query}, statusCode:{response.StatusCode}, response:{result}");
+                var logMessage = $"message:Observer SQL query request failed, route:{route}, query:{query}, " +
+                    $"statusCode:{response.StatusCode}, response:{result}, latencyMs:{timer.ElapsedMilliseconds}";
+                Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider", logMessage);
                 throw;
             }
 
+            if (timer.ElapsedMilliseconds > 1000)
```
rekhaswaminathan: Is this logging for slow requests only when exception happened?

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -249,11 +255,20 @@ public override async Task<DataTable> ExecuteSqlQueryAsync(string cloudServiceNa
...
+            {
+                var logMessage = $"message:Observer SQL query ran for longer than one second, route:{route}, " +
+                    $"query:{query}, statusCode:{response.StatusCode}, response:{result}, " +
+                    $"latencyMs:{timer.ElapsedMilliseconds}";
+                Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider", logMessage);
+            }
+
             return TryDeserializeDataTable(result);
```
rekhaswaminathan: shouldn't this be outside of catch?

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -235,12 +236,17 @@ public override async Task<DataTable> ExecuteSqlQueryAsync(string cloudServiceNa
                 query = $"\"{query}\"";
             }
 
-            var request = new HttpRequestMessage(HttpMethod.Post, $"/api/service/{cloudServiceName}/invokesql?api-version=2.0")
+            var route = $"/api/service/{cloudServiceName}/invokesql?api-version=2.0";
```
hforeste: You can remove the `api-version` query string parameter

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
hforeste: Lets log this in a finally clause. We can log both success and failure SQL query calls. This way we can also look back to the SQL query that people are using for even successful but slow calls.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -249,11 +255,20 @@ public override async Task<DataTable> ExecuteSqlQueryAsync(string cloudServiceNa
...
-                    $"message:Observer SQL query request failed, query:{query}, statusCode:{response.StatusCode}, response:{result}");
+                var logMessage = $"message:Observer SQL query request failed, route:{route}, query:{query}, " +
+                    $"statusCode:{response.StatusCode}, response:{result}, latencyMs:{timer.ElapsedMilliseconds}";
+                Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider", logMessage);
                 throw;
             }
 
+            if (timer.ElapsedMilliseconds > 1000)
+            {
+                var logMessage = $"message:Observer SQL query ran for longer than one second, route:{route}, " +
```
hforeste: You don't need to log this. We can already track this info with existing kusto data from event id `3002`. Which is logged in the `DataProviderLogDecorator` class

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -249,11 +255,20 @@ public override async Task<DataTable> ExecuteSqlQueryAsync(string cloudServiceNa
             }
             catch (HttpRequestException)
             {
-                Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider",
-                    $"message:Observer SQL query request failed, query:{query}, statusCode:{response.StatusCode}, response:{result}");
+                var logMessage = $"message:Observer SQL query request failed, route:{route}, query:{query}, " +
+                    $"statusCode:{response.StatusCode}, response:{result}, latencyMs:{timer.ElapsedMilliseconds}";
```
hforeste: We already log latency in event id `3002` which is called in `DataLogProviderDecoreator` See `MakeDependencyCall` in `DataLogProviderDecorator` for more details.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -249,9 +253,13 @@ public override async Task<DataTable> ExecuteSqlQueryAsync(string cloudServiceNa
             }
             catch (HttpRequestException)
             {
-                Logger.LogDataProviderMessage(RequestId, "ObserverDataProvider",
-                    $"message:Observer SQL query request failed, query:{query}, statusCode:{response.StatusCode}, response:{result}");
-                throw;
+                loggingResponse = result;
```
hforeste: Throw the exception otherwise it will be looked at as a success

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Diagnostics;
```
hforeste: You can remove this since your not using the `StopWatch` class anymore

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
dagrooms52: Oops, moving too fast

PR #175, opened 2019-05-29
Moving the try catch to cover search service request, to handle devel…

SharmaAjay19:
…opment cases when search service is not available

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -244,13 +244,13 @@ namespace Diagnostics.RuntimeHost.Controllers
                     string[] utterances = null;
                     if (jsonBody.DetectorUtterances != null && invoker.EntryPointDefinitionAttribute.Description.ToString().Length > 3)
                     {
-                        utterances = JsonConvert.DeserializeObject<string[]>(jsonBody.DetectorUtterances);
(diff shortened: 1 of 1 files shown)


src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -244,13 +244,13 @@ protected async Task<IActionResult> ExecuteQuery<TPostBodyResource>(TResource re
...
-                        string description = invoker.EntryPointDefinitionAttribute.Description.ToString();
-                        var resourceParams = _internalApiHelper.GetResourceParams(invoker.ResourceFilter);
-                        var searchUtterances = await _searchService.SearchUtterances(runtimeContext.OperationContext.RequestId, description, utterances, resourceParams);
-                        string resultContent = await searchUtterances.Content.ReadAsStringAsync();
                         try
                         {
+                            utterances = JsonConvert.DeserializeObject<string[]>(jsonBody.DetectorUtterances);
+                            string description = invoker.EntryPointDefinitionAttribute.Description.ToString();
+                            var resourceParams = _internalApiHelper.GetResourceParams(invoker.ResourceFilter);
+                            var searchUtterances = await _searchService.SearchUtterances(runtimeContext.OperationContext.RequestId, description, utterances, resourceParams);
```
dagrooms52: Any true errors with the service call are going to get squashed in production, which is more important than being able to run the service locally. We may want to wire in the configuration setting for "local server" and skip the service if it's set

PR #179, opened 2019-05-31
kusto fail over

craigmaher:
This change aims to fail over to a backup kusto cluster when the primary kusto cluster is failing.

This change creates a heart beat service which monitors the primary kusto cluster by sending a simple kusto query every 30 seconds.
when 5 consecutive kusto failures happen for a give cluster it will fail over to a backup cluster and will only switch back after 5 consecutive success.


src/Diagnostics.DataProviders/DataProviders.cs
```diff
@@ -7,21 +7,21 @@ public class DataProviders
...
         private OperationDataCache _cache = new OperationDataCache();
 
-        public IKustoDataProvider Kusto;
-        public ISupportObserverDataProvider Observer;
-        public IGeoMasterDataProvider GeoMaster;
-        public IAppInsightsDataProvider AppInsights;
-        public IChangeAnalysisDataProvider ChangeAnalysis;
-        public IAscDataProvider Asc;
-        public Func<MdmDataSource, IMdmDataProvider> Mdm;
+        public IKustoDataProvider Kusto { get; }
```
hforeste: Please change these back to fields and not properties otherwise detector would fail because they would need to be recompiled. We ran into this before and it was a sad day :(

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -451,6 +451,24 @@ public void LogKustoQueryInformation(string Source, string RequestId, string Mes
...
         }
 
+        /// <summary>
+        /// Log kusto query information.
+        /// </summary>
+        /// <param name="RequestId">Request id.</param>
+        /// <param name="Message">The message.</param>
+        /// <param name="ExceptionType">Exception type.</param>
+        /// <param name="ExceptionDetails">Exception details.</param>
+        [Event(3004, Level = EventLevel.Informational, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogKustoQueryInformation)]
```
hforeste: This has to be a new eventid. Increment it and make it EventId `3005`

hforeste: Also add a new template for `ETWMessageTemplates`

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
```diff
@@ -83,6 +83,7 @@ private void BeginRequest_Handle(HttpContext httpContext)
             var dataSourcesConfigurationService = ((ServiceProvider)httpContext.RequestServices).GetService<IDataSourcesConfigurationService>();
             var wawsObserverTokenService = ((ServiceProvider)httpContext.RequestServices).GetService<IWawsObserverTokenService>();
             var supportBayApiObserverTokenService = ((ServiceProvider)httpContext.RequestServices).GetService<ISupportBayApiObserverTokenService>();
+            var kustoHeartBeatService = ((ServiceProvider)httpContext.RequestServices).GetService<IKustoHeartBeatService>();
```
hforeste: This is actually an AntiPattern called ServiceLocator that I started with the wawsObserverTokenService. Instead of using the `GetService` method to get the service we want we should take advantage of the dependency injecting capabilities of .NET Core. We can proceed with this behavior for the sake and urgency of this PR but lets correct this behavior in a separate PR. I'll create a work item for this in our backlog.

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -0,0 +1,176 @@
...
+                }
+                else
+                {
+                    _ConsecutiveFailureCount++;
+                    _ConsecutiveSuccessCount = 0;
+                }
+
+                // if not in failover state
+                //  should failover?
+                if (UsePrimaryCluster && _ConsecutiveFailureCount >= 5)
```
hforeste: Lets make the ConsecutiveFailure threshold configurable in the `KustoDataProviderConfiguration` so that we don't have to recompile the source code every time we want to change the threshold.

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -0,0 +1,176 @@
...
+        {
+            while (true) {
+                bool heartBeatSuccess = false;
+                Exception exceptionForLog = null;
+                string requestID = Guid.NewGuid().ToString();
+                try
+                {
+                    // Run kusto query with  10 second time out
+                    string query = "RoleInstanceHeartbeat | where TIMESTAMP >= ago(30m) | take 1";
+                    var result = await _kustoDataProvider.ExecuteQueryDirect(query, PrimaryCluster, 10, requestID);
```
hforeste: Let's make the timeout configurable in the `KustoDataProviderConfiguration` so that we don't have to recompile the source code when we change the timeout.

src/Diagnostics.DataProviders/DataProviders.cs
```diff
@@ -7,21 +7,21 @@ public class DataProviders
...
+            Kusto = new DataProviderLogDecorator(context, new KustoDataProvider(_cache, context.Configuration.KustoConfiguration, context.RequestId, context.KustoHeartBeatService));
             Observer = new DataProviderLogDecorator(context, SupportObserverDataProviderFactory.GetDataProvider(_cache, context.Configuration, context));
             GeoMaster = new DataProviderLogDecorator(context, new GeoMasterDataProvider(_cache, context.Configuration.GeoMasterConfiguration));
             AppInsights = new DataProviderLogDecorator(context, new AppInsightsDataProvider(_cache, context.Configuration.AppInsightsConfiguration));
-            ChangeAnalysis = new DataProviderLogDecorator(context, new ChangeAnalysisDataProvider(_cache, context.Configuration.ChangeAnalysisDataProviderConfiguration, context.Configuration.KustoConfiguration, context.RequestId, context.clientObjectId, context.clientPrincipalName));
+            ChangeAnalysis = new DataProviderLogDecorator(context, new ChangeAnalysisDataProvider(_cache, context.Configuration.ChangeAnalysisDataProviderConfiguration, context.RequestId, context.clientObjectId, context.clientPrincipalName, Kusto));
```
hforeste: 👍

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -56,14 +58,19 @@ public async Task<KustoQuery> GetKustoQuery(string query, string stampName)
             return kustoQuery;
         }
 
+        internal async Task<DataTable> ExecuteQueryDirect(string query, string cluster, int timeoutSeconds, string requestId = null, string operationName = null)
+        {
+            await AddQueryInformationToMetadata(query, cluster);
```
hforeste: I assume you wanted to remove this and that's why you have it added to line 45

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -56,14 +58,19 @@ public async Task<KustoQuery> GetKustoQuery(string query, string stampName)
             return kustoQuery;
         }
 
+        internal async Task<DataTable> ExecuteQueryDirect(string query, string cluster, int timeoutSeconds, string requestId = null, string operationName = null)
```
ShekharGupta1988: What is this method intended for? The name is not very clear

hforeste:
Adding @cindylovescoding  as she was curious on how this will be implemented and adding @rekhaswaminathan since we slightly changed something in her code path.

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -0,0 +1,176 @@
...
+        {
+            _heartbeats = new Dictionary<string, KustoHeartBeat>();
+            _kustoDataProvider = new KustoDataProvider(new OperationDataCache(), _configuration, Guid.NewGuid().ToString(), this);
+            _threads = new List<Thread>();
+            foreach (var value in _configuration.FailoverClusterNameCollection)
+            {
+                _heartbeats[value.Key] = new KustoHeartBeat(value.Key, value.Value, _kustoDataProvider);
+                // Start threads for each heartbeat
+                ThreadStart threadStart = new ThreadStart(_heartbeats[value.Key].RunHeartBeatThread);
+                Thread thread = new Thread(threadStart);
```
ShekharGupta1988: To acheive the same, Should we use Tasks instead of creating new Threads.  That will use TaskScheduler (running on thread using ThreadPool) and every task will not create its own thread

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -0,0 +1,176 @@
...
+            UsePrimaryCluster = true;
+            _kustoDataProvider = kustoDataProvider;
+        }
+
+        public async void RunHeartBeatThread()
+        {
+            while (true) {
+                bool heartBeatSuccess = false;
+                Exception exceptionForLog = null;
+                string requestID = Guid.NewGuid().ToString();
```
ShekharGupta1988: nit: just to remove confusion here, should this be renamed to be invocationId instead of requestId (as there is no API call here)

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -0,0 +1,176 @@
...
+                    UsePrimaryCluster = false;
+                } // else should stop failover
+                else if (!UsePrimaryCluster && _ConsecutiveSuccessCount >= 5)
+                {
+                    UsePrimaryCluster = true;
+                }
+
+                LogHeartBeatInformation(heartBeatSuccess, PrimaryCluster, FailoverCluster, requestID, _ConsecutiveSuccessCount, _ConsecutiveFailureCount, exceptionForLog);
+
+                Thread.Sleep(TimeSpan.FromSeconds(20));
```
ShekharGupta1988: this will suspend the whole thread. 
Instead, you should use `await Task.Delay(..)`

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -451,6 +451,24 @@ public void LogKustoQueryInformation(string Source, string RequestId, string Mes
...
+        public void LogKustoHeartbeatInformation(string RequestId, string Message, string ExceptionType, string ExceptionDetails)
+        {
+            WriteDiagnosticsEvent(
+                3005,
+                RequestId,
```
hforeste: Remove requestid since there is no request coming in from client. This will make it confusing for our other requestid

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        private void LogHeartBeatInformation(bool success, string cluster, string failoverCluster, string requestID, int successCount, int failureCount, Exception exception)
+        {
+            var status = success ? "Success" : "Failed";
+
+
+            DiagnosticsETWProvider.Instance.LogKustoHeartbeatInformation(
+               requestID,
+               $"Status:{status},PriamryCluster:{cluster},FasiloverCluster:{failoverCluster},ConsecutiveSuccessCount:{successCount},ConsecutiveFailureCount:{failureCount}",
```
ShekharGupta1988: spell mistake : FailoverCluster

ShekharGupta1988: also PrimaryCluster :)

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -39,8 +41,8 @@ public async Task<DataTable> ExecuteClusterQuery(string query, string requestId
 
         public async Task<DataTable> ExecuteQuery(string query, string stampName, string requestId = null, string operationName = null)
         {
-            await AddQueryInformationToMetadata(query, stampName);
             var cluster = GetClusterNameFromStamp(stampName);
+            await AddQueryInformationToMetadata(query, cluster);
             return await _kustoClient.ExecuteQueryAsync(query, cluster, _configuration.DBName, requestId, operationName);
```
hforeste: Can you log additional message that states the cluster that were about to send a request to?

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
dagrooms52: I have a [User Story](https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/341/) but haven't added this file to it yet, I'll update it

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -0,0 +1,183 @@
...
+                if (UsePrimaryCluster && _ConsecutiveFailureCount >= _configuration.HeartBeatConsecutiveFailureLimit)
+                {
+                    UsePrimaryCluster = false;
+                } // else should stop failover
+                else if (!UsePrimaryCluster && _ConsecutiveSuccessCount >= _configuration.HeartBeatConsecutiveFailureLimit)
+                {
+                    UsePrimaryCluster = true;
+                }
+
+                LogHeartBeatInformation(primaryHeartBeatSuccess, failoverHeartBeatSuccess, PrimaryCluster, FailoverCluster, activityId, _ConsecutiveSuccessCount, _ConsecutiveFailureCount, exceptionForLog);
```
ShekharGupta1988: How about also logging the startTime and endTime for the activity? 
It will give you insights on the performance of this task in production.

craigmaher: Since KustoClient.cs already logs time taken we could just cross reference the Activation Id with the Request Id. I can add redundant logging if cross referencing the log information is too difficult?

ShekharGupta1988: I see.. good point.

ShekharGupta1988: its just a nice to have logging. But you are right. we can figure that out using Kusto logging also.

src/Diagnostics.DataProviders/Interfaces/IKustoClient.cs
```diff
@@ -5,6 +5,8 @@ namespace Diagnostics.DataProviders
 {
     public interface IKustoClient
     {
+        Task<DataTable> ExecuteQueryAsync(string query, string cluster, string database, int timeoutSeconds, string requestId = null, string operationName = null);
```
hforeste: Did you decide to stick with the overloaded method name or instead use `ExecuteQueryAsyncForHeartbeat`?

hforeste:
@craigmaher the PR is looking real good. I am thinking that it may be a good idea to get a unit test or two for the failover logic that you have. I just want to make sure that in the future when we touch this code that we don't accidentally regress it. If you want to sync in person to discuss implementation then feel free to ping me.

src/Diagnostics.DataProviders/Interfaces/IKustoClient.cs
craigmaher: I think for this an overload is a better method name. My change in KustoClient.cs one method calls the other just with a default parameter.

hforeste:
So the reason for Travis CI failing was the delay time?

craigmaher:
Yes, Travis CI does not have enough time for the heartbeat task to run in a small delay, too ensure the test is not flakey I’m checking that a heartbeat has been run after a change is made to weather a heartbeat should succeed or fail.

PR #181, opened 2019-05-31
Fix author info missing in applens

cindylovescoding:
Seems like we are modifying existing invoker cache instance while doing remove PII.
Created a copied definition object as response.


src/Diagnostics.ModelsAndUtils/Attributes/Definition.cs
```diff
@@ -92,5 +92,22 @@ public bool Equals(Definition other)
...
+
+        public Definition()
+        {
+        }
+
+        public Definition(Definition definition)
+        {
+            Id = definition.Id;
+            Name = definition.Name;
+            Description = definition.Description;
```
ShekharGupta1988: Random thought : instead of copying each element, maybe we serialize the object to a string or memory stream and then create a new object by deserializing that.

I am just wondering that if we add a new member to this class, we might forget to update here

cindylovescoding: I agree Shekhar, good suggestion!

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -808,9 +808,14 @@ private bool VerifyEntity(EntityInvoker invoker, ref QueryResponse<DiagnosticApi
...
         private Definition RemovePIIFromDefinition(Definition definition, bool isInternal)
-        {
-            if (!isInternal) definition.Author = string.Empty;
-            return definition;
+        { 
+            string definitionString = JsonConvert.SerializeObject(definition);
+            Definition definitionCopy = JsonConvert.DeserializeObject<Definition>(definitionString);
+            if (!isInternal)
+            {
+                definitionCopy.Author = string.Empty;
```
dagrooms52: May need to null check `definition`.

Also, this raises the opportunity for a `CloneJson` extension method like the one described here: https://stackoverflow.com/questions/78536/deep-cloning-objects

cindylovescoding: We will want to revisit the clone logic for the diagrole for the root fix. Thanks Daniel!

PR #184, opened 2019-06-05
ChangeAnalysis gist migration

rekhaswaminathan:
These changes will make Change Analysis Data Provider easily consumed by Gist. 

- We can maintain models and logic in the gist and call the Data Provider to forward the request. (Like what ASC Data Provider gist does)
- Logs the exception message if status code is not successful. 
- Increased the form input value to 150 characters so that we can pass detector query params from the UI and it can be consumed in the gist.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/ChangeAnalysisDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/ChangeAnalysisDataProvider.cs
@@ -7,6 +7,7 @@ using Diagnostics.DataProviders.DataProviderConfigurations;
 using Diagnostics.DataProviders.Interfaces;
 using Diagnostics.ModelsAndUtils.Models;
 using Diagnostics.ModelsAndUtils.Models.ChangeAnalysis;
+using System.Net.Http;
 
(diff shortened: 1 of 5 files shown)


src/Diagnostics.DataProviders/DataProviders/ChangeAnalysisDataProvider.cs
```diff
@@ -199,5 +200,17 @@ private List<string> GetHostNamesFromTable(DataTable kustoTable)
...
         }
+
+        /// <summary>
+        /// Forwards the request to Change Analysis Client
+        /// </summary>
+        /// <param name="requestUri">Request URI</param>
+        /// <param name="postBody">Post body</param>
+        /// <param name="method">HTTP Method.</param>
+        /// <returns>JSON string</returns>
+        public async Task<string> ForwardRequestToClient(string requestUri, object postBody = null, HttpMethod method = null)
```
ShekharGupta1988: nitpick : Can the name be little bit more obvious to discover? like InvokeChangeAnalysisRequest or something like that?

PR #185, opened 2019-06-05
Configuration switch for search service integration

SharmaAjay19:


src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -571,24 +574,38 @@ private async Task<IEnumerable<DiagnosticApiResponse>> ListDetectorsInternal(Run
...
+                            {
+                                p.EntryPointDefinitionAttribute.Score = 0;
+                            }
+                        });
+                        allDetectors = allDetectors.Where(x => x.EntryPointDefinitionAttribute.Score > 0).ToList();
                     }
                     else
                     {
-                        p.EntryPointDefinitionAttribute.Score = 0;
+                        throw new Exception("Search API returned null");
```
ShekharGupta1988: are we throwing a 500 if search service returned no detectors?

ShekharGupta1988: if yes, then we should return [] with a 200 Response. 
If this is some error condition, a more verbose error message will help in production.

src/Diagnostics.RuntimeHost/Services/SearchService/SearchService.cs
```diff
@@ -30,54 +34,97 @@ public class SearchService : ISearchService
...
+                isEnabled = Convert.ToBoolean(Registry.GetValue(RegistryConstants.SearchAPIRegistryPath, RegistryConstants.SearchAPIEnabledKey, string.Empty));
+            }
+            else
+            {
+                isEnabled = Convert.ToBoolean(_config[$"SearchAPI:{RegistryConstants.SearchAPIEnabledKey}"]);
+            }
         }
 
-        public Task<HttpResponseMessage> Get(HttpRequestMessage request)
+        public async Task<HttpResponseMessage> Get(HttpRequestMessage request)
```
ShekharGupta1988: out of curiosity, why this change?  looks like you dont need to await on any statement in this method

src/Diagnostics.RuntimeHost/Services/SearchService/SearchService.cs
```diff
@@ -30,54 +34,97 @@ public class SearchService : ISearchService
...
+                parameters.Add("requestId", requestId ?? string.Empty);
+                HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Post, QueryDetectorsUrl);
+                request.Content = new StringContent(JsonConvert.SerializeObject(parameters), Encoding.UTF8, "application/json");
+                return await Get(request);
+            }
+            return null;
         }
 
-        public Task<HttpResponseMessage> SearchUtterances(string requestId, string query, string[] detectorUtterances, Dictionary<string, string> parameters)
+        public async Task<HttpResponseMessage> SearchUtterances(string requestId, string query, string[] detectorUtterances, Dictionary<string, string> parameters)
```
ShekharGupta1988: same here

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -571,24 +574,38 @@ private async Task<IEnumerable<DiagnosticApiResponse>> ListDetectorsInternal(Run
...
                     }
-                });
-                allDetectors = allDetectors.Where(x => x.EntryPointDefinitionAttribute.Score > 0).ToList();
+                }
+                else
+                {
+                    throw new Exception("Search service is disabled.");
```
dagrooms52: This would still break the local environment if we disable search service.

I have a branch with changes to disable search service - https://github.com/Azure/Azure-AppServices-Diagnostics/blob/c91f0cc20707bfa82d6d53d1d9ef8aae8899f634/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs

PR #186, opened 2019-06-06
make heartbeat better able to handle cases with no failover

craigmaher:
this change makes heartbeat still log information where there is no fail over.


craigmaher:
Thought of a bug I would like to get fixed with this change

PR #187, opened 2019-06-07
Cmaher/kusto failover improovments

craigmaher:
this change makes heartbeat still log information where there is no fail over.

this change also aims to fix a bug where an exception of the secondary heart beat would cause the primary heart beat to fail. now both heart beats exceptions are caught and logged separately.

### DIFF
--- a/src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
+++ b/src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
@@ -113,86 +113,91 @@ namespace Diagnostics.DataProviders
             _configuration = configuration;
         }
 
+        private async Task RunHeartBeatPrimary(string activityId)
+        {
+            bool primaryHeartBeatSuccess = false;
+            Exception exceptionForLog = null;
+            try
+            {
+                var primaryHeartBeat = await _kustoDataProvider.ExecuteQueryForHeartbeat(_configuration.HeartBeatQuery, PrimaryCluster, _configuration.HeartBeatTimeOut, activityId);
+
+                if (primaryHeartBeat.Rows.Count >= 1)
+                {
+                    primaryHeartBeatSuccess = true;
+                }
+            }
+            catch (Exception ex)
+            {
+                exceptionForLog = ex;
+            }
+
+            if (primaryHeartBeatSuccess)
+            {
+                _ConsecutiveFailureCount = 0;
+                _ConsecutiveSuccessCount++;
+            }
+            else
+            {
+                _ConsecutiveFailureCount++;
+                _ConsecutiveSuccessCount = 0;
+            }
+
+            // if not in failover state
+            //  should failover?
+            if (UsePrimaryCluster && _ConsecutiveFailureCount >= _configuration.HeartBeatConsecutiveFailureLimit)
+            {
+                UsePrimaryCluster = false;
+            } // else should stop failover
+            else if (!UsePrimaryCluster && _ConsecutiveSuccessCount >= _configuration.HeartBeatConsecutiveFailureLimit)
+            {
+                UsePrimaryCluster = true;
+            }
+
+            LogHeartBeatInformation("Primary", primaryHeartBeatSuccess, PrimaryCluster, activityId, UsePrimaryCluster, exceptionForLog);
+        }
+
+        private async Task RunHeartBeatFailover(string activityId)
+        {
+            bool failoverHeartBeatSuccess = false;
+            Exception exceptionForLog = null;
+            try
+            {
+                var failoverHeartBeat = await _kustoDataProvider.ExecuteQueryForHeartbeat(_configuration.HeartBeatQuery, PrimaryCluster, _configuration.HeartBeatTimeOut, activityId);
+
+                if (failoverHeartBeat.Rows.Count >= 1)
+                {
+                    failoverHeartBeatSuccess = true;
+                }
+            }
+            catch (Exception ex)
+            {
+                exceptionForLog = ex;
+            }
+
+            LogHeartBeatInformation("Failover", failoverHeartBeatSuccess, FailoverCluster, activityId, UsePrimaryCluster, exceptionForLog);
+        }
+
         public async Task RunHeartBeatTask(CancellationToken cancellationToken)
         {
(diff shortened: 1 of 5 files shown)


src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -59,7 +65,7 @@ public string GetClusterNameFromStamp(string stampName)
                 }
             }
 
-            if (_heartbeats.ContainsKey(kustoClusterName))
+            if (!string.IsNullOrEmpty(_heartbeats[kustoClusterName].FailoverCluster))
```
hforeste: Use `IsNullOrWhitespace`

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -107,75 +113,97 @@ public KustoHeartBeat(string primaryCluster, string failoverCluster, KustoDataPr
...
                 }
-
-                LogHeartBeatInformation(primaryHeartBeatSuccess, failoverHeartBeatSuccess, PrimaryCluster, FailoverCluster, activityId, _ConsecutiveSuccessCount, _ConsecutiveFailureCount, exceptionForLog);
-
                 await Task.Delay(TimeSpan.FromSeconds(_configuration.HeartBeatDelay), cancellationToken);
             }
         }
 
-        private void LogHeartBeatInformation(bool primaryClustersuccess, bool failoverClusterSuccess, string cluster, string failoverCluster, string invocationId, int successCount, int failureCount, Exception exception)
+        private void LogHeartBeatInformation(string primaryOrFailover, bool clustersuccess, string cluster, string activityId, bool usingPrimaryCluster, Exception exception)
```
hforeste: Maybe rename `clustersuccess` to `clusterStatus` for better readability

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -107,75 +113,97 @@ public KustoHeartBeat(string primaryCluster, string failoverCluster, KustoDataPr
...
-            var primaryClusterStatus = primaryClustersuccess ? "Success" : "Failed";
-            var failoverClusterStatus = failoverClusterSuccess ? "Success" : "Failed";
-
+            var clusterStatus = clustersuccess ? "Success" : "Failed";
 
             DiagnosticsETWProvider.Instance.LogKustoHeartbeatInformation(
-               invocationId,
-               $"PrimaryClusterStatus:{primaryClusterStatus},FailoverClusterStatus:{failoverClusterStatus},PrimaryCluster:{cluster},FailoverCluster:{failoverCluster},PrimaryClusterConsecutiveSuccessCount:{successCount},PrimaryClusterConsecutiveFailureCount:{failureCount}",
+               activityId,
+               $"ClusterType:{primaryOrFailover},ClusterStatus:{clusterStatus},Cluster:{cluster},UsingPrimaryCluster:{usingPrimaryCluster}",
```
hforeste: Perhaps change `{clusterStatus}` to `clusterStatus ? "Success" : "Failed"`

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -107,75 +113,97 @@ public KustoHeartBeat(string primaryCluster, string failoverCluster, KustoDataPr
...
-                    {
-                        failoverHeartBeatSuccess = true;
-                    }
+            // if not in failover state
+            //  should failover?
+            if (UsePrimaryCluster && _ConsecutiveFailureCount >= _configuration.HeartBeatConsecutiveFailureLimit)
+            {
+                UsePrimaryCluster = false;
+            } // else should stop failover
+            else if (!UsePrimaryCluster && _ConsecutiveSuccessCount >= _configuration.HeartBeatConsecutiveFailureLimit)
```
hforeste: Can we create another configuration for `ConsecutiveSuccessLimit`? For me it reduces the confusion of comparing `ConsecutiveSuccessCount` to `HeartBeatConsecutiveFailureLimit`. And also it would be nice to have tighter grain of control of success and failures. For example, if the Kusto service fails 5 consecutive times then we'll failover but Kusto has to succeed 8 consecutive times before we go back to the primary cluster. Having these knobs now will allow us to find the best configuration we need in the future as we monitor our service.

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -107,75 +113,97 @@ public KustoHeartBeat(string primaryCluster, string failoverCluster, KustoDataPr
...
-                {
-                    UsePrimaryCluster = true;
+                    var primaryTask = RunHeartBeatPrimary(activityId);
+                    var failoverTask = RunHeartBeatFailover(activityId);
+                    await Task.WhenAll(new Task[] { primaryTask, failoverTask });
                 }
-
-                LogHeartBeatInformation(primaryHeartBeatSuccess, failoverHeartBeatSuccess, PrimaryCluster, FailoverCluster, activityId, _ConsecutiveSuccessCount, _ConsecutiveFailureCount, exceptionForLog);
-
                 await Task.Delay(TimeSpan.FromSeconds(_configuration.HeartBeatDelay), cancellationToken);
```
hforeste: Could we rename the configuration from `HeartBeatDelay` to `HeartBeatDelayInSeconds`. That way when we configure it in the future we don't have to look back to the code to check whether the delay is for seconds or milliseconds or minutes.

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
craigmaher: Are you saying to inline the conditional operation?
If so I think it looks cleaner outside the string as I don't have to escape the quotation marks.

hforeste: Ah sounds good okay

PR #189, opened 2019-06-07
Ignore Null Sitename When Looking for Swap History Removed Message

dagrooms52:


src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -94,7 +94,9 @@ public override async Task<dynamic> GetCertificatesInResourceGroupAsync(string s
             const string missingDataSignature = "SWAP HISTORY REMOVED";
             foreach (var slotMap in slotTimeRange)
             {
-                var missingHistoricalSwapData = slotMap.Value.Where(swapInfo => swapInfo.RuntimeSitename.Equals(missingDataSignature, StringComparison.CurrentCultureIgnoreCase));
```
hforeste: Did this cause `NullReferenceExceptions`?

dagrooms52: Yes, RuntimeSitename isn't always present in the JSON data, I pulled the raw data over the API to see what was happening

PR #192, opened 2019-06-10
Automatic refresh of search models with github change

SharmaAjay19:


src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/GitHubWatcher.cs
```diff
@@ -306,6 +316,22 @@ private async Task DownloadContentAndUpdateCache(GithubEntry parentGithubEntry,
             }
         }
 
+        private void HitModelRefresh(string productId)
+        {
+            Dictionary<string, string> parameters = new Dictionary<string, string>();
+            parameters.Add("productId", productId);
+            var requestId = Guid.NewGuid().ToString();
```
rekhaswaminathan: Should this be requestId of the Diagnostics request?

rekhaswaminathan: Never mind, I just checked that requestId gets logged for Runtime host events only

PR #193, opened 2019-06-11
Support regional geomasters

michimune:


src/Diagnostics.DataProviders/DataProviderContext.cs
```diff
@@ -13,6 +13,7 @@ public class DataProviderContext
         public IWawsObserverTokenService WawsObserverTokenService { get; private set; }
         public ISupportBayApiObserverTokenService SupportBayApiObserverTokenService { get; private set; }
         public IKustoHeartBeatService KustoHeartBeatService { get; private set; }
+        public string GeomasterName { get; private set; }
```
ShekharGupta1988: nit : is this the geomasterEndpoint or just the name of the cloud service?  If its an endpoint, lets call it out explicitly in the variable.

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
```diff
@@ -110,7 +110,13 @@ private void BeginRequest_Handle(HttpContext httpContext)
                 clientPrincipalName = httpContext.Request.Headers[ClientPrincipalNameHeader];
             }
 
-            httpContext.Items.Add(HostConstants.DataProviderContextKey, new DataProviderContext(dataSourcesConfigurationService.Config, values.FirstOrDefault() ?? string.Empty, cTokenSource.Token, startTimeUtc, endTimeUtc, wawsObserverTokenService, supportBayApiObserverTokenService, clientObjId, clientPrincipalName, kustoHeartBeatService));
+            string geomasterName = string.Empty;
+            if (httpContext.Request.Headers.ContainsKey(GeomasterNameHeader))
+            {
```
rekhaswaminathan: Will this header be sent for all requests regional and global? We can do `else geoMaster=<global-geo>` for safety

src/Diagnostics.DataProviders/DataProviderContext.cs
michimune: It's just a DNS name. I initially thought I wanted to use an endpoint name like "https://...." but found that it's a bit difficult to get the CSM port number there. So I just ended up with the DNS name.

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
michimune: That default logic is in the GeoMasterCertClient constructor:
  var geoEndpoint = new UriBuilder(geoMasterName ?? configuration.GeoEndpointAddress)
But I noticed that geoMasterName is Empty if not supplied, so I will update the logic.

michimune: Updated

PR #196, opened 2019-06-17
use azure key vault for dev secrets

craigmaher:
This change enables the use of azure key vault for dev secrets.
This change should allow developers to quickly get started developing and allow appsettings.Development.json to be checked into revision control to make it easier to make changes to the dev environment globally.


src/Diagnostics.DataProviders/ConfigurationFactory.cs
```diff
@@ -24,6 +27,17 @@ public AppSettingsDataProviderConfigurationFactory(IHostingEnvironment env)
                 .SetBasePath(Directory.GetCurrentDirectory())
                 .AddJsonFile($"appsettings.{env.EnvironmentName}.json", optional: true, reloadOnChange: true);
 
+            var tokenProvider = new AzureServiceTokenProvider();
+            var keyVaultClient = new KeyVaultClient(
+                new KeyVaultClient.AuthenticationCallback(
+                    tokenProvider.KeyVaultTokenCallback
```
hforeste: Will this open up a modal box to sign in?

src/Diagnostics.DataProviders/ConfigurationFactory.cs
```diff
@@ -24,6 +27,17 @@ public AppSettingsDataProviderConfigurationFactory(IHostingEnvironment env)
...
+                )
+            );
+
+            builder.AddAzureKeyVault("https://appservicediagnostics.vault.azure.net/",
```
hforeste: What subscription is this key vault in?

.gitignore
```diff
@@ -314,6 +314,3 @@ build/
 dist/
 aot/
 **/Diagnostics.AIProjects/run/**
-
-# Appsettings file that may contain secrets
-appsettings.*.json
```
dagrooms52: https://gph.is/18ttluO

src/Diagnostics.RuntimeHost/Program.cs
```diff
@@ -13,12 +16,26 @@ public static void Main(string[] args)
...
+
+                    if (context.HostingEnvironment.IsDevelopment())
+                    {
+                        var azureServiceTokenProvider = new AzureServiceTokenProvider();
+                        var keyVaultClient = new KeyVaultClient(
+                            new KeyVaultClient.AuthenticationCallback(
+                                azureServiceTokenProvider.KeyVaultTokenCallback));
+
+                        config.AddAzureKeyVault(
+                            $"https://appservicediagnostics.vault.azure.net/",
```
dagrooms52: Since this is copied around a few times, we should store it in a constants file in case it needs to change

src/Diagnostics.DataProviders/ConfigurationFactory.cs
craigmaher: it is currently hosted under my test subscription but as soon as I get an AME account I plan to move it to the new subscription.

src/Diagnostics.DataProviders/ConfigurationFactory.cs
craigmaher: yes

src/Diagnostics.DataProviders/ConfigurationFactory.cs
craigmaher: Talking to Shekhar the dev secrets will need to be in someones dev private subscription so I'm just going to leave the developer subscription in my subscription. everyone in the team will have access via a security group though.

src/Diagnostics.DataProviders/ConfigurationFactory.cs
ShekharGupta1988: How will this run in Production then?

src/Diagnostics.DataProviders/DataProviderConstants.cs
```diff
@@ -15,6 +15,8 @@ public class DataProviderConstants
 
         public const string AzureUSGovernment = "Government";
 
+        public const string KeyVaultURL = "https://appservicediagnostics.vault.azure.net/";
```
ShekharGupta1988: 1. Is this the right file for this Const.
2. Ideally, this should be coming from AppSettings (overridden for different environments) ?

src/Diagnostics.RuntimeHost/Program.cs
```diff
@@ -13,12 +17,26 @@ public static void Main(string[] args)
...
-
             return WebHost.CreateDefaultBuilder(args)
-                .UseConfiguration(config)
+                .ConfigureAppConfiguration((context, config) =>
+                {
+                    var builtConfig = config
+                        .AddCommandLine(args)
+                        .Build();
+
+                    if (context.HostingEnvironment.IsDevelopment())
```
ShekharGupta1988: So this code path is only for dev mode and not for Prod?

src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -3,50 +3,50 @@
...
-    "HeartBeatTimeOutInSeconds": 10,
-    "HeartBeatDelayInSeconds": 20,
-    "AADAuthority": "https://login.windows.net/microsoft.com",
-    "AADKustoResource": "https://wawskustotest.kusto.windows.net"
-  },
+"Kusto": {
+  "ClientId": "",
+  "AppKey": "",
+  "DBName": "wawsprod",
+  "KustoRegionGroupings": "BAY:MWH:MSFTBAY,BLU:BN1:YQ1:MSFTBLU,CH1:CQ1:CY4:DM1:SN1:YT1,AM2,DB3:MSFTDB3,*",
```
ShekharGupta1988: Lets keep this appsettings file clean, just schema and not values..

src/Diagnostics.DataProviders/ConfigurationFactory.cs
craigmaher: there are multiple ways to get credentials: this initial change is to allow key vault for dev access, I intend to build on top of this to allow keyvault for production

src/Diagnostics.RuntimeHost/Program.cs
craigmaher: yes, this change is for dev mode I intend to follow up with a change for prod

src/Diagnostics.RuntimeHost/appsettings.json
craigmaher: I understand but I think changes in app settings are worth it to allow developers to quickly get up and running without having to pass a appsettings.developer.json file around. If you think it should be removed I will do that.

src/Diagnostics.DataProviders/DataProviderConstants.cs
craigmaher: I was planning to separate prod from dev as this is only Dev. and the key vault is loaded at the same time as the configuration files, there is a workaround but it would result in a slow down of startup times.

craigmaher: would you still like me to make the constant come from AppSettings?

src/Diagnostics.RuntimeHost/appsettings.json
ShekharGupta1988: Ya, earlier we used to have this but we decided to keep the app setting file clear in order to not leak the secrets accidentally. In general, I feel its a good practice to keep the app settings clean and have environment specific overrides.

With this PR, developers wont even need appsettings.developer.json file right?

src/Diagnostics.DataProviders/DataProviderConstants.cs
ShekharGupta1988: I believe you would have to do that eventually since this endpoint would need to be configurable for dev/prod/mooncake

ShekharGupta1988:
@craigmaher  - do you mean to take this change to master or master2 for Liberation?

src/Diagnostics.DataProviders/ConfigurationFactory.cs
craigmaher: This uses multiple methods of authentication. one of which is a modal box another is visual studio credentials and in azure you can use the resource ID.

src/Diagnostics.RuntimeHost/appsettings.json
craigmaher: The aim is to not need a appsettings.developer.json file but if we keep app setting file clear then we will need it. I want to check in at the least an example app settings that when used  should not require anything new.

craigmaher:
> @craigmaher - do you mean to take this change to master or master2 for Liberation?
I have updated this.

PR #198, opened 2019-06-17
Fix Runtime Host Requesting Support Bay API Routes from Observer

dagrooms52:
- Fix routes looking for `/observer` in Observer
- Observer does not have a route for `subscriptions/{}/hostingEnvironments/{}/deployments`, send those to support-bay-api for now

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
@@ -67,9 +67,9 @@ namespace Diagnostics.DataProviders
             {
                 return GetWawsObserverResourceAsync(uri);
             }
-            else if (Configuration.ObserverLocalHostEnabled && TryGetLocalObserverRoute(uri, out var newUri))
+            else if (Configuration.ObserverLocalHostEnabled)
             {
-                return GetWawsObserverResourceAsync(newUri);
+                return GetWawsObserverResourceAsync(ConvertToLocalObserverRoute(uri));
             }
             else
             {
@@ -77,22 +77,14 @@ namespace Diagnostics.DataProviders
             }
         }
 
-        private static bool TryGetLocalObserverRoute(Uri uri, out Uri localObserverUri)
+        private static Uri ConvertToLocalObserverRoute(Uri uri)
         {
-            localObserverUri = uri;
-
             if (!uri.AbsolutePath.StartsWith("/observer"))
             {
-                return true;
+                return uri;
             }
 
-            if (uri.AbsolutePath.Contains("subscriptions") && (uri.AbsolutePath.EndsWith("webspaces") || uri.AbsolutePath.EndsWith("sites")))
-            {
-                localObserverUri = new Uri(uri, uri.AbsolutePath.Remove(0, "/observer".Length));
-                return true;
-            }
-
-            return false;
+            return new Uri(uri, uri.AbsolutePath.Remove(0, "/observer".Length));
         }
 
         private async Task<dynamic> GetWawsObserverResourceAsync(Uri uri)
--- a/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
@@ -63,14 +63,36 @@ namespace Diagnostics.DataProviders
                 throw new FormatException(exceptionMessage, ex);
             }
 
-            if (uri.Host.Contains(allowedHosts[0]) || uri.Host.Contains(allowedHosts[1]) || Configuration.ObserverLocalHostEnabled)
+            if (uri.Host.Contains(allowedHosts[0]) || uri.Host.Contains(allowedHosts[1]))
             {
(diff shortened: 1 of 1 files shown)


src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -63,16 +63,38 @@ public Task<dynamic> GetResource(string resourceUrl)
...
+        private static bool TryGetLocalObserverRoute(Uri uri, out Uri localObserverUri)
+        {
+            localObserverUri = uri;
+
+            if (!uri.AbsolutePath.StartsWith("/observer"))
+            {
+                return true;
+            }
+
+            if (uri.AbsolutePath.Contains("subscriptions") && (uri.AbsolutePath.EndsWith("webspaces") || uri.AbsolutePath.EndsWith("sites")))
```
hforeste: Why only a subset of support-bay-api URLs?

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -63,16 +63,38 @@ public Task<dynamic> GetResource(string resourceUrl)
...
+            {
+                localObserverUri = new Uri(uri, uri.AbsolutePath.Remove(0, "/observer".Length));
```
hforeste: If I am following the logic correctly, in the case that the Uri is something like `"https://support-bay-api.azurewebsites.net/observer/subscriptions/.../webspaces/..."` then this will make the localObserverUri equal the absolute path of the support-bay-api URL absent the `/observer` string. Once this request object make it eventually to `SendObserverRequestAsync` then the HttpClient base URI to the LocalHostObserver will be overided with this Uri's absolute path and the HttpClient will send the request to the SupportBayApi.

Am I understanding this correctly? Is this the intentional behavior?

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -63,16 +63,38 @@ public Task<dynamic> GetResource(string resourceUrl)
                 throw new FormatException(exceptionMessage, ex);
             }
 
-            if (uri.Host.Contains(allowedHosts[0]) || uri.Host.Contains(allowedHosts[1]) || Configuration.ObserverLocalHostEnabled)
+            if (uri.Host.Contains(allowedHosts[0]) || uri.Host.Contains(allowedHosts[1]))
             {
                 return GetWawsObserverResourceAsync(uri);
```
hforeste: So it looks like even without checking if `Configuration.ObserverLocalHostEnabled`, the logic will still send the request to LocalHostObserver in `SendObserverRequestAsync` because we strip the base URL in `GetWawsObserverResourceAsync` and append the URL path to the HttpClient base URI. Did you remove this in the if clause because it was redundant or some other reason?

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
dagrooms52: `/api/subscriptions/{}/webspaces` & `api/subscriptions/{}/sites` work in selfhost Observer
`/api/subscriptions/{}/hostingEnvironments/{}/deployments` isn't found in selfhost Observer, 404

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
dagrooms52: Yes, exactly

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
dagrooms52: If the observer url is requested (`allowedHosts[0 or 1]`) then we can `GetWawsObserverResource` no matter what.
Else, we have a support-bay-api route, some of which can't hit the selfhost Observer. I check if I can convert the route; if yes, convert and send to Observer. If not, send back to support-bay-api

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
hforeste: Can we move the business logic of this API from support-bay-api into the DiagnosticController on SelfHostObserver. I believe I built this API long ago and it is a a few lines of code.

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
hforeste: Can we verify what APIs are already there in the SelfHostObserver? I did some work sometime back where I migrated a good chunk of APIs that existing detectors use from support-bay-api into SelfHostObserver

hforeste:
Can you pull the latest changes from master? I'd like to see your changes with the recent changes you made to SupportObserverDataProviderBase like the one you did for the minienvironent call.

PR #199, opened 2019-06-18
Use Service Name in DataSource Timeout Message

dagrooms52:
- Broke log decorator into one class per data source interface
- Added code analysis to each project


hforeste:
Hey Daniel, did the team agree in a scrum awhile ago to not have any code analyzers?

dagrooms52:
> Hey Daniel, did the team agree in a scrum awhile ago to not have any code analyzers?

We agreed to take out StyleCop since it complains about style without providing automatic fixes. I've only added the basic analyzers that are needed to see that variables are unused or make reminders to check inputs for null values before usage. These analyzers are actually useful because they present real issues with the code, and can be fixed automatically with the shortcut `ctrl + .`

hforeste:
What is the context behind separating into different classes?

dagrooms52:
> What is the context behind separating into different classes?

Mainly separation of concerns; this makes it easier to tell which interfaces are being implemented where (regions do not programatically encapsulate the implementations), and also makes sure that private data that only the LogDecoratorBase is concerned with stays in the base class. 

Line 41 of LogDecoratorBase is `var dataSourceName = this.GetType().Name;` - this will get the name of the calling LogDecorator, so in the case of a TimeoutException we will see a message like `"DataSource timed out: Kusto"` or `"Observer"`, making it clear which DataSource is having issues to make our logging more direct.

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
```diff
@@ -29,8 +28,8 @@ public async Task Invoke(HttpContext httpContext)
         {
             Exception exception = null;
             int statusCode = 0;
-            GenerateMissingRequestHeaders(ref httpContext);
-            BeginRequest_Handle(httpContext);
+            GenerateMissingRequestHeaders(httpContext);
```
hforeste: I think we should be good here, the httpContext itself I think is already being passed by reference to the method, so `GenerateMissingRequestHeaders` will have no problem modifying the http context right?

src/Diagnostics.DataProviders/DataProviders.cs
```diff
@@ -17,20 +17,20 @@ public class DataProviders
...
         {
-            Kusto = new DataProviderLogDecorator(context, new KustoDataProvider(_cache, context.Configuration.KustoConfiguration, context.RequestId, context.KustoHeartBeatService));
-            Observer = new DataProviderLogDecorator(context, SupportObserverDataProviderFactory.GetDataProvider(_cache, context.Configuration, context));
-            GeoMaster = new DataProviderLogDecorator(context, new GeoMasterDataProvider(_cache, context));
-            AppInsights = new DataProviderLogDecorator(context, new AppInsightsDataProvider(_cache, context.Configuration.AppInsightsConfiguration));
-            ChangeAnalysis = new DataProviderLogDecorator(context, new ChangeAnalysisDataProvider(_cache, context.Configuration.ChangeAnalysisDataProviderConfiguration, context.RequestId, context.clientObjectId, context.clientPrincipalName, Kusto));
-            Asc = new DataProviderLogDecorator(context, new AscDataProvider(_cache, context.Configuration.AscDataProviderConfiguration, context.RequestId));
+            Kusto = new KustoLogDecorator(context, new KustoDataProvider(_cache, context.Configuration.KustoConfiguration, context.RequestId, context.KustoHeartBeatService));
```
hforeste: I'm pretty sure we're okay but just want to double check that this doesn't need detectors to recompile.

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
dagrooms52: Right, the `ref` was unnecessary since the reference itself is not being reassigned, just the underlying properties.

src/Diagnostics.DataProviders/DataProviders.cs
dagrooms52: Tested detectors on Windows and Linux apps, all worked without recompiling

src/Diagnostics.DataProviders/LogDecoratorBase.cs
```diff
@@ -0,0 +1,85 @@
...
+                else
+                {
+                    var dataSourceName = this.GetType().Name;
+                    var logDecoratorSuffix = "LogDecorator";
+
+                    if (dataSourceName.EndsWith(logDecoratorSuffix, StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        dataSourceName = dataSourceName.Substring(0, dataSourceName.Length - logDecoratorSuffix.Length);
+                    }
+                    throw new TimeoutException($"DataSource timed out: {dataSourceName}");
```
hforeste: Can we say `DataSource timed out: {dataSourceName}. DataProviderOperation: {dataProviderOperation}`. Observer and Geomaster have multiple operations so this will boil things down to which operations might be timing out in those data providers and if we want to look at it from a higher level we can just parse our the DataSourceName in Kusto and summarize on top of that.

PR #200, opened 2019-06-19
Direct /minienvironment Routes to WawsObserver

dagrooms52:
- SelfHost Observer cannot pull Georegion connection string from geomaster
- WawsObserver gets the connection string from Service Health Dashboard, a dependency we don't want to add to SelfHost Observer

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
@@ -96,16 +96,19 @@ namespace Diagnostics.DataProviders
             }
 
             var request = new HttpRequestMessage(HttpMethod.Get, url);
+            HttpResponseMessage response;
 
             // TODO: remove redirect to wawsobserver when Geomaster API implements GeoRegion connection strings
-            if (url.StartsWith("minienvironments/"))
+            if (url.StartsWith("minienvironments/") && Configuration.ObserverLocalHostEnabled)
             {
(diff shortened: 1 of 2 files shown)


src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -96,16 +96,19 @@ protected async Task<string> GetObserverResource(string url, string resourceId =
...
+            HttpResponseMessage response;
 
             // TODO: remove redirect to wawsobserver when Geomaster API implements GeoRegion connection strings
-            if (url.StartsWith("minienvironments/"))
+            if (url.StartsWith("minienvironments/") && Configuration.ObserverLocalHostEnabled)
             {
-                url = $"api/{url}";
-                var uri = new Uri(new Uri("https://wawsobserver.azurewebsites.windows.net"), url);
-                request = new HttpRequestMessage(HttpMethod.Get, uri);
+                request = new HttpRequestMessage(HttpMethod.Get, $"/api/{url}");
```
hforeste: Since you already set the baseAddress with `/api`, you wont need to add the `/api` prefix to your request message right?

dagrooms52: Apparently the httpclient ignores the route set in BaseAddress - I was getting 406 "Unacceptable" without the `/api` prefix. I'm just going to remove `/api` from the baseurl set on the httpclient.

PR #202, opened 2019-06-29
Change TenantIdQuery to also query with EventStampName column

hforeste:
Cloud service domain name suffix is different across clouds, instead of coding for multiple domain names eg., `*.cloudapp.net, *.chinacloudapp.cn` and etc, we should use EventStampName instead which is the SourceMoniker equivalent.

### DIFF
--- a/src/Diagnostics.RuntimeHost/Services/ResourceService/NationalCloudStampService.cs
+++ b/src/Diagnostics.RuntimeHost/Services/ResourceService/NationalCloudStampService.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Diagnostics.DataProviders;
+using Diagnostics.ModelsAndUtils.Attributes;
+
+namespace Diagnostics.RuntimeHost.Services
+{
+    public sealed class NationalCloudStampService : StampService
+    {
+        protected override async Task<List<string>> GetTenantIdsAsync(string stamp, DateTime startTime, DateTime endTime, DataProviderContext dataProviderContext, PlatformType platformType)
+        {
+            if (platformType == PlatformType.Windows)
+            {
+                return await base.GetTenantIdsAsync(stamp, startTime, endTime, dataProviderContext, platformType);
+            }
+            else
+            {
+                return await Task.FromResult(new List<string>());
+            }
+        }
+
+        protected override string GetTenantIdQuery(string stamp, string startTimeStr, string endTimeStr, string tableName)
+        {
+            return
+                $@"{RoleInstanceHeartBeatTableName}
+                | where TIMESTAMP >= datetime({startTimeStr}) and TIMESTAMP <= datetime({endTimeStr}) and where EventStampName =~ ""{stamp}""
+                | summarize by Tenant, PublicHost";
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


src/Diagnostics.RuntimeHost/Services/ResourceService/StampService.cs
```diff
@@ -104,7 +104,7 @@ private string GetTenantIdQuery(string stamp, DateTime startTime, DateTime endTi
             return
                 $@"{tableName}
                 | where TIMESTAMP >= datetime({startTimeStr}) and TIMESTAMP <= datetime({endTimeStr})
-                | where PublicHost =~ ""{stamp}.cloudapp.net"" or PublicHost matches regex ""{stamp}([a-z{{1}}]).cloudapp.net""
+                | where PublicHost =~ ""{stamp}.cloudapp.net"" or PublicHost matches regex ""{stamp}([a-z{{1}}]).cloudapp.net"" or EventStampName =~ ""{stamp}""
```
ShekharGupta1988: Just do a quick check for:

- **Mega stamps** - I think there is another column called 'EventPrimaryStampName' that will hold the mega stamp name and not this column. This 'EventStampName' would have something like 'waws-prod-sn1-107a'
Ex:
```
AntaresIISLogFrontEndTable
| where TIMESTAMP > ago(20m) and EventStampName  == 'waws-prod-sn1-107a'
| summarize by EventStampName, EventPrimaryStampName
```

  |   |  
-- | -- | --
  | **EventStampName** | **EventPrimaryStampName**
  | waws-prod-sn1-107a | waws-prod-sn1-107



- **Sites in ASE** - Make sure you are seeing the right name in this column which is passed in the stamp parameter to this method and there is no friendly name issue.

src/Diagnostics.RuntimeHost/Services/ResourceService/NationalCloudStampService.cs
```diff
@@ -0,0 +1,31 @@
...
+            {
+                return await Task.FromResult(new List<string>());
+            }
+        }
+
+        protected override string GetTenantIdQuery(string stamp, string startTimeStr, string endTimeStr, string tableName)
+        {
+            return
+                $@"{RoleInstanceHeartBeatTableName}
+                | where TIMESTAMP >= datetime({startTimeStr}) and TIMESTAMP <= datetime({endTimeStr}) and where EventStampName =~ ""{stamp}""
```
ShekharGupta1988: will this work for mega stamp?

hforeste: There are no megastamps in China

ShekharGupta1988: you should use EventStampPrimary column
in that way, mega stamps are also taken care of..

hforeste: So this column actually does not even exist there in China Kusto

ShekharGupta1988: ok.. cool. I think for now, its fine then.. Approved the PR

PR #204, opened 2019-07-09
Add Kusto Query operationName to detector reponse

puneetg1983:


src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -152,7 +157,8 @@ public async Task<KustoQuery> GetKustoQueryAsync(string query, string cluster, s
                 {
                     Text = query,
                     Url = $"https://dataexplorer.azure.com/clusters/{cluster}/databases/{database}?query={encodedQuery}",
-                    KustoDesktopUrl = $"https://{cluster}.kusto.windows.net:443/{database}?query={encodedQuery}&web=0"
+                    KustoDesktopUrl = $"https://{cluster}.kusto.windows.net:443/{database}?query={encodedQuery}&web=0",
```
dagrooms52: We should pull `Url` and `KustoDesktopUrl` from `appsettings.json` for easier configuration management; it already contains a key for `wawsus.kusto.windows.net`

PR #206, opened 2019-07-10
White list function disable settings

cindylovescoding:
To get the function level disable information from appsettings, currently we can only get this information by doing this change.

The settings will be like AzureWebJobs.{functionName}.Disabled.
Add the logic to whitelist this too.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Diagnostics.ModelsAndUtils.Models;
 using Newtonsoft.Json;
+using System.Text.RegularExpressions;
 
 namespace Diagnostics.DataProviders
 {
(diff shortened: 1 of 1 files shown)


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -80,12 +82,11 @@ private IGeoMasterClient InitClient()
             Dictionary<string, string> appSettings = new Dictionary<string, string>();
             foreach (var item in properties)
             {
-                if (AllowedlistAppSettingsStartingWith.Any(x => item.Key.StartsWith(x)))
+                if (AllowedlistAppSettingsStartingWith.Any(x => item.Key.StartsWith(x)) && !SensitiveAppSettingsEndingWith.Any(x => item.Key.EndsWith(x))
```
hforeste: Is `StartsWith` and `EndsWith` case agnostic? Just asking in case property keys in the future would be a different casing.

puneetg1983: While it is not a bad suggestion, it is highly unlikely that casing for some of these app settings will change in future just because a lot of code (Antares and outside Antares) has now a dependency on these settings ;-)

hforeste: > While it is not a bad suggestion, it is highly unlikely that casing for some of these app settings will change in future just because a lot of code (Antares and outside Antares) has now a dependency on these settings ;-)

Yea I was thinking the same too, its just one of those really small things that you don't want to bite you in the butt later  ¯\_(ツ)_/¯

PR #210, opened 2019-07-16
[DiagRoleLiberation] RuntimeHost auth with CompilerHost

rekhaswaminathan:
- RuntimeHost interacts with CompilerHost using auth token

- CompilerHost is protected to allow access only for RuntimeHost. In future if we want to allow another App, we just need to add the App Id to the settings

- Pulling the config from appsettings.Development.json for running locally


src/Diagnostics.CompilerHost/appsettings.json
```diff
@@ -11,5 +11,10 @@
         "Default": ""
       }
     }
+  },
+  "AzureAd": {
+    "AADAuthority": "https://login.microsoftonline.com/microsoft.onmicrosoft.com",
+    "ClientId": "cda572d8-1f3e-4119-9a11-0e4ceb2fb3e1", // Azure AAd App id of CompilerHost
```
ShekharGupta1988: lets remove the values of AppIds from AppSettings and keep this file clean.
Hopefully we would move to keyvault after Craig's work.

src/Diagnostics.CompilerHost/Startup.cs
```diff
@@ -38,6 +51,41 @@ public Startup(IConfiguration configuration)
...
+                        ValidateLifetime = true,
+                        RequireSignedTokens = true,
+                        IssuerSigningKeys = signingKeys
+                    };
+
+                    options.Events = new JwtBearerEvents
+                    {
+                        OnTokenValidated = context =>
+                        {
+                            var allowedAppIds = Configuration["AzureAd:AllowedAppIds"].Split(",").ToList();
```
ShekharGupta1988: Just apply Trimming also to ensure no white spaces left accidentally.

src/Diagnostics.DataProviders/DataSourcesConfiguration.cs
```diff
@@ -19,5 +19,7 @@ public class DataSourcesConfiguration
         public AntaresMdmDataProviderConfiguration AntaresMdmConfiguration { get; set; }
 
         public NetworkingMdmDataProviderConfiguration NetworkingMdmConfiguration { get; set; }
+
+        public CompilerHostAuthConfiguration CompilerHostAuthConfiguration { get; set; }
```
ShekharGupta1988: Compiler host is not a data provider. Don't think its a good idea to take dependencies on data provider project for this one. Structurally also, it will create confusion going forward.

src/Diagnostics.CompilerHost/appsettings.json
rekhaswaminathan: Should we remove in this PR itself? Because once its merged, it wouldn't work. I thought of maintaining appsettings.Development.json, but that would be redundant again.

src/Diagnostics.CompilerHost/Startup.cs
```diff
@@ -21,9 +29,14 @@ public class Startup
         /// Initializes a new instance of the <see cref="Startup"/> class.
         /// </summary>
         /// <param name="configuration">The configuration.</param>
-        public Startup(IConfiguration configuration)
+        public Startup(IHostingEnvironment hostingEnvironment)
         {
-            Configuration = configuration;
+            var builder = new ConfigurationBuilder()
+              .SetBasePath(Directory.GetCurrentDirectory())
+              .AddJsonFile($"appsettings.json", optional: false, reloadOnChange: true)
```
hforeste: I believe I read this somewhere but doing `.AddJsonFile` followed by `.AddEnvironmentVariables` means the .NET Core runtime will first go to appsettings.json file to get settings and if not found then environment variables next. I was working on a project onetime and I deployed my .NET Core app with Visual Studio publish which included the appsettings.json file. When I ran the web app, I was confused where it was getting these different settings than what I had set in the App Service AppSettings. Its not the biggest deal but it might save a headache in the future by changing the order.

src/Diagnostics.CompilerHost/Startup.cs
```diff
@@ -38,6 +51,41 @@ public Startup(IConfiguration configuration)
...
+                    options.Events = new JwtBearerEvents
+                    {
+                        OnTokenValidated = context =>
+                        {
+                            var allowedAppIds = Configuration["AzureAd:AllowedAppIds"].Split(",").ToList();
+                            var claimPrincipal = context.Principal;
+                            var appId = claimPrincipal.Claims.FirstOrDefault(c => c.Type.Equals("appid", StringComparison.CurrentCultureIgnoreCase));
+                            if (appId == null || !allowedAppIds.Exists(p => p.Equals(appId.Value, StringComparison.OrdinalIgnoreCase)))
+                            {
+                                context.Fail($"App Id : {appId.Value} is not in the allowed list of App Ids");
```
hforeste: I haven't used `context.Fail()`. Does this shortcircuit the request pipeline and give a 401 Unauthorized?
Also maybe we shouldn't say `App Id : {appId.Value} is not in the allowed list of App Ids`. I think that is too much info for a hacker to know. I think just a 401 Unauthorized is good enough.

rekhaswaminathan: Yes it will give 401 authorized, I tried it with a different App id and got 401. and yes, I will change the message

src/Diagnostics.CompilerHost/appsettings.json
ShekharGupta1988: I think using Development.JSON is a better option which you are not going to check-in. Keeping this file just as a schema for config should be good.

ShekharGupta1988: Anyhow, it would be temporary as we should rely on Craig's work for keyvault

src/Diagnostics.DataProviders/TokenService/CompilerHostTokenService.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Microsoft.IdentityModel.Clients.ActiveDirectory;
```
ShekharGupta1988: For this file also, we should not create this in Data Provider project.

src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -82,5 +82,11 @@
   },
   "SearchAPI": {
     "SearchAPIEnabled": false
+  },
+  "CompilerHostAuth": {
```
ShekharGupta1988: If I recall, there is a section for Compiler Host settings in this file. It will be nice to move these settings with that section so that all Compiler host settings are together.

ShekharGupta1988:
Looks great @rekhaswaminathan )

PR #212, opened 2019-07-18
Added production keyvault. Made prod rely on app settings

craigmaher:
Added production key vault. Made everything rely on app settings and key vault for values.


src/Diagnostics.DataProviders/ConfigurationFactory.cs
```diff
@@ -37,9 +37,18 @@ public AppSettingsDataProviderConfigurationFactory(IHostingEnvironment env)
                 )
             );
 
-            builder.AddAzureKeyVault($"https://{builtConfig["DevKeyVaultName"]}.vault.azure.net/",
-                                     keyVaultClient,
-                                     new DefaultKeyVaultSecretManager());
+            if (env.IsProduction())
```
dagrooms52: You can set `ProdKeyVaultName` or `DevKeyVaultName` based on `IsProduction()` to avoid copying code (and the magic strings) here

src/Diagnostics.RuntimeHost/Program.cs
```diff
@@ -24,15 +24,23 @@ public static IWebHost BuildWebHost(string[] args)
...
 
-                    if (context.HostingEnvironment.IsDevelopment())
-                    {
+                    
                         var azureServiceTokenProvider = new AzureServiceTokenProvider();
                         var keyVaultClient = new KeyVaultClient(
                             new KeyVaultClient.AuthenticationCallback(
                                 azureServiceTokenProvider.KeyVaultTokenCallback));
 
+                    if (context.HostingEnvironment.IsDevelopment())
```
dagrooms52: Same here, can avoid copying code by just picking DevKeyVaultName / ProdKeyVaultName
May be worth making a static utility function to make the `$"https://{KeyvaultName}.vault.azure.net/"`

src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -52,14 +52,9 @@ public void ConfigureServices(IServiceCollection services)
...
-            if (Environment.IsProduction())
-            {
-                searchIsEnabled = Convert.ToBoolean(Registry.GetValue(RegistryConstants.SearchAPIRegistryPath, RegistryConstants.SearchAPIEnabledKey, string.Empty));
-            }
-            else
-            {
-                searchIsEnabled = Convert.ToBoolean(Configuration[$"SearchAPI:{RegistryConstants.SearchAPIEnabledKey}"]);
-            }
+            
+            searchIsEnabled = Convert.ToBoolean(Configuration[$"SearchAPI:{RegistryConstants.SearchAPIEnabledKey}"]);
```
dagrooms52: This should be a `TryParse` style operation if you're not checking for `FormatException`

PR #213, opened 2019-07-18
load detectors on startup

craigmaher:


src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -86,6 +86,9 @@ public void ConfigureServices(IServiceCollection services)
             ChangeAnalysisTokenService.Instance.Initialize(dataSourcesConfigService.Config.ChangeAnalysisDataProviderConfiguration);
             AscTokenService.Instance.Initialize(dataSourcesConfigService.Config.AscDataProviderConfiguration);
             CompilerHostTokenService.Instance.Initialize(Configuration);
+
+            // Initialize on startup
+            servicesProvider.GetService<ISourceWatcherService>();
```
hforeste: How does this line start the source watcher to download the detectors?

craigmaher: This line forces the initializes of the service Singleton SourceWatcherService which upon initialization will in turn initialize a GitHubWatcher or a LocalFileSystemWatcher.

The line more specifically requests the ISourceWatcherService and since the singleton has not been initialized yet it will initialize the ISourceWatcherService so it can return the ISourceWatcherService

dagrooms52:
I've been having trouble with long startup times on RuntimeHost, this will probably also slow it down. We need to architect a way to pull these resources continually outside of Startup, I'm trying out [IHostedService](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-2.2&tabs=visual-studio)

ShekharGupta1988:
I had a discussion with @craigmaher regarding this one.
This is a temporary solution for now. 
Craig's idea was for now, we will use slot to deploy the code first and let all detectors load up there.
Then do the swap to PROD.
@craigmaher  - do note that even swap can cause process to recycle in some conditions like app setting update, local cache etc.

A long term solution we discussed sometime back to prep the deployment package in such a way that wwwroot already have these detectors folder present. So the runtime host on startup just have to load them in memory and not download all of them.

PR #214, opened 2019-07-18
Protect runtimehost with AAD

rekhaswaminathan:
- Runtimehost is protected with aad to allow request only from applens
- Healthping controller has not changed so that we can test it to see if the app is up
- In the future sprint, runtimehost will be protected to allow requests from partner app ids as well.

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
@@ -8,11 +8,9 @@ using Diagnostics.ModelsAndUtils.Models;
 using Diagnostics.RuntimeHost.Models;
 using Diagnostics.RuntimeHost.Services;
 using Diagnostics.RuntimeHost.Utilities;
-using Microsoft.AspNetCore.Authorization;
 
 namespace Diagnostics.RuntimeHost.Controllers
 {
-    [Authorize]
     public abstract class SiteControllerBase : DiagnosticControllerBase<App>
     {
         protected ISiteService _siteService;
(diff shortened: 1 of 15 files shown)


src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
```diff
@@ -8,9 +8,11 @@
 using Diagnostics.RuntimeHost.Models;
 using Diagnostics.RuntimeHost.Services;
 using Diagnostics.RuntimeHost.Utilities;
+using Microsoft.AspNetCore.Authorization;
 
 namespace Diagnostics.RuntimeHost.Controllers
 {
+    [Authorize]
     public abstract class SiteControllerBase : DiagnosticControllerBase<App>
```
ShekharGupta1988: We add the tag to Abstract base class as well? Will this work?

src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -86,5 +86,10 @@
   },
   "SearchAPI": {
     "SearchAPIEnabled": false
+  },
+  "AzureAD": {
+    "AADAuthority": "https://login.microsoftonline.com/microsoft.onmicrosoft.com",
+    "ClientId": "b9a1efcd-32ee-4330-834c-c04eb00f4b33",
```
ShekharGupta1988: Lets keep app settings clean.

src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -86,5 +86,10 @@
   },
   "SearchAPI": {
     "SearchAPIEnabled": false
+  },
+  "AzureAD": {
```
ShekharGupta1988: how about renaming this to something more apt like 'AccessControlSettings' or 'SecuritySettings', something like that.

src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
hforeste: You also have authorize attribute on the `SitesController` class so I assume you won't need this here.

PR #215, opened 2019-07-30
Point users to main cluster instead of follower in data sources

rekhaswaminathan:


dagrooms52:
I talked to Rekha - we should follow the configured fallback for the cluster instead of taking "follower" off the name

puneetg1983:
Do we maintain a mapping of follower and primary cluster?

rekhaswaminathan:
Yes we do. Refer: https://github.com/Azure/Azure-AppServices-Diagnostics/blob/a8f2b4744f39c184a61af2f76da50a7445bcad7d/src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs#L35

PR #216, opened 2019-07-31
Remove PII for app settings value

cindylovescoding:
There could be SAS Uri value for the settings "WEBSITE_RUN_FROM_PACKAGE". Mask the SAS URI before returning it to detector.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -51,7 +51,7 @@ namespace Diagnostics.DataProviders
         public string RemovePIIFromSettings(string content)
         {
             // Mask SAS Uri
-            if (Regex.Match(content, @"https*:\/\/.*\.core\.windows\.net.*sig=.*", RegexOptions.IgnoreCase).Success)
+            if (Regex.Match(content, @"https*:\/\/[\w.]*[\w]+.core.windows.net?.*sig=.*", RegexOptions.IgnoreCase).Success)
             {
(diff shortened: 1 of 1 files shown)


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -48,6 +48,17 @@ private IGeoMasterClient InitClient()
             return geoMasterClient;
         }
 
+        public string RemovePIIFromSettings(string content)
```
dagrooms52: I'd make this static and add to a static utility class

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -48,6 +48,17 @@ private IGeoMasterClient InitClient()
...
+        {
+            // Mask SAS Uri
+            if (Regex.Match(content, @"https*:\/\/.*\.core\.windows\.net.*sig=.*", RegexOptions.IgnoreCase).Success)
```
nmallick1: Nitpick.. 
Although the above regex does the job as it matches all the requirements of a webapp, possibly use this following Regex ?

https*:\/\/[\w\.]*[\w]+\.core\.windows\.net\?.*sig=.*

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
cindylovescoding: So this function is a specific scenario to the GetAppSettings of this class . Not sure whether it will be useful to be added to a static utility class.

PR #220, opened 2019-08-05
Fix for VMSS tenant ids

nmallick1:
For sites hosted on VMSS Utilities.TenantId was returning an empty collection as a result queries would fail. Fixed the query to fetch TenantId's for VMSS hosted sites.

Per Phil though, in VMSS, tenant ID's will be irrelevant because every instance will have a unique name so we might have to change detectors to consider this.


src/Diagnostics.RuntimeHost/Services/ResourceService/StampService.cs
```diff
@@ -114,7 +114,7 @@ protected virtual string GetTenantIdQuery(string stamp, string startTimeStr, str
             return
                 $@"{tableName}
                 | where TIMESTAMP >= datetime({startTimeStr}) and TIMESTAMP <= datetime({endTimeStr})
-                | where PublicHost =~ ""{stamp}.cloudapp.net"" or PublicHost matches regex ""{stamp}([a-z{{1}}]).cloudapp.net""
+                | where PublicHost =~ ""{stamp}.cloudapp.net"" or PublicHost matches regex ""{stamp}([a-z{{1}}]).cloudapp.net"" or PublicHost matches regex ""{stamp}-([a-z0-9\\.]+).cloudapp.azure.com""
```
hforeste: nitpik: Can you create a const string in this method that just labels what this regex is doing. Something like `const string vmssRegexPublicHost` and if you can also do that for the megastamp regex too just so its readable.

nmallick1: Added the strings

PR #221, opened 2019-08-05
Cmaher/compiler host key vault

craigmaher:
Added a key vault for compiler Host


src/Diagnostics.CompilerHost/appsettings.json
```diff
@@ -1,4 +1,8 @@
 {
+  "Secrets": {
+    "DevKeyVaultName": "Diag-CompilerHost",
```
hforeste: Is runtimehost and compilerhost sharing the same keyvault?

craigmaher: No they are going to use different key vaults as a security measure to give apps least privileged access.

dagrooms52: It's a good idea but perhaps overkill even by Azure KV's best practices: https://docs.microsoft.com/bs-latn-ba/Azure/key-vault/key-vault-best-practices#use-separate-key-vault
Given we have 4 apps in 3 environments, this will immediately take 12 keyvaults with 3 new keyvaults every time we add a service.
It's likely we'll just shutdown and reset all our service keys if they're leaked anyway.

rekhaswaminathan: If we use the same KV there are chances of running into name conflict for some app settings and also chance that one web app may have access to another web app's secrets.

dagrooms52: What kind of app settings could conflict? If it's KustoKey for example (although they could share this), why not store "CompilerHost.KustoKey" and "RuntimeHost.KustoKey"?
What's the harm of our apps accessing secrets for our other apps vs. the ability to revoke all key access in a single action?

PR #222, opened 2019-08-09
Use cert to talk to Geo

rekhaswaminathan:
Load cert from Keyvault on app startup and use that to talk to Geo

### DIFF
--- a/src/Diagnostics.DataProviders/GeoCertLoader.cs
+++ b/src/Diagnostics.DataProviders/GeoCertLoader.cs
@@ -1,0 +1,33 @@
+﻿using System;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Azure.KeyVault;
+using Microsoft.Azure.Services.AppAuthentication;
+using Microsoft.Extensions.Configuration;
+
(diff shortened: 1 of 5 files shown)


src/Diagnostics.DataProviders/GeoCertLoader.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public X509Certificate2 GeoCert { get; private set; }
+        
+        public void LoadCertFromKeyVault(IConfiguration configuration)
+        {
+            try
+            {
+                var azureServiceTokenProvider = new AzureServiceTokenProvider();
+                var prodKeyVaultClient = new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(azureServiceTokenProvider.KeyVaultTokenCallback));
+                var certificateSecret = prodKeyVaultClient.GetSecretAsync($"https://{configuration["Secrets:ProdKeyVaultName"]}.vault.azure.net", configuration["GeoMaster:CertificateName"]).Result;
```
ShekharGupta1988: lets use Async await instead of .Result

PR #229, opened 2019-08-16
Auto stamp region to kusto mapping

nmallick1:
If we do not already have the stamp to region mapping in Kusto, populate it from Analytics DB


nmallick1:
> Would it ever occur that a mapping is not in the kusto table?
> 
> I'm a little worried about the performance of this in the case where a mapping is not available we would impact performance of things by running a kusto query every time.

This table is used in billing so the mapping should always be present. I tested the performance of this query and it typically executes in under a second. Since we are adding the results of this query to the Dictionary, it will only be executed when a mapping isn't already present.

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -59,10 +60,41 @@ public string GetClusterNameFromStamp(string stampName)
 
             if (!_configuration.RegionSpecificClusterNameCollection.TryGetValue(appserviceRegion.ToLower(), out kustoClusterName))
             {
+                if(_configuration.CloudDomain == DataProviderConstants.AzureCloud)
+                {
+                    try
+                    {
+                         DataTable regionMappings = _kustoDataProvider.ExecuteClusterQuery($"cluster('wawseusfollower').database('wawsprod').WawsAn_regionsincluster | where pdate >= ago(5d) | summarize by Region, ClusterName", "RegionMappingInit").Result;
```
hforeste: `GetClusterNameFromStamp` gets called in the request pipeline by `KustoDataProvider` in the `ExecuteQuery` method. Using the `Task.Result` could block the thread. Maybe make this method an asynchronous task.

nmallick1: made the change

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -52,17 +53,48 @@ public KustoHeartBeatService(KustoDataProviderConfiguration configuration)
...
+                {
+                    try
+                    {
+                         DataTable regionMappings = await _kustoDataProvider.ExecuteClusterQuery($"cluster('wawseusfollower').database('wawsprod').WawsAn_regionsincluster | where pdate >= ago(5d) | summarize by Region, ClusterName", "RegionMappingInit");
+                        if (regionMappings.Rows.Count > 0)
+                        {
+                            foreach (DataRow dr in regionMappings.Rows)
+                            {
+                                _configuration.RegionSpecificClusterNameCollection.TryAdd(((string)dr["Region"]).ToLower(), ((string)dr["ClusterName"]).ToLower());
+                                _configuration.FailoverClusterNameCollection.TryAdd(((string)dr["Region"]).ToLower(), $"{((string)dr["ClusterName"])}follower");
```
hforeste: Switch this. Follower clusters is diagnostics main clusters and vice versa for the failover clusters.

hforeste:
Just curious, what is the context of this pull request?

nmallick1:
> Just curious, what is the context of this pull request?

With VMSS stamps, we need a mapping of which kusto cluster to go to for them. We can potentially update the mapping in our config, however since the rate at which VMSS / new stamps may be deployed is not known to us, there might be sites for which Applens will end up querying the wrong Kusto and return with no data. This will help us be ready for the VMSS move / new stamp deployment whenever it happens.

hforeste:
Okay, sounds good. Please look into the travis CI failure, otherwise this is a sign off from me.

PR #232, opened 2019-08-21
Log kusto response body in case of exception / error

nmallick1:
Log kusto response body in case of exception / error


src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -125,10 +127,12 @@ private DataTableResponseObjectCollection ProcessKustoResponse(string responseCo
             return JsonConvert.DeserializeObject<DataTableResponseObjectCollection>(responseContent);
         }
 
-        private void LogKustoQuery(string query, string cluster, string operationName, Stopwatch timeTakenStopWatch, string kustoClientId, Exception kustoApiException, DataTableResponseObjectCollection dataSet)
+        private void LogKustoQuery(string query, string cluster, string operationName, Stopwatch timeTakenStopWatch, string kustoClientId, Exception kustoApiException, DataTableResponseObjectCollection dataSet, string kustoResponse = "")
         {
             var status = kustoApiException == null ? "Success" : "Failed";
 
+            kustoResponse = (kustoResponse != "") ? $" {Environment.NewLine} --> KustoResponseBody : {kustoResponse} " : string.Empty ;
```
hforeste: nitpik: Since you already have this `NewLine` in the beginning maybe you don't need the `-->`. The new line will delimit the exception.ToString() and the response body plus you already prefix kustoResponse with `KustoResponseBody` so we should be able to easily figure out which is the kusto exception

hforeste: nitpik: Since you already have this `NewLine` in the beginning maybe you don't need the `-->`. The new line will delimit the exception.ToString() and the response body plus you already prefix kustoResponse with `KustoResponseBody` so we should be able to easily figure out which is the kusto exception

PR #241, opened 2019-09-03
Detector Loading performance Improvements

craigmaher:
Made changes to load detectors in parallel and load detectors on startup. Made changes so we can use environment variables in prod instead of keyvault for potentially faster load times.

Also made a change to allow us to only load public detectors.


ShekharGupta1988:
@craigmaher - lets update the app settings file as well with the new flag to load public detectors.

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Workers/GithubWorkerBase.cs
```diff
@@ -22,6 +25,15 @@ public abstract class GithubWorkerBase : IGithubWorker
...
         public abstract string Name { get; }
 
+        private readonly bool _loadOnlyPublicDetectors;
+
+        public GithubWorkerBase(bool loadOnlyPublicDetectors)
+        {
+            _loadOnlyPublicDetectors = loadOnlyPublicDetectors;
+        }
+
+        private static Regex regexPublicDetectors = new Regex(@"InternalOnly\s*=\s*false",RegexOptions.IgnoreCase);
```
ShekharGupta1988: Still a possibility with this RegEx to catch few internal detectors.

craigmaher: Catching a few internal detectors shouldn't slow things down too much, I would prefer to be too permissive then accidentally exclude some external detector.

PR #242, opened 2019-09-04
Adding additional properties to TableRendering

puneetg1983:


src/Diagnostics.ModelsAndUtils/Models/Rendering.cs
```diff
@@ -39,10 +39,20 @@ public class TableRendering : Rendering
 
         public string GroupByColumnName { get; set; }
 
+        public string DescriptionColumnName { get; set; }
+
+        public string Height { get; set; }
+
+        public dynamic TableOptions { get; set; }
```
ShekharGupta1988: What is the schema of these options?

puneetg1983: You can basically pass anything from the list @ https://swimlane.gitbook.io/ngx-datatable/api/table/inputs

Whatever we set here, the ux code just applies these settings \ properties on the ngx-datatable  component...

PR #243, opened 2019-09-04
Add collectVnetConfig to GeoMasterDataProvider

michimune:
This new API calls to a geomaster API which gets a collection of VNET related configurations for the specified subscription ID.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -207,6 +207,48 @@ namespace Diagnostics.DataProviders
         }
 
         /// <summary>
+        /// Gets all the networking configurations for the hosting environment
+        /// </summary>
+        /// <param name="subscriptionId">Subscription Id for the resource</param>
+        /// <param name="vnetResourceGroup">The resource group in which the VNet is a part of</param>
+        /// <param name="vnetName">Name of the VNET</param>
+        /// <param name="vnetSubnetName">Subnet name of the VNET</param>
+        /// <param name="cancellationToken">(Optional)</param>
+        /// <returns></returns>
+        /// <example>
+        /// This sample shows how you can call this function to get the NSG rules
+        /// for this App Service environment
+        /// <code>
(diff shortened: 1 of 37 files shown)


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -206,6 +206,48 @@ public async Task<VnetValidationRespone> VerifyHostingEnvironmentVnet(string sub
...
+        ///                   vnetRg,
+        ///                   vnetName,
+        ///                   subnet);
+        ///
+        /// }
+        /// </code>
+        /// </example>
+        public async Task<VnetConfiguration> CollectVirtualNetworkConfig(string subscriptionId, string vnetResourceGroup, string vnetName, string vnetSubnetName, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var path = string.Format(@"subscriptions/{0}/providers/Microsoft.Web/collectVnetConfiguration", subscriptionId);
```
dagrooms52: Style: `@` is only necessary for multi-line strings

src/Diagnostics.ModelsAndUtils/Models/GeoMaster/NetworkResourceProviderEntities.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        public string ProvisioningState { get; set; }
+        public string NextHopIpAddress { get; set; }
+        public string AddressPrefix { get; set; }
+        public string NextHopType { get; set; }
+    }
+
+    public class NetworkAddressSpace
+    {
+        public List<string> AddressPrefixes { get; set; }
```
dagrooms52: Let's use `IEnumerable` for iterative needs (foreach) or `IList` if this requires `.Add()`

michimune: That's right

hforeste:
Hey Michi, what do you think about resorting to using ASC to get networking details?

michimune:
It's definitely an option and it's already available, but it's sometimes slow (> 30 seconds). We expect this API to be faster.

PR #244, opened 2019-09-05
List all the appsettings keys from geomaster data provider

cindylovescoding:


nmallick1:
Are we sure we want to allow listing of all appsettings ? We had an allow list to make sure only certain settings can be retrieved via detector for security reasons. With the liberation project, we will no longer be limited with long deployment times.

PR #249, opened 2019-09-06
Cmaher/staging key vault

craigmaher:
allow  for a staging key vault

### DIFF
--- a/src/Diagnostics.CompilerHost/Startup.cs
+++ b/src/Diagnostics.CompilerHost/Startup.cs
@@ -42,7 +42,20 @@ namespace Diagnostics.CompilerHost
 
             var builtConfig = builder.Build();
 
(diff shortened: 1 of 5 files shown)


src/Diagnostics.CompilerHost/Startup.cs
```diff
@@ -42,7 +42,20 @@ public Startup(IHostingEnvironment hostingEnvironment)
 
             var builtConfig = builder.Build();
 
-            string keyVaultConfig = hostingEnvironment.IsProduction() ? "Secrets:ProdKeyVaultName" : "Secrets:DevKeyVaultName";
+            string keyVaultConfig;
+            switch (hostingEnvironment.EnvironmentName)
```
dagrooms52: This can be a static function that each of the 3 methods in this PR can call

PR #256, opened 2019-09-16
log startup failures to kusto

craigmaher:
This should enable startup failures to be logged to Kusto.


dagrooms52:
Were you able to test this at all?

craigmaher:
> Were you able to test this at all?

Yes, I was able to verify that LogRuntimeHostUnhandledException and LogCompilerHostUnhandledException were called

PR #260, opened 2019-09-18
Send Additional headers to support CheckAccess ClassAdmin Validation

rekhaswaminathan:
Change Analysis API requires 4 additional headers : x-ms-client-issuer, x-ms-client-puid, x-ms-client-alt-sec-id, x-ms-client-identity-provider. 

Passing the incoming request headers from DiagnosticRequestMiddleware to ChangeAnalysisClient which adds these headers.


src/Diagnostics.DataProviders/ChangeAnalysisClient.cs
```diff
@@ -310,6 +318,46 @@ public async Task<string> PrepareAndSendRequest(string requestUri, object postBo
                 requestMessage.Headers.Add(ClientPrincipalNameHeader, clientPrincipalNameHeader);
             }
 
+            string clientIssuer = string.Empty;
```
hforeste: One way you can improve this is have a string array of all the header names and iterate over this list grabbing the values and adding them to your requestMessage

PR #267, opened 2019-09-25
Catch empty tenant list scenario for ASE and respond appropriately

SharmaAjay19:
image

### DIFF
--- a/src/Diagnostics.DataProviders/KustoClient.cs
+++ b/src/Diagnostics.DataProviders/KustoClient.cs
@@ -57,10 +57,6 @@ namespace Diagnostics.DataProviders
 
         public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, string database, string requestId = null, string operationName = null)
         {
-            if(query!=null && query.Contains("Tenant in ()"))
-            {
-                throw new KustoTenantListEmptyException("KustoDataProvider", "Malformed Query: Query contains an empty tenant list.");
-            }
             return await ExecuteQueryAsync(query, cluster, database, DataProviderConstants.DefaultTimeoutInSeconds, requestId, operationName);
         }
 
@@ -99,6 +95,13 @@ namespace Diagnostics.DataProviders
 
                 if (!responseMsg.IsSuccessStatusCode)
                 {
+                    if ((int)responseMsg.StatusCode == 400)
+                    {
+                        if (query != null && query.Contains("Tenant in ()"))
+                        {
+                            throw new KustoTenantListEmptyException("KustoDataProvider", "Malformed Query: Query contains an empty tenant list.");
+                        }
+                    }
                     timeTakenStopWatch.Stop();
                     LogKustoQuery(query, cluster, operationName, timeTakenStopWatch, kustoClientId, new Exception($"Kusto call ended with a non success status code : {responseMsg.StatusCode.ToString()}"), dataSet, responseContent);
                     throw new Exception(responseContent);
--- a/src/Diagnostics.DataProviders/KustoClient.cs
+++ b/src/Diagnostics.DataProviders/KustoClient.cs
@@ -14,6 +14,7 @@ using System.Web;
 using Diagnostics.Logger;
 using Diagnostics.ModelsAndUtils.Models;
 using Newtonsoft.Json;
+using Diagnostics.DataProviders.Exceptions;
 
 namespace Diagnostics.DataProviders
 {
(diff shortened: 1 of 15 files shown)


src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -56,7 +57,10 @@ public KustoClient(KustoDataProviderConfiguration config, string requestId)
 
         public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, string database, string requestId = null, string operationName = null)
         {
-
+            if(query!=null && query.Contains("Tenant in ()"))
```
ShekharGupta1988: This logic would also catch cases where people have comments in the query.
Instead of this, can we let the query run and catch Kusto Bad Exception here and then check if the exception has Tenant in ()

ShekharGupta1988: That will make sure that the query indeed failed due to missing Tenant Info

SharmaAjay19: Good point. This will be a stronger catch.

SharmaAjay19: Although even in that case a query can be a bad query due to other reasons and still have a commented "Tenant in ()".

src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -95,6 +95,13 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
 
                 if (!responseMsg.IsSuccessStatusCode)
                 {
+                    if ((int)responseMsg.StatusCode == 400)
```
ShekharGupta1988: This section should be below LogKustoQuery Method .. 
You do want to stop the timeTakeStopWatch and log query before throwing the exception

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -306,6 +308,24 @@ protected async Task<IActionResult> ExecuteQuery<TPostBodyResource>(TResource re
...
+                            if (
+                                   (
+                                   // Resource is ASE
+                                    resource as IResource is HostingEnvironment env && (env.TenantIdList != null && env.TenantIdList.Count() == 0)
+                                   )
+                                   ||
+                                   (
+                                   // Resource is app in a Stamp of type ASE V1 or V2
+                                    (resource as IResource is App currentApp && ((currentApp.Stamp.HostingEnvironmentType == HostingEnvironmentType.V1) || (currentApp.Stamp.HostingEnvironmentType == HostingEnvironmentType.V2)))
+                                    && (currentApp.Stamp.TenantIdList != null && currentApp.Stamp.TenantIdList.Count() == 0)
```
ShekharGupta1988: nit: you can use .Any() instead of count() == 0

src/Diagnostics.RuntimeHost/Models/ASETenantListEmptyException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Diagnostics.RuntimeHost.Models.Exceptions
+{
+    public class ASETenantListEmptyException: Exception
```
ShekharGupta1988: internal instead of public should be fine here?

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
```diff
@@ -51,6 +59,10 @@ public async Task Invoke(HttpContext httpContext)
                 {
                     httpContext.Response.Clear();
                     httpContext.Response.StatusCode = statusCode;
+                    if (responseMessage != null)
+                    {
+                        httpContext.Response.WriteAsync(responseMessage);
```
ShekharGupta1988: no await ?

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -306,6 +308,24 @@ protected async Task<IActionResult> ExecuteQuery<TPostBodyResource>(TResource re
                     }
                     catch (Exception ex)
                     {
+                        if (ex.InnerException is KustoTenantListEmptyException)
+                        {
+                            if (
+                                   (
+                                   // Resource is ASE
+                                    resource as IResource is HostingEnvironment env && (env.TenantIdList != null && !env.TenantIdList.Any())
```
hforeste: No a big deal but it looks like your using this logic in two locations of this file, I recommend just creating a method that does this logic for you to avoid any risk of mismatched logic and etc.

PR #269, opened 2019-10-01
Reduce Bad Requests To Observer Missing Stamp Name

dagrooms52:
- Use GetAdminSites instead of GetSite for faster stamp name lookup
- Reduce 404 requests to Observer by requiring stamp name for GET RuntimeSiteSlotMap
  - Observer does not accept requests to this route without a stamp name
- Check for null or empty stamp name inputs before sending Observer requests to prevent 404 response
  - The route `/api/stamps/sites/{sitename}/{resource}` was commonly created, causing 404 responses

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
@@ -231,9 +231,7 @@ namespace Diagnostics.DataProviders
         public abstract Task<JArray> GetAdminSitesAsync(string siteName);
 
         public abstract Task<string> GetStorageVolumeForSiteAsync(string stampName, string siteName);
-
-        public abstract Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string siteName);
-
+        
         public abstract Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string stampName, string siteName);
 
         public abstract Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string stampName, string siteName, string slotName);
@@ -245,6 +243,11 @@ namespace Diagnostics.DataProviders
         public DataProviderMetadata GetMetadata()
         {
             return null;
+        }
+
+        public Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string siteName)
+        {
+            throw new NotImplementedException("Cannot retrieve slot map without a stamp name.");
         }
     }
 }
--- a/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
@@ -244,10 +244,5 @@ namespace Diagnostics.DataProviders
         {
             return null;
         }
-
-        public Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string siteName)
-        {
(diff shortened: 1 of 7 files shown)


src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -29,6 +29,20 @@ public override Task<JObject> GetAdminSitesBySiteNameAsync(string stampName, str
...
+        public override async Task<JArray> GetAdminSitesAsync(string siteName)
+        {
+            if (string.IsNullOrWhiteSpace(siteName))
+            {
+                throw new ArgumentNullException(nameof(siteName));    
+            }
+
+            var path = $"sites/{siteName}/adminsites";
+            
+            var response = await GetObserverResource(path);
```
michimune: Don't we need to check the response here?

src/Diagnostics.RuntimeHost/Controllers/ObserverController.cs
```diff
@@ -18,12 +18,27 @@ public class ObserverController : Controller
         [HttpGet(UriElements.ObserverGetSites)]
         public async Task<IActionResult> GetSiteDetails(string siteName)
         {
+            if (string.IsNullOrWhiteSpace(siteName))
```
hforeste: Ignore this file for changes as this code path is only executed when using LocalHostObserver which were not using in PublicAzure and we'll soon migrate from in national clouds.

src/Diagnostics.DataProviders/DataProviders/MockSupportObserverDataProvider.cs
```diff
@@ -20,21 +20,16 @@ public MockSupportObserverDataProvider(OperationDataCache cache, SupportObserver
         {
         }
 
-        public override Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string siteName)
```
hforeste: Lets find out how many detectors are using this signature and then convert them to using the `GetRuntimeSiteSlotMap(string stampName, string siteName)` variant

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -244,5 +244,10 @@ public DataProviderMetadata GetMetadata()
         {
             return null;
         }
+
+        public Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string siteName)
```
hforeste: Instead of taking this approach, I suggest we just make the changes to detectors to use the `GetRuntimeSiteSlotMap(string stampName, string siteName)` variant

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
dagrooms52: Yes, good catch

src/Diagnostics.DataProviders/DataProviders/MockSupportObserverDataProvider.cs
dagrooms52: I don't see any using this method - this is a precaution so that users don't run into it, but I could also remove it from the API

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
dagrooms52: I don't see any detectors using this - it would probably be safe to remove it from the API

src/Diagnostics.DataProviders/DataProviders/MockSupportObserverDataProvider.cs
hforeste: Yea I think we can just remove this then

PR #272, opened 2019-10-01
Cmaher/source watcher concurrency

craigmaher:
This aims to solve the issue of multiple instances accessing the same files.
If a conflict arises this should pick up the changes in the next iteration.

### DIFF
--- a/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/GitHubWatcher.cs
+++ b/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/GitHubWatcher.cs
@@ -20,6 +20,7 @@ using Microsoft.AspNetCore.Hosting;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Win32;
 using Newtonsoft.Json;
+using System.Diagnostics;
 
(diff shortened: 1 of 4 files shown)


src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/GitHubWatcher.cs
```diff
@@ -227,6 +230,8 @@ private async Task StartWatcherInternal(bool startup)
             }
             finally
             {
+                stopWatch.Stop();
+                Console.WriteLine("RunTime " + stopWatch.ElapsedMilliseconds);
```
ShekharGupta1988: Remove this console logging. use LogMessage() method instead to log this to kusto.

src/Diagnostics.Scripts/Models/EntityMetadata.cs
```diff
@@ -6,16 +6,18 @@ public sealed class EntityMetadata
 
         public string ScriptText;
         public string Metadata;
+        public string LastModifiedMarker;
```
ShekharGupta1988: lets call it Sha here in the EntityMetadata Model instead of LastModifiedMarker?

PR #275, opened 2019-10-02
Catching actual kusto exception from kusto response

SharmaAjay19:


src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -138,8 +138,20 @@ private DataTableResponseObjectCollection ProcessKustoResponse(string responseCo
...
             }
-
-            return JsonConvert.DeserializeObject<DataTableResponseObjectCollection>(responseContent);
+            try
+            {
+                return JsonConvert.DeserializeObject<DataTableResponseObjectCollection>(responseContent);
+            }
+            catch (Newtonsoft.Json.JsonSerializationException ex)
+            {
+                var responseObj = JsonConvert.DeserializeObject<DataTableExceptionResponseObject>(responseContent);
```
nmallick1: Perhaps put this inside a try catch block ? This line could also possibly throw an exception just in case the returned response happens to be of some other format.

src/Diagnostics.DataProviders/Exceptions/KustoResponseSchemaException.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+
+namespace Diagnostics.DataProviders.Exceptions
+{
+    public class KustoResponseSchemaException : Exception
```
ShekharGupta1988: The name can create confusions between our ExceptionTypes and Kusto Team ExceptionTypes in future. Instead of calling Kusto, can we use this Exception for any provider in future. How about 'MalformedResponseException' or something like that?

src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -138,8 +138,20 @@ private DataTableResponseObjectCollection ProcessKustoResponse(string responseCo
             {
                 throw new Exception($"Kusto query response exceeded the Kusto record count limit: {_queryLimitDocsLink}");
             }
-
-            return JsonConvert.DeserializeObject<DataTableResponseObjectCollection>(responseContent);
+            try
+            {
+                return JsonConvert.DeserializeObject<DataTableResponseObjectCollection>(responseContent);
+            }
+            catch (Newtonsoft.Json.JsonSerializationException ex)
```
dagrooms52: The above checks for "Kusto response exceeded record count limit" and "data size limit" are redundant with the try/catch - if this catch block is capable of recognizing those checks and returning the proper error, you can remove them.

src/Diagnostics.DataProviders/Exceptions/KustoResponseSchemaException.cs
```diff
@@ -0,0 +1,11 @@
...
+    {
+        public KustoResponseSchemaException(string sourceName, string message): base($"Exception occurred in {sourceName}. {message}")
```
dagrooms52: VS should let you generate all inherited constructors from the base class - I believe there are 3 standard public exception constructors, they should be available on custom exceptions

src/Diagnostics.DataProviders/KustoClient.cs
SharmaAjay19: We are not sure right now if the kusto response in the "response exceeded record count limit" follows the same exception structure. Will look further if this can be generalized.

PR #276, opened 2019-10-02
Make kusto client request id's unique to aid in investigation

nmallick1:
Append a new Guid to ClientRequestId to make the request id unique in Kusto. This helps in investigating Kusto issues by the Kusto team.

Intentionally creating a new request id only while sending the request to Kusto without impacting the logging so that we don't need to change any of our exiting dashboards / monitoring.

### DIFF
--- a/src/Diagnostics.DataProviders/KustoClient.cs
+++ b/src/Diagnostics.DataProviders/KustoClient.cs
@@ -76,7 +76,7 @@ namespace Diagnostics.DataProviders
                 csl = query,
                 properties = new
                 {
-                    ClientRequestId = kustoClientId,
+                    ClientRequestId = $"{kustoClientId}_{(new Guid()).ToString()}",
                     Options = new
                     {
                         servertimeout = new TimeSpan(0, 1, 0)


src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -76,7 +76,7 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
                 csl = query,
                 properties = new
                 {
-                    ClientRequestId = kustoClientId,
+                    ClientRequestId = $"{kustoClientId}_{(new Guid()).ToString()}",
```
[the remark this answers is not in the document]
nmallick1: duh !!! I should have thought about this :( Made the change. Thanks @ShekharGupta1988 :)

src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -64,11 +64,11 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
         {
             var timeTakenStopWatch = new Stopwatch();
             var authorizationToken = await KustoTokenService.Instance.GetAuthorizationTokenAsync();
-            var kustoClientId = $"Diagnostics.{operationName ?? "Query"};{_requestId}##{0}";
+            var kustoClientId = $"Diagnostics.{operationName ?? "Query"};{_requestId}##{0}_{(new Guid()).ToString()}";
```
dagrooms52: Why is the string `"##0"` appended to `_requestId`? 
I wonder if this is intentional or a leftover `{0}` from a previous `string.Format()` operation.

nmallick1: It most likely is, however I am a little reluctant to change it as we might be parsing this field somewhere and I don't want whatever (if anything ) is doing it.

PR #281, opened 2019-10-07
upgraded to .net core 2.2

craigmaher:


src/Diagnostics.DataProviders/LogDecoratorBase.cs
```diff
@@ -29,23 +29,25 @@ protected async Task<T> MakeDependencyCall<T>(Task<T> dataProviderTask, [CallerM
             try
             {
                 var tcs = new TaskCompletionSource<bool>();
-                cTokenRegistration = _dataSourceCancellationToken.Register(() => tcs.TrySetResult(true));
-                var completedTask = await Task.WhenAny(new Task[] { dataProviderTask, tcs.Task });
-
-                if (completedTask.Id == dataProviderTask.Id)
+                using (cTokenRegistration = _dataSourceCancellationToken.Register(() => tcs.TrySetResult(true)))
```
ShekharGupta1988: These changes intentional as part of upgrade to 2.2?

craigmaher: Yes  cTokenRegistration is no longer a null-able type so this follows best practice for disposing of objects instead.

PR #291, opened 2019-10-09
Set environment to Production

rekhaswaminathan:


infrastructure/ArmTemplates/antareskvtemplate.json
```diff
@@ -334,7 +334,7 @@
               },
               {
                 "Name": "ASPNETCORE_ENVIRONMENT",
-                "Value": "Staging"
+                "Value": "Production"
```
ShekharGupta1988: How will this work when we deploy to staging slot? wont the value be set to Production there?

rekhaswaminathan: https://github.com/Azure/Azure-AppServices-Diagnostics/blob/5b3edd6eceedcc0dbeb7530a7a18e942c44d7258/infrastructure/ArmTemplates/antareskvtemplate.json#L390

There is a different variable for staging slot

PR #293, opened 2019-10-10
WEBSITE_HOSTNAME fix

craigmaher:
as WEBSITE_HOSTNAME appears to be flaky I think we need to create an environment variable ourselves.

### DIFF
--- a/infrastructure/ArmTemplates/antareskvtemplate.json
+++ b/infrastructure/ArmTemplates/antareskvtemplate.json
@@ -325,7 +325,7 @@
             "logsDirectorySizeLimit": 50,
             "appSettings": [
               {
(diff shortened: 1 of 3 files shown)


infrastructure/ArmTemplates/antareskvtemplate.json
```diff
@@ -324,6 +324,10 @@
             "httpLoggingEnabled": true,
             "logsDirectorySizeLimit": 50,
             "appSettings": [
+              {
+                  "Name" : "WEBSITE_HOSTNAME2",
```
ShekharGupta1988: Instead of calling it website_hostname2, let's call it DIAG_HOST

PR #295, opened 2019-10-10
Auto heal rule

craigmaher:
Adding an auto heal rule to prod slots

### DIFF
--- a/infrastructure/ArmTemplates/antareskvtemplate.json
+++ b/infrastructure/ArmTemplates/antareskvtemplate.json
@@ -362,7 +362,12 @@
                 {
                   "status": 500,
                   "subStatus": 0,
-                  "win32Status": 0,
+                  "count": 100,
+                  "timeInterval": "00:10:00"
+                },
+                {
+                  "status": 503,
+                  "subStatus": 0,
                   "count": 100,
                   "timeInterval": "00:10:00"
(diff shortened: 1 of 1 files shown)


infrastructure/ArmTemplates/antareskvtemplate.json
```diff
@@ -352,7 +352,29 @@
...
+                      "subStatus": 0,
+                      "win32Status": 0,
+                      "count": 100,
+                      "timeInterval": "00:10:00"
+                  }
+                ],
+                "slowRequests": null
+              },
+              "actions": {
+                "actionType": "Recycle",
```
ShekharGupta1988: can we collect dumps and then recycle

ShekharGupta1988: Also is there a way to recycle for StatusCode >=500

craigmaher: there is no way to specify status code above 500 we can specify multiple unique status codes though. are there any others you can think of?

ShekharGupta1988: 503 would be good to add.
The substatus and win32 is optional I believe as per Autoheal UI, so we can prob ignore those

PR #298, opened 2019-10-14
Fix Guid creation in Kusto client request id generation

nmallick1:
Changed to using the Static NewGuid method to create an actual Guid instead of an empty Guid object.

### DIFF
--- a/src/Diagnostics.DataProviders/KustoClient.cs
+++ b/src/Diagnostics.DataProviders/KustoClient.cs
@@ -64,7 +64,7 @@ namespace Diagnostics.DataProviders
         {
             var timeTakenStopWatch = new Stopwatch();
             var authorizationToken = await KustoTokenService.Instance.GetAuthorizationTokenAsync();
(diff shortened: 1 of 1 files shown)


src/Diagnostics.DataProviders/KustoClient.cs
```diff
@@ -64,7 +64,7 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
         {
             var timeTakenStopWatch = new Stopwatch();
             var authorizationToken = await KustoTokenService.Instance.GetAuthorizationTokenAsync();
-            var kustoClientId = $"Diagnostics.{operationName ?? "Query"};{_requestId}##{0}_{(new Guid()).ToString()}";
+            var kustoClientId = $"Diagnostics.{operationName ?? "Query"};{_requestId}##{0}_{Guid.NewGuid().ToString()}";
```
ShekharGupta1988: ya :)

PR #301, opened 2019-10-15
compiler host revert

craigmaher:
revert compiler host to net core 2.0 to avoid further conflicts with diag role

### DIFF
--- a/src/Diagnostics.CompilerHost/Diagnostics.CompilerHost.csproj
+++ b/src/Diagnostics.CompilerHost/Diagnostics.CompilerHost.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


puneetg1983:
You think that the SEV 2 that happened today, (caused by the following exception) was caused due to this change?

`System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.IO.FileNotFoundException: Could not load file or assembly 'System.Data.Common, Version=4.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.`

craigmaher:
> eption: Exception has been thrown by the target of an invocation. ---> System.IO.FileNotFoundException: Could not load file or assembly 'System.Data.Common, Version=4.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.

yes I believe it was

puneetg1983:
So even in liberation, compilerhost will use 2.0?

puneetg1983:
> This is a temporary change until we move DaaS over to the liberation project then we can upgrade to net core 3.0.

DaaS is a site extension and it will not be moved to liberation.

craigmaher:
> > This is a temporary change until we move DaaS over to the liberation project then we can upgrade to net core 3.0.
> 
> DaaS is a site extension and it will not be moved to liberation.

I mean until we can deprecate the old Diag role and  all of the detector flows are using the liberation project

PR #302, opened 2019-10-16
Search models download and refresh not required

SharmaAjay19:


rekhaswaminathan:
@SharmaAjay19  This will not affect anything if this change only goes to runtime host liberation project first and not runtime host inside diagrole? Just to make sure.

SharmaAjay19:
> @SharmaAjay19 This will not affect anything if this change only goes to runtime host liberation project first and not runtime host inside diagrole? Just to make sure.

This change will never go to diagrole as diagrole still needs to do the job. Its the liberation that should not do this and hence removing only in master2.

rekhaswaminathan:
Got it, Thanks Ajay

PR #307, opened 2019-10-20
Start time cannot be after 15 minutes before current time

SharmaAjay19:


src/Diagnostics.RuntimeHost/Utilities/DateTimeHelper.cs
```diff
@@ -103,9 +103,9 @@ internal static bool PrepareStartEndTimeUtc(string startTime, string endTime, ou
                 errorMessage = "Invalid Start Time and End Time. End Time cannot be earlier than Start Time.";
                 return false;
             }
-            else if (startTimeUtc > currentUtcTime)
+            else if (startTimeUtc > currentUtcTime.AddMinutes(-15))
             {
-                errorMessage = "Invalid Start Time. Start Time cannot be a future date.";
+                errorMessage = "Invalid Start Time. Start Time cannot be after 15 minutes before current time.";
```
hforeste: Can we say instead `Invalid Start Time. Start Time needs to at least be 15 mins ago.`

src/Diagnostics.RuntimeHost/Utilities/DateTimeHelper.cs
```diff
@@ -103,9 +103,9 @@ internal static bool PrepareStartEndTimeUtc(string startTime, string endTime, ou
                 errorMessage = "Invalid Start Time and End Time. End Time cannot be earlier than Start Time.";
                 return false;
             }
-            else if (startTimeUtc > currentUtcTime)
+            else if (startTimeUtc > currentUtcTime.AddMinutes(-15))
```
nmallick1: Should we have a similar check for end time too ? UI forces the endtime to be 15 minutes less than the current time.

src/Diagnostics.RuntimeHost/Utilities/DateTimeHelper.cs
```diff
@@ -66,10 +66,11 @@ internal static bool PrepareStartEndTimeUtc(string startTime, string endTime, ou
...
             errorMessage = string.Empty;
+            int kustoDelayInMinutes = -HostConstants.KustoDelayInMinutes;
 
             if (string.IsNullOrWhiteSpace(startTime) && string.IsNullOrWhiteSpace(endTime))
             {
-                endTimeUtc = currentUtcTime;
+                endTimeUtc = currentUtcTime.AddMinutes(kustoDelayInMinutes);
                 startTimeUtc = endTimeUtc.AddDays(-1);
             }
             else if (string.IsNullOrWhiteSpace(startTime))
```
nmallick1: It is possible in this condition that the passed in end time is > (now - 15 minutes)

src/Diagnostics.RuntimeHost/Utilities/DateTimeHelper.cs
```diff
@@ -103,9 +104,14 @@ internal static bool PrepareStartEndTimeUtc(string startTime, string endTime, ou
                 errorMessage = "Invalid Start Time and End Time. End Time cannot be earlier than Start Time.";
                 return false;
             }
-            else if (startTimeUtc > currentUtcTime)
+            if (endTimeUtc > currentUtcTime.AddMinutes(kustoDelayInMinutes))
             {
-                errorMessage = "Invalid Start Time. Start Time cannot be a future date.";
+                errorMessage = $"Invalid Start Time. End Time should be at least {HostConstants.KustoDelayInMinutes} minutes from now";
```
nmallick1: Possibly change the error message to 
"Invalid Start Time. End Time should be less by at least {HostConstants.KustoDelayInMinutes} minutes from now"

PR #309, opened 2019-10-22
Add gist template for partner teams on Applens

nmallick1:
Add gist template for all resource types that have been onboarded to Applens


hforeste:
Its exciting to see all the partners that has on boarded

PR #310, opened 2019-10-23
Refactor GeoMasterDataProvider

michimune:
This PR refactors GeoMasterDataProvider to remove frequent HttpClient constructions/destructions. It should reduce the number of socket open/close operations and lower the chance of SNAT port exhaustions. This PR also enables to log the full stack traces for first level exceptions.

There was an issue where lots of NullReferenceExceptions happening in the first 20 seconds when the project starts up. This was due to a race condition of the TokenServiceBase class initialization. This PR fixes it by waiting for a property to be populated.

### DIFF
--- a/src/Diagnostics.DataProviders/ArmClient.cs
+++ b/src/Diagnostics.DataProviders/ArmClient.cs
@@ -7,22 +7,24 @@ namespace Diagnostics.DataProviders
     internal class ArmClient : IGeoMasterClient
     {
         private const string CsmEndpointUrl = "https://management.azure.com/";
-        public HttpClient Client { get; }
-        public Uri BaseUri { get; }
+        private static readonly HttpClient _client = new HttpClient();
+        public HttpClient Client => _client;
+        public Uri BaseUri => _client.BaseAddress;
+        public string AuthenticationToken { get; }
+
+        static ArmClient()
+        {
+            _client.BaseAddress = new Uri(CsmEndpointUrl);
+            _client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+        }
 
         public ArmClient(GeoMasterDataProviderConfiguration configuration)
         {
-            var handler = new HttpClientHandler();
-            BaseUri = new Uri(CsmEndpointUrl);
-
-            Client = new HttpClient(handler)
+            if (configuration.Token == null)
             {
-                BaseAddress = BaseUri
-            };
-
-            Client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
-            Client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", configuration.Token);
-            Client.BaseAddress = BaseUri;
+                throw new ArgumentNullException("NULL TOKEN!!");
+            }
+            this.AuthenticationToken = configuration.Token;
         }
     }
 }
--- a/src/Diagnostics.DataProviders/ArmClient.cs
+++ b/src/Diagnostics.DataProviders/ArmClient.cs
@@ -20,9 +20,9 @@ namespace Diagnostics.DataProviders
 
         public ArmClient(GeoMasterDataProviderConfiguration configuration)
         {
-            if (configuration.Token == null)
+            if (string.IsNullOrWhiteSpace(configuration.Token))
             {
-                throw new ArgumentException("configuration.Token is null");
+                throw new ArgumentException("configuration.Token is null or empty");
             }
             this.AuthenticationToken = configuration.Token;
         }
--- a/src/Diagnostics.DataProviders/ArmClient.cs
+++ b/src/Diagnostics.DataProviders/ArmClient.cs
@@ -22,7 +22,7 @@ namespace Diagnostics.DataProviders
         {
             if (configuration.Token == null)
             {
-                throw new ArgumentNullException("NULL TOKEN!!");
+                throw new ArgumentException("configuration.Token is null");
             }
             this.AuthenticationToken = configuration.Token;
         }
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -584,11 +584,12 @@ namespace Diagnostics.DataProviders
         private async Task<R> PerformHttpRequest<R>(HttpMethod method, string path, string queryString, string content, string apiVersion, CancellationToken cancellationToken)
         {
             var query = SitePathUtility.CsmAnnotateQueryString(queryString, apiVersion);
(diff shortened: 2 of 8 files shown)


src/Diagnostics.DataProviders/ArmClient.cs
```diff
@@ -7,22 +7,24 @@ namespace Diagnostics.DataProviders
...
+        {
+            if (configuration.Token == null)
             {
-                BaseAddress = BaseUri
-            };
-
-            Client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
-            Client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", configuration.Token);
-            Client.BaseAddress = BaseUri;
+                throw new ArgumentNullException("NULL TOKEN!!");
```
rekhaswaminathan: Can we change this to `ArgumentNullException(nameof(configuration.Token))`?

src/Diagnostics.DataProviders/TokenService/TokenServiceBase.cs
```diff
@@ -99,6 +99,10 @@ public async Task<string> GetAuthorizationTokenAsync()
         {
             if (!tokenAcquiredAtleastOnce)
             {
+                while (acquireTokenTask == null)
+                {
+                    await Task.Delay(1000);
```
rekhaswaminathan: So we want to wait 1s before acquiring token the first time?

rekhaswaminathan:
@michimune  It looks like Travis-ci build is failing because the `DataProviderTest.cs` is invoking `GeoMasterDataProvider` without the token.

src/Diagnostics.DataProviders/ArmClient.cs
michimune: Thanks. I tried it and it showed "Value cannot be null.Parameter name: Token" instead of "configuration.Token." So I will put a literal string "configuration.Token" instead.

src/Diagnostics.DataProviders/TokenService/TokenServiceBase.cs
michimune: Yes, it's much faster than 20 seconds. But obviously there is a better way to implement here by using a proper synchronization object. Let me improve so it makes more sense.

michimune:
I will take a look at the test failure.

src/Diagnostics.DataProviders/TokenService/TokenServiceBase.cs
michimune: Replaced Task.Delay() with ManualResetEvent.WaitOne()

michimune:
The tests have passed after fixing the test code.

src/Diagnostics.DataProviders/ArmClient.cs
```diff
@@ -7,22 +7,24 @@ namespace Diagnostics.DataProviders
...
-            var handler = new HttpClientHandler();
-            BaseUri = new Uri(CsmEndpointUrl);
+            _client.BaseAddress = new Uri(CsmEndpointUrl);
+            _client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+        }
 
-            Client = new HttpClient(handler)
+        public ArmClient(GeoMasterDataProviderConfiguration configuration)
+        {
+            if (configuration.Token == null)
```
cindylovescoding: Can we use this check: string.IsNullOrWhiteSpace(configuration.Token) to cover the empty string or whitespace as well?

src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -22,6 +23,14 @@ public Startup(IConfiguration configuration, IHostingEnvironment environment)
         {
             Configuration = configuration;
             Environment = environment;
+
+            if (!Environment.IsProduction())
```
cindylovescoding: Thank you for adding this Michimune!

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -200,12 +201,11 @@ public string RemovePIIFromSettings(string content)
         /// }
         /// </code>
         /// </example>
-        public async Task<VnetValidationRespone> VerifyHostingEnvironmentVnet(string subscriptionId, string vnetResourceGroup, string vnetName, string vnetSubnetName, CancellationToken cancellationToken = default(CancellationToken))
```
cindylovescoding: Why do we want to remove all these?

src/Diagnostics.DataProviders/ArmClient.cs
michimune: Yes, that's apparently better. I will update.

src/Diagnostics.RuntimeHost/Startup.cs
michimune: Your welcome!

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
michimune: It's all about avoiding potential deadlocking.

https://stackoverflow.com/questions/21033150/any-difference-between-await-task-run-return-and-return-task-run

With those changes, you will not be able to set a breakpoint at the await or return statements, but you could still set it at their callers.

src/Diagnostics.DataProviders/ArmClient.cs
michimune: Updated

src/Diagnostics.DataProviders/TokenService/TokenServiceBase.cs
```diff
@@ -97,12 +98,7 @@ public async Task StartTokenRefresh()
         /// </summary>
         public async Task<string> GetAuthorizationTokenAsync()
         {
-            if (!tokenAcquiredAtleastOnce)
-            {
-                var authResult = await acquireTokenTask;
-                return GetAuthTokenFromAuthenticationResult(authResult);
-            }
-
+            tokenAcquiredAtleastOnce.WaitOne();
```
puneetg1983: It seems here we are mixing sync code with Async code and this has a potential to deadlock

michimune: Fixed

src/Diagnostics.DataProviders/GeoMasterCertClient.cs
```diff
@@ -11,11 +11,30 @@ internal class GeoMasterCertClient : IGeoMasterClient
...
+            }
+
+            var geoEndpoint = new UriBuilder(geoMasterHostName)
+            {
+                Scheme = "https",
+                Port = GeoMasterCsmApiPort
+            };
+
+            BaseUri = geoEndpoint.Uri;
+            _httpClient.BaseAddress = BaseUri;
```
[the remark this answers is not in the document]
michimune: Fixed

PR #314, opened 2019-10-29
Add GetLogFolderZip API to GeoMasterDataProvider

michimune:
This PR adds a new method named GetLogFolderZip to the GeoMasterDataProvider. This new method allows detectors to download the entire contents under app's /home/LogFiles as a zipped binary image. Detectors can unzip and read log files from it. Typical use cases include php log analysis, dotnet log analysis, getting logs older than 10 days, etc.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -7,6 +7,7 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
+using Diagnostics.Logger;
 using Diagnostics.ModelsAndUtils.Models;
 using Newtonsoft.Json;
 
@@ -19,6 +20,16 @@ namespace Diagnostics.DataProviders
         private readonly IGeoMasterClient _geoMasterClient;
         private GeoMasterDataProviderConfiguration _configuration;
         private string _geoMasterHostName;
+        private Lazy<HttpClient> _scmClient = new Lazy<HttpClient>(() =>
+        {
+            var httpClient = new HttpClient();
+            httpClient.DefaultRequestHeaders.Accept.Clear();
+            httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(HeaderConstants.OctetStreamContentType));
+            httpClient.DefaultRequestHeaders.Add(HeaderConstants.UserAgentHeaderName, "appservice-diagnostics");
+            httpClient.Timeout = TimeSpan.FromSeconds(30);
+
+            return httpClient;
+        });
 
         private string[] AllowedlistAppSettingsStartingWith = new string[] { "WEBSITE_", "WEBSITES_", "FUNCTION_", "FUNCTIONS_", "AzureWebJobsSecretStorageType"};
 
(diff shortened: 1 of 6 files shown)


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -481,6 +492,72 @@ public Task<string> GetLinuxContainerLogs(string subscriptionId, string resource
...
+        /// the sensitive settings like connectionstrings, tokens, secrets, keys, content shares etc.
+        /// </summary>
+        /// <param name="subscriptionId">Subscription Id for the resource</param>
+        /// <param name="resourceGroupName">The resource group that the resource is part of </param>
+        /// <param name="name">Name of the resource</param>
+        /// <param name="slotName">slot name (if querying for a slot, defaults to production slot)</param>
+        /// <returns>A dictionary of AppSetting Keys and values</returns>
+        /// <example>
+        /// <code>
+        /// This sample shows how to call the <see cref="GetAppSettings"/> method in a detector
```
rekhaswaminathan: Can we update the documentation here to reflect `GetLogFolderZip`?

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -19,6 +20,16 @@ public class GeoMasterDataProvider : DiagnosticDataProvider, IDiagnosticDataProv
         private readonly IGeoMasterClient _geoMasterClient;
         private GeoMasterDataProviderConfiguration _configuration;
         private string _geoMasterHostName;
+        private Lazy<HttpClient> _scmClient = new Lazy<HttpClient>(() =>
+        {
+            var httpClient = new HttpClient();
+            httpClient.DefaultRequestHeaders.Accept.Clear();
+            httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue(HeaderConstants.OctetStreamContentType));
+            httpClient.DefaultRequestHeaders.Add(HeaderConstants.UserAgentHeaderName, "appservice-diagnostics");
```
nmallick1: Do we have a constant for useragent that we use elsewhere (e.g.. Kusto data provider)? If not, may be we should create one.

michimune: Yes, it's used at multiple places. I will create one and use it.

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -481,6 +492,72 @@ public Task<string> GetLinuxContainerLogs(string subscriptionId, string resource
...
+        /// <code>
+        /// This sample shows how to call the <see cref="GetAppSettings"/> method in a detector
+        /// public async static Task<![CDATA[<Response>]]> Run(DataProviders dp, OperationContext<![CDATA[<App>]]> cxt, Response res)
+        /// {
+        ///     var subId = cxt.Resource.SubscriptionId;
+        ///     var rg = cxt.Resource.ResourceGroup;
+        ///     var name = cxt.Resource.Name;
+        ///     var slot = cxt.Resource.Slot;
+        ///
+        ///     byte[] zip = await dp.GeoMaster.GetLogFilesZip(subId, rg, name, slot);
```
nmallick1: Possibly put this in try catch since if detector authors copy paste the code from documentation, it won't impact our SLA.

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
michimune: Updated

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
michimune: Updated

michimune:
@puneetg1983 For your comments, the HTTP client has 30 seconds timeout. I am going to update the sample code to have a try-catch as Nazeef suggested which will help prevent from SLA dropping.

For ILB ASE, it indeed won't work. Alternative way is to call the SCM from the geomaster itself. I did some research on it but I could not find an easy way but adding a new one. So we need a caveat until we can find a solution. We can add a caveat insight to detectors which use this API.

I was thinking to add such an API (Get only relevant logs), but it's tricky because finding relevant logs itself isn't free and requires some CPU execution on the Kudu site. There is no single answer solving this problem and there is always a trade-off.

In my opinion, I would like to move forward rather than keep thinking about the best design. We can replace internal logic with something better at any time.

ShekharGupta1988:
Closing this old PR. Please reopen if needed.

PR #315, opened 2019-10-29
added batch sample detector

mksuni:


mksuni:
@nmallick1  I added the gist template as well for batch. Please review

PR #318, opened 2019-11-02
Geomaster and AppLens both sending slightly different hostnames in th…

hforeste:
…e requets headers. Uri.TryCreate is failing to create URI due to hostname from geomaster missing the HTTP(S) scheme.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -656,7 +656,7 @@ namespace Diagnostics.DataProviders
         {
             string geoMasterName = null;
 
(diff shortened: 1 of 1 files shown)


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -656,6 +656,11 @@ private string ParseGeoMasterName(string geomasterHostName)
         {
             string geoMasterName = null;
 
+            if (!string.IsNullOrWhiteSpace(geomasterHostName) && !geomasterHostName.StartsWith(Uri.UriSchemeHttp, StringComparison.CurrentCultureIgnoreCase))
```
SharmaAjay19: Shouldn't we use UriSchemeHttps here? Below we are prepending Https but here we are checking using Http.

hforeste: For our particular purposes it doesn't matter if its HTTPs or HTTP.

PR #319, opened 2019-11-04
Prevent Non-Existent Route Calls To Observer

dagrooms52:
- Check for null strings that directly cause non-existent routes to be created and called against Observer
- Users are currently seeing 404s on Observer, so the requests already fail
- Users will now see a proper error message in RuntimeHost, leading them to fix detector code

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
@@ -19,22 +19,10 @@ namespace Diagnostics.DataProviders
         {
         }
 
-        public override Task<JObject> GetAdminSitesByHostNameAsync(string stampName, string[] hostNames)
-        {
-            throw new NotImplementedException();
-        }
-
-        public override Task<JObject> GetAdminSitesBySiteNameAsync(string stampName, string siteName)
-        {
-            throw new NotImplementedException();
-        }
-
         public override async Task<JArray> GetAdminSitesAsync(string siteName)
         {
             if (string.IsNullOrWhiteSpace(siteName))
-            {
-                throw new ArgumentNullException(nameof(siteName));    
-            }
+                throw new ArgumentNullException(nameof(siteName));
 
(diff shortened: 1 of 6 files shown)


src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -115,64 +98,104 @@ public override async Task<dynamic> GetCertificatesInResourceGroupAsync(string s
...
         {
+            if (string.IsNullOrWhiteSpace(siteName))
+                throw new ArgumentNullException(nameof(siteName));
+
             return await GetSiteInternal(stampName, siteName, null);
         }
 
         public override async Task<dynamic> GetSite(string stampName, string siteName, string slotName)
         {
+            if (string.IsNullOrWhiteSpace(siteName))
```
hforeste: Since `GetSiteInternal` is already checking if sitename is empty, do we need to do this again here?

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -299,12 +330,40 @@ public override HttpClient GetObserverClient()
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public override Task<string> GetStorageVolumeForSiteAsync(string stampName, string siteName)
+        {
+            throw new NotImplementedException();
+        }
+
+        public override Task<IEnumerable<object>> GetAppServiceEnvironmentDeploymentsAsync(string hostingEnvironmentName)
```
hforeste: Let's just get rid of these not implemented methods. I haven't heard any one needing them.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
dagrooms52: Trying to fail as high in the call stack as possible - it makes the output clearer from a user perspective. 
So this is for a better debugging experience, the one in the internal call is a catch-all.

PR #324, opened 2019-11-06
Prevent RuntimeHost From Creating Non-Existent Observer Routes

dagrooms52:
- Had to close previous PR, Travis CI was hanging

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
@@ -19,22 +19,10 @@ namespace Diagnostics.DataProviders
         {
         }
 
-        public override Task<JObject> GetAdminSitesByHostNameAsync(string stampName, string[] hostNames)
-        {
-            throw new NotImplementedException();
-        }
-
-        public override Task<JObject> GetAdminSitesBySiteNameAsync(string stampName, string siteName)
-        {
-            throw new NotImplementedException();
-        }
-
         public override async Task<JArray> GetAdminSitesAsync(string siteName)
         {
             if (string.IsNullOrWhiteSpace(siteName))
-            {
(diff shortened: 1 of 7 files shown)


src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -72,20 +55,20 @@ public override async Task<dynamic> GetCertificatesInResourceGroupAsync(string s
         public override async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string stampName, string siteName, string slotName = null)
         {
             if (string.IsNullOrWhiteSpace(stampName))
```
hforeste: You don't really need these conditional checks since `GetRuntimeSiteSlotMapInternal` is already doing it for you.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -115,64 +98,104 @@ public override async Task<dynamic> GetCertificatesInResourceGroupAsync(string s
...
+                throw new ArgumentNullException(nameof(subscriptionId));
+            if (string.IsNullOrWhiteSpace(webSpaceName))
+                throw new ArgumentNullException(nameof(webSpaceName));
+
             return await Get($"subscriptionid/{subscriptionId}/webspacename/{webSpaceName}/resourcegroupname");
         }
 
         public override async Task<dynamic> GetSite(string siteName)
         {
+            if (string.IsNullOrWhiteSpace(siteName))
```
hforeste: Same here, `GetSiteInternal` is already doing the conditional check. Keeping this logic in one place is easier I think.

hforeste:
https://travis-ci.org/Azure/Azure-AppServices-Diagnostics/builds/608290882?utm_medium=notification&utm_source=github_status

PR #326, opened 2019-11-08
Adding a test method for querying, filter detectors in training and gitignores

SharmaAjay19:


src/Diagnostics.AIProjects/SearchAPI/SearchModule/views.py
```diff
@@ -108,6 +115,28 @@ def queryDetectorsMethod():
...
+    productid = getProductId(data)
+    if not productid:
+        return ('Resource data not available', 404)
+    productid = productid[0]
+    try:
+        loadModel(productid)
+    except Exception as e:
+        loggerInstance.logHandledException(requestId, e)
+        loggerInstance.logToFile(requestId, e)
+        return (json.dumps({"query": txts, "results": []}), 200)
```
ShekharGupta1988: you do want to return 200 in case of exception?

ShekharGupta1988: Do you want to differentiate at caller side that LoadModel failed vs genuine no results?

SharmaAjay19: Yeah we should return a 404 with exception message.

PR #342, opened 2019-11-20
Added log traces to see which GeoMaster URIs are failing

rekhaswaminathan:


hforeste:
Thank you for this change, it is much needed

PR #344, opened 2019-11-20
Word mover distance in search and renovating search architecture

SharmaAjay19:
Key changes:
1. Use word2vec technique in finding similarity in search algorithm.
2. Squeeze the word2vec model size by 99% to reduce trained model size.
2. Renovate search architecture to segregate algorithm from the app. It will allow plugging in any new algorithm into existing pipeline with just the algorithm name.
3. Making the app more developer friendly by adding the choice to use a pre-trained model by disabling model sync from storage account.
4. Refactor the code around modules for reusability in future testing pipeline.


src/Diagnostics.AIProjects/SearchAPI/SearchModule/views.py
```diff
@@ -56,33 +60,25 @@ def logger(*args, **kwargs):
...
-            time.sleep(2)
-        else:
-            break
+    if app.config['MODEL_SYNC_ENABLED']:
+        productIds = getAllProductIds(resourceConfig)
+        sah = StorageAccountHelper(loggerInstance)
+        loggerInstance.logInsights("Starting model sync for {0}".format(','.join(productIds)))
+        thread = threading.Thread(target=sah.watchModels, args=(productIds,))
+        thread.start()
+        while True:
```
dagrooms52: This should include a fail-safe, say if the loop has gone through 1000 iterations break out with an exception. Just to prevent infinite loops due to logic errors.

PR #345, opened 2019-11-21
Load cert default way with App Service

rekhaswaminathan:


src/Diagnostics.DataProviders/KeyVaultCertLoader/KeyVaultCertLoaderBase.cs
```diff
@@ -1,32 +1,42 @@
...
     {
-        protected abstract string KeyVault { get; set; }
-
-        protected abstract string CertificateName { get; set; }
-
+        protected abstract string Thumbprint { get; set; }
         public X509Certificate2 Cert { get; private set; }
 
-        public async void LoadCertFromKeyVault()
+        public async void LoadCertFromAppService()
```
dagrooms52: Using `async void` means callers can't `await` this to get exception details. There is some code outside of `try/catch` in the method body, so this has a possibility of losing an exception or crashing the service.

rekhaswaminathan: @dagrooms52 `async` was previously there when we were loading from Key Vault. I missed removing that when I made the change to load from App Service. Should be updated now

src/Diagnostics.DataProviders/KeyVaultCertLoader/CertLoaderBase.cs
```diff
@@ -0,0 +1,42 @@
...
+                // Get the first cert with the thumbprint
+                if (certCollection.Count > 0)
+                {
+                    Cert = certCollection[0];
+                    DiagnosticsETWProvider.Instance.LogRuntimeHostMessage($"Successfully loaded Cert with thumbprint {Thumbprint}");
+                }
+            }
+            catch (Exception ex)
+            {
+                DiagnosticsETWProvider.Instance.LogRuntimeHostMessage($"Error: {ex.Message} occurred while trying to load cert {Thumbprint} ");
```
ShekharGupta1988: Lets include the full stack trace as well in this message.

rekhaswaminathan: Sure

PR #353, opened 2019-12-05
Clearer Logic For Pulling Stamp Name From AdminSites

dagrooms52:
- Added clear exception messages if stamp name is not found on adminsites
- Previous logic allowed "StampName" to be present but still returned empty "InternalStampName"; logic now prefers InternalStampName before falling back to StampName

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
@@ -13,8 +13,6 @@ namespace Diagnostics.DataProviders
 {
     public class SupportObserverDataProvider : SupportObserverDataProviderBase
     {
-        private object _lockObject = new object();
-
         public SupportObserverDataProvider(OperationDataCache cache, SupportObserverDataProviderConfiguration configuration, DataProviderContext dataProviderContext) : base(cache, configuration, dataProviderContext)
         {
         }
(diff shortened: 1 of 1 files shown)


ShekharGupta1988:
Hey @dagrooms52 , Can you also FI this fix to master2 and trigger a deployment for Liberation endpoints.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -198,17 +196,29 @@ private async Task<dynamic> GetSiteInternal(string stampName, string siteName, s
...
+
+            var internalStampObjects = objects.Where(x => x.ContainsKey("InternalStampName") && !string.IsNullOrWhiteSpace(x["InternalStampName"].ToString()));
+            if (internalStampObjects.Any())
+                return internalStampObjects.First()["InternalStampName"].ToString();
+
+            var stampNameObjects = objects.Where(x => x.ContainsKey("StampName") && !string.IsNullOrWhiteSpace(x["StampName"].ToString()));
+            if (stampNameObjects.Any())
+                return stampNameObjects.First()["StampName"].ToString();
+
+            throw new Exception($"Admin Sites response did not contain stamp name for site {siteName}");
```
hforeste: In what conditions do we run into this?

dagrooms52: Lately we've been seeing an issue in which the call to `/adminsites` succeeds, but this method was unable to retrieve the stamp name from the response body.

PR #356, opened 2019-12-06
Add support of regional accounts to MDM data provider

enihcam:
0. Add 'Location' to ArmResource and get location from request header;
1. Add support of regional accounts to MDM data provider;
2. Add MDM data providers for Azure Spring Cloud scenarios.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviderConfigurations/IMdmDataProviderConfiguration.cs
+++ b/src/Diagnostics.DataProviders/DataProviderConfigurations/IMdmDataProviderConfiguration.cs
@@ -24,10 +24,5 @@ namespace Diagnostics.DataProviders.DataProviderConfigurations
         /// Gets or sets monitoring account.
         /// </summary>
         string MonitoringAccount { get; set; }
-
-        /// <summary>
-        /// Gets or sets if monitoring account is regional.
-        /// </summary>
-        bool IsRegional { get; set; }
     }
 }
--- a/src/Diagnostics.DataProviders/DataProviderConfigurations/IMdmDataProviderConfiguration.cs
+++ b/src/Diagnostics.DataProviders/DataProviderConfigurations/IMdmDataProviderConfiguration.cs
@@ -24,5 +24,10 @@ namespace Diagnostics.DataProviders.DataProviderConfigurations
         /// Gets or sets monitoring account.
         /// </summary>
         string MonitoringAccount { get; set; }
+
+        /// <summary>
+        /// Gets or sets if monitoring account is regional.
+        /// </summary>
+        bool IsRegional { get; set; }
     }
 }
--- a/src/Diagnostics.DataProviders/DataProviders.cs
+++ b/src/Diagnostics.DataProviders/DataProviders.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using Diagnostics.DataProviders.DataProviderConfigurations;
 using Diagnostics.DataProviders.Interfaces;
 
 
@@ -15,6 +16,7 @@ namespace Diagnostics.DataProviders
         public IChangeAnalysisDataProvider ChangeAnalysis;
         public IAscDataProvider Asc;
         public Func<MdmDataSource, IMdmDataProvider> Mdm;
+        public Func<IMdmDataProviderConfiguration, IMdmDataProvider> MdmGeneric;
 
         public DataProviders(DataProviderContext context)
         {
@@ -36,6 +38,7 @@ namespace Diagnostics.DataProviders
                         throw new NotSupportedException($"{ds} is not supported.");
                 }
             };
+            MdmGeneric = (IMdmDataProviderConfiguration config) => new MdmLogDecorator(context, new MdmDataProvider(_cache, config, context.RequestId, context.receivedHeaders));
         }
     }
 }
--- a/src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
@@ -4,7 +4,9 @@ using System.Data;
 using System.Threading.Tasks;
 using Diagnostics.DataProviders.DataProviderConfigurations;
 using Diagnostics.DataProviders.Interfaces;
+using Diagnostics.Logger;
 using Diagnostics.ModelsAndUtils.Models;
+using Microsoft.AspNetCore.Http;
 
 namespace Diagnostics.DataProviders
 {
@@ -22,10 +24,17 @@ namespace Diagnostics.DataProviders
         /// <param name="cache">Operation cache.</param>
         /// <param name="configuration">Data provider configuration.</param>
         /// <param name="requestId">Request id.</param>
-        public MdmDataProvider(OperationDataCache cache, IMdmDataProviderConfiguration configuration, string requestId)
+        public MdmDataProvider(OperationDataCache cache, IMdmDataProviderConfiguration configuration, string requestId, IHeaderDictionary headers = null)
             : base(cache)
         {
             _configuration = configuration;
+
+            // If the MDM account is regional, get the location from request headers and update the mdm account name with location suffix.
+            if (configuration.IsRegional && headers != null && headers.TryGetValue(HeaderConstants.LocationHeader, out var locations) && locations.Count > 0)
+            {
+                _configuration.MonitoringAccount += locations[0];
+            }
+
             _mdmClient = MdmClientFactory.GetMdmClient(configuration, requestId);
         }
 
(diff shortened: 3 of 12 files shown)


enihcam:
@Sneezry @xscript @cindylovescoding @ShekharGupta1988

src/Diagnostics.DataProviders/DataProviders.cs
```diff
@@ -15,6 +16,7 @@ public class DataProviders
         public IChangeAnalysisDataProvider ChangeAnalysis;
         public IAscDataProvider Asc;
         public Func<MdmDataSource, IMdmDataProvider> Mdm;
+        public Func<IMdmDataProviderConfiguration, IMdmDataProvider> Mdm2;
```
cindylovescoding: Maybe we can come up with a better name for this to avoid confusion

tests/DataProviderTests/GenericMdmDataProviderTest.cs
```diff
@@ -0,0 +1,122 @@
+﻿using System.Collections.Generic;
```
cindylovescoding: Thanks for adding the tests!

src/Diagnostics.RuntimeHost/Controllers/ArmResource/ArmResourceController.cs
```diff
@@ -21,39 +22,39 @@ public ArmResourceController(IServiceProvider services, IRuntimeContext<ArmResou
         }
 
         [HttpPost(UriElements.Query)]
-        public async Task<IActionResult> ExecuteQuery(string subscriptionId, string resourceGroupName, string provider, string resourceTypeName, string resourceName, [FromBody]CompilationPostBody<dynamic> jsonBody, string startTime = null, string endTime = null, string timeGrain = null, [FromQuery][ModelBinder(typeof(FormModelBinder))] Form Form = null)
+        public Task<IActionResult> ExecuteQuery(string subscriptionId, string resourceGroupName, string provider, string resourceTypeName, string resourceName, [FromBody]CompilationPostBody<dynamic> jsonBody, string startTime = null, string endTime = null, string timeGrain = null, [FromQuery][ModelBinder(typeof(FormModelBinder))] Form Form = null)
         {
-            return await base.ExecuteQuery(new ArmResource(subscriptionId, resourceGroupName, provider, resourceTypeName, resourceName), jsonBody, startTime, endTime, timeGrain, Form: Form);
+            return base.ExecuteQuery(new ArmResource(subscriptionId, resourceGroupName, provider, resourceTypeName, resourceName, GetLocation()), jsonBody, startTime, endTime, timeGrain, Form: Form);
```
cindylovescoding: What will happen if GetLocation() returns null here?

src/Diagnostics.DataProviders/DataProviders.cs
enihcam: @nmallick1 what do you suggest?

src/Diagnostics.RuntimeHost/Controllers/ArmResource/ArmResourceController.cs
enihcam: location would just be null. detectors need to expect this case happens.

src/Diagnostics.DataProviders/DataProviders.cs
enihcam: Renamed to MdmGeneric

src/Diagnostics.DataProviders/DataProviderConfigurations/IMdmDataProviderConfiguration.cs
```diff
@@ -24,5 +24,10 @@ public interface IMdmDataProviderConfiguration
         /// Gets or sets monitoring account.
         /// </summary>
         string MonitoringAccount { get; set; }
+
+        /// <summary>
+        /// Gets or sets if monitoring account is regional.
+        /// </summary>
+        bool IsRegional { get; set; }
```
nmallick1: @enihcam , The location header is not currently present in Applens (when CSS uses the tool) so they will not be able to pass the location header. Also, an addition of this property will mean that we update our config too and our deployment pipeline.

I looked at the rest of code and seems like the location is being used to identify the correct MDM endpoint depending on the location. Can this be handled in the GIST that we discussed about instead of letting the platform construct the MDM endpoint. Also, if the MDM endpoint is constructed inside the gist, each partner can create their own endpoints based on their logic instead of depending on the platforms logic.

I'd suggest removing this property from the backend and let the gist take care of creating the correct MDM endpoint.

src/Diagnostics.DataProviders/DataProviders.cs
```diff
@@ -36,6 +38,7 @@ public DataProviders(DataProviderContext context)
                         throw new NotSupportedException($"{ds} is not supported.");
                 }
             };
+            MdmGeneric = (IMdmDataProviderConfiguration config) => new MdmLogDecorator(context, new MdmDataProvider(_cache, config, context.RequestId, context.receivedHeaders));
```
nmallick1: Mdm = (MdmDataSource ds, IMdmDataProviderConfiguration? mdmConfiguration) =>
            {
				if(mdmConfiguration.Equals(null))
				{
					switch (ds)
					{
						case MdmDataSource.Antares:
							return new MdmLogDecorator(context, new MdmDataProvider(_cache, context.Configuration.AntaresMdmConfiguration, context.RequestId));
						case MdmDataSource.Networking:
							return new MdmLogDecorator(context, new MdmDataProvider(_cache, context.Configuration.NetworkingMdmConfiguration, context.RequestId));
						default:
							throw new NotSupportedException($"{ds} is not supported.");
					}
				}
				else
				{
					if(ds == MdmDataSource.Generic)
					{
						return new MdmLogDecorator(context, new MdmDataProvider(_cache, mdmConfiguration, context.RequestId, context.receivedHeaders));
					}
					else
					{
						throw new NotSupportedException($"{ds} is not supported.");
					}
				}
            };


The reason I am suggesting this is because if someone wants to use a Networking provider the detector code would look like
        dp.Mdm(MdmDataSource.Networking).someFunction 

Now if someone wants to use the generic Mdm, then the code would look like
               dp.MdmGeneric(mdmConfig).someFunction 

Instead, with the change I am suggesting, the code would be
For Networking : 
                  dp.Mdm(MdmDataSource.Networking).someFunction 
For any other Mdm provider 
                dp.Mdm(MdmDataSource.Generic, mdmConfig).someFunction 

There is consistency in the data provider usage and it is less confusing for people using the provider and the code is more readable.

src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
```diff
@@ -22,10 +24,17 @@ public class MdmDataProvider : DiagnosticDataProvider, IDiagnosticDataProvider,
...
         /// <param name="configuration">Data provider configuration.</param>
         /// <param name="requestId">Request id.</param>
-        public MdmDataProvider(OperationDataCache cache, IMdmDataProviderConfiguration configuration, string requestId)
+        public MdmDataProvider(OperationDataCache cache, IMdmDataProviderConfiguration configuration, string requestId, IHeaderDictionary headers = null)
             : base(cache)
         {
             _configuration = configuration;
+
+            // If the MDM account is regional, get the location from request headers and update the mdm account name with location suffix.
+            if (configuration.IsRegional && headers != null && headers.TryGetValue(HeaderConstants.LocationHeader, out var locations) && locations.Count > 0)
```
nmallick1: If we let the detector pass the correct MDM endpoint via the gist, we won't need the request headers. It can be removed.

src/Diagnostics.Logger/HeaderConstants.cs
```diff
@@ -50,6 +50,11 @@ public static class HeaderConstants
         /// </summary>
         public static readonly string GeomasterNameHeader = "x-ms-geomaster-name";
 
+        /// <summary>
+        /// Location header name.
+        /// </summary>
+        public static readonly string LocationHeader = "x-ms-location";
```
nmallick1: Applens (internal CSS accessing the diagnostics) won't be able to pass location details for all resource types. I think we should remove this. This header will be present for ARM, however the same codebase is shared for internal and external so it is best not to take dependency on this header.

src/Diagnostics.ModelsAndUtils/Models/Resource/ArmResource.cs
```diff
@@ -23,6 +23,11 @@ public class ArmResource : ArmResourceFilter, IResource
         /// </summary>
         public string Name { get; set; }
 
+        /// <summary>
+        /// Location of the resource
+        /// </summary>
+        public string Location { get; set; }
```
nmallick1: Same thing here as well about not taking a dependency on the location

Sneezry:
Hello everyone, any progress on this PR? Because we are trying to test MDM detector for Spring E2E, we depend on this PR. Is there a timeline for this PR to go live on product? Thanks 😊

PR #358, opened 2019-12-10
Add new DetectorType for CategoryOverview page

XiaoxuHU:


src/Diagnostics.ModelsAndUtils/Models/DetectorType.cs
```diff
@@ -22,6 +22,11 @@ public enum DetectorType
         /// <summary>
         /// Analysis means this is a detector that can act as an analysis
         /// </summary>
-        Analysis
+        Analysis,
+
+        /// <summary>
+        /// CategoryOverview means is the overpage for each category in external UI
```
ShekharGupta1988: nit : fix the comment : overview page :)

PR #365, opened 2019-12-11
Adding a RadioButtonList as an input type

puneetg1983:


src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Form.cs
```diff
@@ -91,19 +91,22 @@ public abstract class FormInputBase
...
+
         /// <summary>
         /// Creates an input with given id and input type
         /// </summary>
         /// <param name="id">unique id for the input</param>
         /// <param name="inputType">Input type for the input</param>
         /// <param name="label">The label of the input</param>
         /// <param name="isRequired">Indicates whether this is a required input</param>
-        public FormInputBase(int id, FormInputTypes inputType, string label, bool isRequired = false)
+        public FormInputBase(int id, FormInputTypes inputType, string label, bool isRequired = false, List<ListItem> items = null)
```
ShekharGupta1988: I dont think the ListItems object should be added to base class since its specific to radio button class. Can this move to its own class constructor

puneetg1983: The reason I added it in the base class is to avoid checking `forformInputs[ip]["inputType"]`  inside the `parseData` method in our UI code but I had to do that anyways so I believe it makes perfect sense to move the `ListItems `property to its own class and not the base class.

Another reason for having it in the Base Class was because I was thinking, we will add more input rendering types like` CheckBoxList, DropDownList` etc. but even then, I believe it makes sense to move the Items in the child class rather than the `FormInputBase `class

PR #366, opened 2019-12-11
User Story 951: Remove Stack filter call from ListDetectors

ShekharGupta1988:
I verified that no detector is using StackFilter in Definition

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
@@ -21,7 +21,7 @@ namespace Diagnostics.RuntimeHost.Controllers
             this._siteService = (ISiteService)services.GetService(typeof(ISiteService));
         }
 
-        protected async Task<App> GetAppResource(string subscriptionId, string resourceGroup, string appName, DiagnosticSiteData postBody, DateTime startTime, DateTime endTime)
+        protected async Task<App> GetAppResource(string subscriptionId, string resourceGroup, string appName, DiagnosticSiteData postBody, DateTime startTime, DateTime endTime, bool fetchStackInformation = true)
         {
             App app = new App(subscriptionId, resourceGroup, appName)
             {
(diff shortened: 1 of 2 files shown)


src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
```diff
@@ -21,7 +21,7 @@ public SiteControllerBase(IServiceProvider services, IRuntimeContext<App> runtim
             this._siteService = (ISiteService)services.GetService(typeof(ISiteService));
         }
 
-        protected async Task<App> GetAppResource(string subscriptionId, string resourceGroup, string appName, DiagnosticSiteData postBody, DateTime startTime, DateTime endTime)
+        protected async Task<App> GetAppResource(string subscriptionId, string resourceGroup, string appName, DiagnosticSiteData postBody, DateTime startTime, DateTime endTime, bool fetchStackInformation = true)
```
cindylovescoding: Should we put the default value as false Shekhar?

ShekharGupta1988: the issue is if we put it to false, then we would have to update lot of places. So the idea here is maintain the current behavior of this method, and only override the behavior at certain places where needed

nmallick1: GetDetector and other places are still going to fetch the Tenant information. The total calls will be a lot less than it is right now but for an analysis / parent child there will still be concurrent calls for tenant id. What would break if we return SiteStack.All for all of them ? Do you think detectors authors are using the stack filter attribute to define the applicability of their detectors ?

ShekharGupta1988: I guess you are talking about Stack and not Tenant. This PR is only taking care of fetching Stacks.
As I explained earlier also, the problem in removing this call from GetDetector currently is context will have Stack as StackType.All. So Detectors which are using cxt.Stack in their code will start to have issues of not showing stack specific data as they are doing right now.

The best way to attack the second problem is first we create a GIST that gets Stack Info, moves all the detectors using stack from Context to this GIST and then remove it from  our backend.

puneetg1983:
@ShekharGupta1988  - I updated all our detectors that were using `cxt.Resource.StackType` and now they are getting the stack information from the gist. You can now remove the `StackType `call from our entire backend if you want.

The detectors are now getting the stack info by calling `GetApplicationStack `method from the webappdownwindows gist.

ShekharGupta1988:
> @ShekharGupta1988 - I updated all our detectors that were using `cxt.Resource.StackType` and now they are getting the stack information from the gist. You can now remove the `StackType `call from our entire backend if you want.
> 
> The detectors are now getting the stack info by calling `GetApplicationStack `method from the webappdownwindows gist.

Cool. Thanks. I Updated the PR

PR #370, opened 2019-12-17
Create cxt.Resource.SubscriptionLocationPlacementId 

rekhaswaminathan:
- Based on incoming headers sent from the client (Portal for Diagnose and Solve), create cxt.Resource.SubscriptionLocationPlacementId which can be used in detectors. 
- Block calls to ASC based on subscription location placement id

### DIFF
--- a/src/Diagnostics.DataProviders/AscClient.cs
+++ b/src/Diagnostics.DataProviders/AscClient.cs
@@ -238,7 +238,7 @@ namespace Diagnostics.DataProviders
         {
             if (!string.IsNullOrWhiteSpace(SubscriptionLocationPlacementId) && SubscriptionLocationPlacementId.Equals(DiagAscHeaderValue, StringComparison.CurrentCultureIgnoreCase))
             {
-                throw new InvalidOperationException();
+                throw new InvalidOperationException("This subscription is not allowed for ASC calls");
             }
             var response = await SendAscRequestAsync(requestMessage, isBlobRequest, cancellationToken);
             string responseContent = await response.Content.ReadAsStringAsync();
--- a/src/Diagnostics.DataProviders/AscClient.cs
+++ b/src/Diagnostics.DataProviders/AscClient.cs
@@ -70,7 +70,11 @@ namespace Diagnostics.DataProviders
         /// </summary>
         private string apiVersion;
 
-        private readonly string MethodNotAllowed = "DataProvider method not allowed";
+        /// <summary>
+        /// Header value based on which we block ASC calls.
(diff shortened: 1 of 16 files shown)


src/Diagnostics.DataProviders/AscClient.cs
```diff
@@ -215,6 +231,10 @@ private static T CastTo<T>(object obj)
 
         private async Task<T> GetAscResponse<T>(HttpRequestMessage requestMessage, bool isBlobRequest, CancellationToken cancellationToken = default(CancellationToken))
         {
+            if (!string.IsNullOrWhiteSpace(SubscriptionLocationPlacementId) && SubscriptionLocationPlacementId.Equals("geos_2020-01-01", StringComparison.CurrentCultureIgnoreCase))
```
ShekharGupta1988: Hawk had a good point in his PR that its prob wise to put this value in the app-setting instead of code just to avoid any sort of  compliance issue

src/Diagnostics.DataProviders/AscClient.cs
```diff
@@ -215,6 +231,10 @@ private static T CastTo<T>(object obj)
...
+            {
+                throw new HttpRequestException(MethodNotAllowed);
```
ShekharGupta1988: Should we throw InvalidOperationException here?

https://docs.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?redirectedfrom=MSDN&view=netframework-4.8

src/Diagnostics.DataProviders/AscClient.cs
```diff
@@ -231,9 +236,9 @@ private static T CastTo<T>(object obj)
 
         private async Task<T> GetAscResponse<T>(HttpRequestMessage requestMessage, bool isBlobRequest, CancellationToken cancellationToken = default(CancellationToken))
         {
-            if (!string.IsNullOrWhiteSpace(SubscriptionLocationPlacementId) && SubscriptionLocationPlacementId.Equals("geos_2020-01-01", StringComparison.CurrentCultureIgnoreCase))
+            if (!string.IsNullOrWhiteSpace(SubscriptionLocationPlacementId) && SubscriptionLocationPlacementId.Equals(DiagAscHeaderValue, StringComparison.CurrentCultureIgnoreCase))
             {
-                throw new HttpRequestException(MethodNotAllowed);
+                throw new InvalidOperationException();
```
hforeste: Can we add a message saying something like "this subscription is not allowed for ASC calls". That way we can just show that message in the markdown message in the detector UI

PR #371, opened 2019-12-17
30 seconds leniency in T-15 condition to accomodate the UI to backend request time

SharmaAjay19:
30 seconds leniency in T-15 condition to accomodate the UI to backend request time


nmallick1:
Shouldn't we zero out the seconds part when using the start time and end time ?

SharmaAjay19:
> Shouldn't we zero out the seconds part when using the start time and end time ?

We are actually adding a leniency of 30 seconds from our earlier 15 minutes rule to accommodate any time delays of request between UI and backend. Hence we are using 14.5 instead of 15.

PR #374, opened 2019-12-19
Adding an outbound network connectivity check in runtime host

SharmaAjay19:


src/Diagnostics.RuntimeHost/Controllers/ProcessHealthController.cs
```diff
@@ -13,18 +15,44 @@ public class ProcessHealthController : Controller
...
-        public ProcessHealthController(ISourceWatcherService sourceWatcherService, ICompilerHostClient compilerHostClient)
+        public ProcessHealthController(ISourceWatcherService sourceWatcherService, ICompilerHostClient compilerHostClient, IConfiguration Configuration)
         {
             // These dependencies are injected for the services to start.
             _sourceWatcherService = sourceWatcherService;
             _compilerHostClient = compilerHostClient;
+            _configuration = Configuration;
+        }
+
+        public static bool PingExternal(string pingUrl)
```
ShekharGupta1988: this can be a private method instead of public

SharmaAjay19: My bad.

src/Diagnostics.RuntimeHost/Controllers/ProcessHealthController.cs
```diff
@@ -13,18 +15,44 @@ public class ProcessHealthController : Controller
...
+        {
+            try
+            {
+                using (var client = new WebClient())
```
ShekharGupta1988: We should not create a new client for each request. This can cause port exhaustion. 
Another simple way is to create a singleton HealthCheckService and initialize the client there. You can move this whole logic in that service

src/Diagnostics.RuntimeHost/Controllers/ProcessHealthController.cs
```diff
@@ -13,18 +15,44 @@ public class ProcessHealthController : Controller
...
+                return false;
+            }
         }
 
         [HttpGet(UriElements.HealthPing)]
         public async Task<IActionResult> HealthPing()
         {
             await _sourceWatcherService.Watcher.WaitForFirstCompletion();
+
+            bool isCheckEnabled = Convert.ToBoolean(_configuration["OutboundConnectivity:IsCheckEnabled"]);
```
ShekharGupta1988: Again, this can be moved to the HealthCheckService and you can read this setting once in ctor and maintain it in a local variable.

src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -104,5 +104,9 @@
     "ClientId": "",
     "AllowedAppIds": ""
   },
-  "CloudDomain": "PublicAzure"
+  "CloudDomain": "PublicAzure",
+  "OutboundConnectivity": {
+    "IsCheckEnabled": true,
+    "PingUrl": "http://google.com/generate_204"
```
hforeste: Can the URL just be one of our internal systems? I don't think this will fly in some other cloud environments.

src/Diagnostics.RuntimeHost/Controllers/ProcessHealthController.cs
```diff
@@ -13,18 +15,44 @@ public class ProcessHealthController : Controller
...
+            if (isCheckEnabled)
+            {
+                bool externalPing = PingExternal(_configuration["OutboundConnectivity:PingUrl"]);
+                if (!externalPing)
+                {
+                    return StatusCode(500);
```
ShekharGupta1988: In case of failure, can you log the exception too

src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -104,5 +104,9 @@
     "ClientId": "",
     "AllowedAppIds": ""
   },
-  "CloudDomain": "PublicAzure"
+  "CloudDomain": "PublicAzure",
+  "OutboundConnectivity": {
```
ShekharGupta1988: A nice setting name here could be HealthCheckSettings --> IsOutboundConnectivityCheckEnabled

src/Diagnostics.RuntimeHost/appsettings.json
SharmaAjay19: The url is configurable. It has to outside our own app ecosystem because if any of our internal system is down, we don't want runtime host to think it has lost network connectivity. So we use an external URL. For other clouds the check can be disabled and later enabled when we have the correct URL to use.

src/Diagnostics.RuntimeHost/Controllers/ProcessHealthController.cs
ShekharGupta1988: However, not sure if the exception will make it to kusto or not.. lol

src/Diagnostics.RuntimeHost/Services/HealthCheckService/HealthCheckService.cs
```diff
@@ -0,0 +1,74 @@
...
+        bool IsOutboundConnectivityCheckEnabled = false;
+
+        public HealthCheckService(IConfiguration Configuration)
+        {
+            _configuration = Configuration;
+            IsOutboundConnectivityCheckEnabled = Convert.ToBoolean(_configuration["HealthCheckSettings:IsOutboundConnectivityCheckEnabled"]);
+            if (IsOutboundConnectivityCheckEnabled)
+            {
+                OutboundConnectivityCheckUrl = _configuration["HealthCheckSettings:OutboundConnectivityCheckUrl"];
+                if (OutboundConnectivityCheckUrl != null)
```
ShekharGupta1988: if null, should we throw a invalid config exception

src/Diagnostics.RuntimeHost/Services/HealthCheckService/HealthCheckService.cs
```diff
@@ -0,0 +1,74 @@
...
+        public async Task RunHealthCheck(bool retry)
+        {
+            HttpRequestMessage request = new HttpRequestMessage(HttpMethod.Get, OutboundConnectivityCheckUrl);
+            try
+            {
+                var response = await Get(request);
+            }
+            catch (Exception ex)
+            {
+                if (retry)
```
ShekharGupta1988: We wrote a RetryAsync  utility sometime back.  we can prob use that as well.

https://github.com/Azure/Azure-AppServices-Diagnostics/blob/7b4bd51ad8b0c228e2c7a85421e755714c328556/src/Diagnostics.RuntimeHost/Utilities/RetryHelper.cs#L10

src/Diagnostics.RuntimeHost/Services/HealthCheckService/HealthCheckService.cs
ShekharGupta1988: also check or empty or null

src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -104,5 +104,9 @@
     "ClientId": "",
     "AllowedAppIds": ""
   },
-  "CloudDomain": "PublicAzure"
+  "CloudDomain": "PublicAzure",
+  "HealthCheckSettings": {
+    "IsOutboundConnectivityCheckEnabled": true,
+    "OutboundConnectivityCheckUrl": "http://google.com/generate_204"
```
hforeste: Just curious, how did you hear about this endpoint?

src/Diagnostics.RuntimeHost/Services/HealthCheckService/HealthCheckService.cs
```diff
@@ -0,0 +1,74 @@
...
+                {
+                    await RunHealthCheck(false);
+                }
+                else
+                {
+                    throw ex;
```
hforeste: Use `throw` versus `throw ex`.

src/Diagnostics.RuntimeHost/Controllers/ProcessHealthController.cs
```diff
@@ -13,18 +12,22 @@ public class ProcessHealthController : Controller
...
             _sourceWatcherService = sourceWatcherService;
             _compilerHostClient = compilerHostClient;
+            _healthCheckService = healthCheckService;
         }
 
         [HttpGet(UriElements.HealthPing)]
         public async Task<IActionResult> HealthPing()
         {
             await _sourceWatcherService.Watcher.WaitForFirstCompletion();
+            await _healthCheckService.RunHealthCheck(true);
```
hforeste: Can we make these concurrent calls? `var healthTasks = arrary[sourcewatchertest, outboundtest]...Task.WhenAll(healthTasks)`

src/Diagnostics.RuntimeHost/appsettings.json
SharmaAjay19: https://stackoverflow.com/questions/2031824/what-is-the-best-way-to-check-for-internet-connectivity-using-net

PR #380, opened 2020-01-02
Adding InitializeAppInsightsKey to read key from Tags

puneetg1983:
Detector code will do something like this...

```
    var isKeyConfigured = await dp.AppInsights.InitializeAppInsightsKey(cxt);
    if (isKeyConfigured)
    {
        res.AddMarkdownView("Yes, key is configured");
        var appInsightsData = await dp.AppInsights.ExecuteAppInsightsQuery("requests | where timestamp > ago(1d) | take 5 | project name, url");
        res.Dataset.Add(new DiagnosticData()
        {
            Table = appInsightsData,
            RenderingProperties = new TableRendering()
        });
    }
```

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/AppInsightsDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/AppInsightsDataProvider.cs
@@ -32,7 +32,7 @@ namespace Diagnostics.DataProviders
             return Task.FromResult(true);
         }
 
-        public async Task<bool> InitializeAppInsightsKey(OperationContext<IResource> cxt)
+        public async Task<bool> SetAppInsightsKey(OperationContext<IResource> cxt)
         {
             bool keyFound = false;
             if (cxt.Resource is App app)
@@ -46,7 +46,7 @@ namespace Diagnostics.DataProviders
             return keyFound;
         }
 
-        public string DecryptString(string encryptedString)
+        private string DecryptString(string encryptedString)
         {
             byte[] iv = new byte[16];
             byte[] buffer = Convert.FromBase64String(encryptedString);
--- a/src/Diagnostics.DataProviders/DataProviders/AppInsightsDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/AppInsightsDataProvider.cs
@@ -1,13 +1,22 @@
-﻿using System.Data;
+﻿using System;
+using System.Collections.Generic;
+using System.Data;
+using System.IO;
+using System.Linq;
+using System.Security.Cryptography;
+using System.Text;
 using System.Threading.Tasks;
+using System.Xml.Linq;
 using Diagnostics.DataProviders.DataProviderConfigurations;
 using Diagnostics.DataProviders.Interfaces;
 using Diagnostics.ModelsAndUtils.Models;
+using Newtonsoft.Json;
 
 namespace Diagnostics.DataProviders
 {
     public class AppInsightsDataProvider : DiagnosticDataProvider, IDiagnosticDataProvider, IAppInsightsDataProvider
     {
+        const string AppInsightsTagName = "hidden-related:diagnostics/applicationInsightsSettings";
         private readonly IAppInsightsClient _appInsightsClient;
         private AppInsightsDataProviderConfiguration _configuration;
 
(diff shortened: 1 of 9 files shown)


src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
```diff
@@ -34,8 +34,9 @@ protected async Task<App> GetAppResource(string subscriptionId, string resourceG
                 Stamp = await GetHostingEnvironment(postBody.Stamp.Subscription, postBody.Stamp.ResourceGroup, postBody.Stamp != null ? postBody.Stamp.Name : string.Empty, postBody.Stamp, startTime, endTime, postBody.Kind),
                 AppType = GetApplicationType(postBody.Kind),
                 PlatformType = (!string.IsNullOrWhiteSpace(postBody.Kind) && postBody.Kind.ToLower().Contains("linux")) ? PlatformType.Linux : PlatformType.Windows,
-                StackType =  StackType.All //await this._siteService.GetApplicationStack(subscriptionId, resourceGroup, appName, (DataProviderContext)HttpContext.Items[HostConstants.DataProviderContextKey])
+                StackType =  StackType.All, //await this._siteService.GetApplicationStack(subscriptionId, resourceGroup, appName, (DataProviderContext)HttpContext.Items[HostConstants.DataProviderContextKey])
                 // Not Fetching Stack Information anymore for kusto optimizations. Instead, detectors should look to leverage a gist created for the same.
+                Tags = postBody.Tags
```
nmallick1: Will requests on applens also contain the Tags information ?

src/Diagnostics.DataProviders/DataProviders/AppInsightsDataProvider.cs
```diff
@@ -23,6 +32,76 @@ public Task<bool> SetAppInsightsKey(string appId, string apiKey)
...
+                var tag = GetAppIdAndKeyFromAppSettingsTags(app.Tags);
+                if (tag != null)
+                {
+                    keyFound = await SetAppInsightsKey(tag.AppId, DecryptString(tag.ApiKey));
+                }
+            }
+            return keyFound;
+        }
+
+        public string DecryptString(string encryptedString)
```
ShekharGupta1988: This should be Private method

src/Diagnostics.ModelsAndUtils/Models/AppInsights/AppInsightsTag.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Diagnostics.ModelsAndUtils.Models
+{
+    public class AppInsightsTag
```
ShekharGupta1988: Question here - Do we use this model in the detector code somewhere? If not, you can potentially just move this class to the project where its used (Data Provider project).

src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -59,6 +59,7 @@
     "Token": ""
   },
   "AppInsights": {
+    "EncryptionKey": ""
```
cindylovescoding: Is this EncryptionKey set by us to decrypt the apikey? How do we know which key to use?

src/Diagnostics.DataProviders/Interfaces/IAppInsightsDataProvider.cs
```diff
@@ -8,5 +9,7 @@ public interface IAppInsightsDataProvider : IMetadataProvider
         Task<DataTable> ExecuteAppInsightsQuery(string query);
 
         Task<bool> SetAppInsightsKey(string appId, string apiKey);
+
+        Task<bool> InitializeAppInsightsKey(OperationContext<IResource> cxt);
```
ShekharGupta1988: nitpick here :- Does it make sense to have both the method names same and use method overloading.

cindylovescoding:
Hey @puneetg1983, will that be the same to add all the rendering types for app insights data response?

    var appInsightsData = await dp.AppInsights.ExecuteAppInsightsQuery("requests | where timestamp > ago(1d) | take 5 | project name, url");
        res.Dataset.Add(new DiagnosticData()
        {
            Table = appInsightsData,
            RenderingProperties = new TableRendering()
        });

src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
ShekharGupta1988: Also, we might want to check geomaster code to see if they are getting populated or not

src/Diagnostics.DataProviders/DataProviders/AppInsightsDataProvider.cs
puneetg1983: Will do

src/Diagnostics.ModelsAndUtils/Models/AppInsights/AppInsightsTag.cs
puneetg1983: Nope here only....will make the chang

src/Diagnostics.DataProviders/Interfaces/IAppInsightsDataProvider.cs
puneetg1983: Yup it does

src/Diagnostics.RuntimeHost/appsettings.json
puneetg1983: Yes, we will set in manually here and in backend project of portal code

src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
puneetg1983: I tested applens locally and the tags were populated...

Sorry i am not following what else you want me to tes

ShekharGupta1988: So this postBody object for  applens calls is populated by runtime host itself but for external calls, its populated by geomaster (since its already has the info) and sent as part of request. I was just wondering if we are populating this tags object in geomaster as well or not before sending the request. Most likely we will be doing it there as well but we might just want to quickly check the code there.  @rekhaswaminathan can help locating the code in geomaster which is doing this operation.

puneetg1983: I see....will wait for @rekhaswaminathan to respond.

So that will be in Antares code I guess

rekhaswaminathan: @puneetg1983  @ShekharGupta1988 
GeoMaster is sending it as part of postbody:
https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites?path=%2Fsrc%2FHosting%2FAdministrationService%2FMicrosoft.Web.Hosting.Administration.GeoScale.Csm%2FControllers%2FDiagnosticCsmController.cs&version=GBdev&line=494&lineEnd=495&lineStartColumn=1&lineEndColumn=1&lineStyle=plain

puneetg1983: Perfect! Thank you so much @rekhaswaminathan

PR #384, opened 2020-01-09
Rekha/passcxt childdetectors

rekhaswaminathan:
This gives detector authors ability to pass any key value pairs to child detectors from parent detector. 
Example code for passing key-value pairs:
```
  var additionalParams = new Dictionary<string, string>();
            additionalParams.Add("Input", "ABCD");
            res.AddDetector("detectorqp", additionalParams ); 
```

```
  var children = new List<string>();
    children.Add("http4xx");
    children.Add("detectorqp");
    var additionalParams = new Dictionary<string, string>();
    additionalParams.Add("Input","abcd");
    res.AddDetectorCollection(children, additionalParams);
```

Example code in child detector to access these key-value pairs:

```
 if(cxt.QueryParams != null) {
       var keys = cxt.QueryParams.Keys;
       string markdown = "";
       foreach(string key in keys) {
           markdown += $"{key} = {cxt.QueryParams[key]} \n\n";
       }
       res.AddMarkdownView(markdown, "Detector Query Params received");
   }
```

image

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/OperationContext.cs
@@ -1,6 +1,7 @@
 ﻿using Diagnostics.ModelsAndUtils.Attributes;
 using Diagnostics.ModelsAndUtils.Models.ResponseExtensions;
 using Diagnostics.ModelsAndUtils.Utilities;
+using System.Collections.Generic;
 
 namespace Diagnostics.ModelsAndUtils.Models
 {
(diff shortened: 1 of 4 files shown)


cindylovescoding:
@rekhaswaminathan This is Awesome! Two questions:
1. If we use use addtional parameters in AddDetectorCollection, those will be the common parameters passed down to each detector in the collection right?
2. Do we plan to pass other objects other than string dictionary in the future?

rekhaswaminathan:
> @rekhaswaminathan This is Awesome! Two questions:
> 
> 1. If we use use addtional parameters in AddDetectorCollection, those will be the common parameters passed down to each detector in the collection right?
> 2. Do we plan to pass other objects other than string dictionary in the future?

1. Yes Cindy, Let's say you open the child detector separately (not from Parent), those common parameters can still be seen 
image

2. I am not sure about this, I think its going to be string for now because we set it from url's query string parameters

PR #388, opened 2020-01-13
Creating overloaded method that accepts additionalParams

rekhaswaminathan:
Previously I added additionalParams as optional param that can causes pre compiled detectors to break. So changed to overloaded method.

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorCollection.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorCollection.cs
@@ -21,14 +21,14 @@ namespace Diagnostics.ModelsAndUtils.Models.ResponseExtensions
             return diagnosticData;
         }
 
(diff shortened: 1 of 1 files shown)


src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/DetectorCollection.cs
```diff
@@ -21,14 +21,47 @@ public static DiagnosticData AddDetectorCollection(this Response response, List<
...
 
-        public static DiagnosticData AddDetector(this Response response, string detectorId, IDictionary<string, string> additionalParams = null )
+        public static DiagnosticData AddDetectorCollection(this Response response, List<string> detectorIds, IDictionary<string, string> additionalParams)
+        {
+            var diagnosticData = new DiagnosticData()
+            {
+                RenderingProperties = new DetectorCollectionRendering()
+                {
+                    DetectorIds = detectorIds,
+                    AdditionalParams = JsonConvert.SerializeObject(additionalParams),
```
ShekharGupta1988: you should prob check for null here for the dictionary

PR #397, opened 2020-01-25
National Cloud Updates

hforeste:
- Remove static KeyVault URL and make it configurable so that it can be configured for different clouds
- By default `AzureServiceTokenProvider` uses public AAD authority, instead make the authority configurable and pass it into the class's constructor
- Log unhanded exceptions to app service application logs and windows event logs
- Create configurable feature flags for each data provider and compilerhost
- Add dependency checker API
- Create overload of ExecuteClusterQuery which takes cluster and database as parameters. This will stop kusto queries from proxying through our clusters
- Remove static text for AAD resource for Observer and create configurable AAD resource.

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/ProcessHealthController.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/ProcessHealthController.cs
@@ -5,6 +5,8 @@ using Diagnostics.RuntimeHost.Utilities;
 using Microsoft.AspNetCore.Mvc;
 using System.Threading.Tasks;
 using System.Collections.Generic;
+using Diagnostics.DataProviders;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
 
 namespace Diagnostics.RuntimeHost.Controllers
 {
@@ -33,6 +35,13 @@ namespace Diagnostics.RuntimeHost.Controllers
             await Task.WhenAll(allChecks);
 
             return Ok("Server is up and running.");
+        }
+
+        [HttpGet("/dependencyCheck")]
+        public async Task<IEnumerable<HealthCheckResult>> DependencyCheck()
+        {
+            var dataProviders = new DataProviders.DataProviders((DataProviderContext)HttpContext.Items[HostConstants.DataProviderContextKey]);
+            return await _healthCheckService.RunDependencyCheck(dataProviders);
         }
     }
 }
--- a/src/Diagnostics.RuntimeHost/Program.cs
+++ b/src/Diagnostics.RuntimeHost/Program.cs
@@ -4,6 +4,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Azure.KeyVault;
 using Microsoft.Azure.Services.AppAuthentication;
 using Microsoft.Extensions.Configuration.AzureKeyVault;
+using Microsoft.Extensions.Logging;
 using Diagnostics.DataProviders;
 using System;
 using Microsoft.CodeAnalysis;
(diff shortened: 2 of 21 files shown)


src/Diagnostics.RuntimeHost/Program.cs
```diff
@@ -41,32 +42,37 @@ public static IWebHost BuildWebHost(string[] args)
...
                             keyVaultClient,
                             new DefaultKeyVaultSecretManager())
                         .AddEnvironmentVariables()
                         .AddCommandLine(args)
                         .Build();
                 })
+                .ConfigureLogging(loggingBuilder =>
+                {
+                    //Send log output to text file application logs
+                    loggingBuilder.AddAzureWebAppDiagnostics();
```
rekhaswaminathan: Correct me if I'm wrong here, but is it a good practice to have application logs in text file in the file system as they can take up space?

hforeste: There is a size limit and when its reached it just starts cleaning away older files. Also it honors the app service settings, so if we disable in the portal then it won't write logs. https://github.com/Azure/Azure-AppServices-Diagnostics/pull/397/files#diff-9d6691fbc408270322f673ad60cb0c82

src/Diagnostics.RuntimeHost/Controllers/ProcessHealthController.cs
```diff
@@ -34,5 +36,12 @@ public async Task<IActionResult> HealthPing()
 
             return Ok("Server is up and running.");
         }
+
+        [HttpGet("/dependencyCheck")]
+        public async Task<IEnumerable<HealthCheckResult>> DependencyCheck()
```
ShekharGupta1988: great addition :). This is something Me,Rekha and Ajay were discussing too last week for Liberation endpoints.

src/Diagnostics.RuntimeHost/Program.cs
ShekharGupta1988: Any reason for using this logging instead of relying on kusto and/or appinsights which are way more scalable?

src/Diagnostics.RuntimeHost/Services/HealthCheckService/HealthCheckService.cs
```diff
@@ -54,6 +58,14 @@ public async Task RunHealthCheck()
             await RetryHelper.RetryAsync(HealthCheckPing, "Healthping", "", 3, 100);
         }
 
+        public async Task<IEnumerable<HealthCheckResult>> RunDependencyCheck(DataProviders.DataProviders dataProviders)
+        {
+            var results = new List<HealthCheckResult>();
+            DiagnosticDataProvider dp = ((LogDecoratorBase)dataProviders.Kusto).DataProvider;
```
ShekharGupta1988: should we get all data providers here and run health check for each one of them?

hforeste: I have that change and it will go into a different PR. This was just part 1 of something that works and I wanted to get it reviewed rather than send one big PR

src/Diagnostics.RuntimeHost/Program.cs
hforeste: Unfortunately AppInsights is not easily consumable in these clouds at the moment and we need to make changes to make our subscription and app a first party app to get information sent to Kusto. This was the quickest way to get a sense of what is going on in the system while we get all the other pieces together to make it send log data to Kusto.

PR #398, opened 2020-01-29
Updating auto heal rules

rekhaswaminathan:
Also renamed file to follow convention


infrastructure/ArmTemplates/antareskvtemplate.json
```diff
@@ -370,9 +370,19 @@
                   "subStatus": 0,
                   "count": 100,
                   "timeInterval": "00:10:00"
-                }
+                },
+                {
+                  "status": 502,
+                  "subStatus": 0,
```
ShekharGupta1988: quick question - will this work for 502.3 errors as well, Basically all 502 errors?

puneetg1983: >subStatus [](start = 19, length = 9)

We should just remove the substatus completely so we get data whenever a 502 happens so we are covered for all

puneetg1983: I just validated by looking at AutoHealing source. If we are setting substatus or win32Status to 0, then it covers status+sub-status scenarios of 502. So the current configuration is accurate and will catch 502.3 and other substatus as well.

PR #399, opened 2020-01-29
ASC bridge with support topic mapping for all partner products

SharmaAjay19:
Modification to the GetInsights method of the controller of every resource type to accept a dynamic post body (where it isn't accepting already) and a supportTopic parameter which is the support topic path.

Pass the post body and supportTopicPath parameter to the base controller i.e. DiagnosticsControllerBase.

This controller takes care of converting the supportTopicPath into a product Id and Support topic Id which is further used in the existing flow of the api.

A support topic service created to carry out the conversion using a support topic path to id mapping obtained from kusto query.

### DIFF
--- a/src/Diagnostics.RuntimeHost/Services/SupportTopicService/SupportTopicService.cs
+++ b/src/Diagnostics.RuntimeHost/Services/SupportTopicService/SupportTopicService.cs
@@ -44,7 +44,7 @@ namespace Diagnostics.RuntimeHost.Services
             string query = $@"cluster('azsupport').database('AzureSupportability').ActiveSupportTopicTree
                             | where Timestamp > ago(3d)
                             | summarize by ProductId, SupportTopicId = SupportTopicL3Id, ProductName, SupportTopicL2Name, SupportTopicL3Name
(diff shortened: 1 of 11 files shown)


src/Diagnostics.RuntimeHost/Services/SupportTopicService/SupportTopicService.cs
```diff
@@ -0,0 +1,75 @@
...
+        public SupportTopicService()
+        {
+            _supportTopicCache = new Dictionary<string, SupportTopicModel>();
+        }
+
+        private string GetSupportTopicKustoQuery() {
+            string query = $@"cluster('azsupport').database('AzureSupportability').ActiveSupportTopicTree
+                            | where Timestamp > ago(3d)
+                            | summarize by ProductId, SupportTopicId = SupportTopicL3Id, ProductName, SupportTopicL2Name, SupportTopicL3Name
+                            | where SupportTopicId != '' //and SupportTopicL2Name != '' //and SupportTopicL3Name != ''
```
nmallick1: nit: Possible remove the comments from this query

SharmaAjay19: My bad.

PR #400, opened 2020-01-30
HyperV Platform Type

dagrooms52:
- HyperV is detected by "xenon" in postbody `Kind` string


src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
```diff
@@ -56,9 +58,17 @@ protected async Task<App> GetAppResource(string subscriptionId, string resourceG
...
                 app.SubscriptionLocationPlacementId = subscriptionLocationPlacementId.FirstOrDefault();
             }
+
             return app;
         }
 
+        private PlatformType GetPlatformType(DiagnosticSiteData postBody)
+        {
+            return postBody.IsXenon ? PlatformType.HyperV | PlatformType.Windows :
+                postBody.Kind != null && postBody.Kind.ToLower().Contains("linux") ? PlatformType.Linux :
```
hforeste: use `string.IsNullOrWhiteSpace` instead of `postBody.Kind != null`

src/Diagnostics.RuntimeHost/Models/PostBodyModel.cs
```diff
@@ -12,6 +12,7 @@ public class DiagnosticSiteData
         public string NamespaceDescriptor { get; set; }
         public bool IsDefaultContainer { get; set; }
         public bool? IsLinux { get; set; }
+        public bool IsXenon => Kind != null && Kind.ToLower().Contains("xenon");
```
hforeste: use `string.IsNullOrWhiteSpace` instead of `Kind != null`

dagrooms52: I felt that this form was more succinct given `Contains` will return `false` on a whitespace/empty string

src/Diagnostics.RuntimeHost/Services/ResourceService/StampService.cs
```diff
@@ -46,15 +46,16 @@ public StampService()
 
             var windowsTenantIdsTask = GetTenantIdsAsync(stamp, startTime, endTime, dataProviderContext, PlatformType.Windows);
             var linuxTenantIdsTask = GetTenantIdsAsync(stamp, startTime, endTime, dataProviderContext, PlatformType.Linux);
+            await Task.WhenAll(windowsTenantIdsTask, linuxTenantIdsTask);
```
hforeste: You dont have to do this since they are awaited in the below 2 lines of code

src/Diagnostics.RuntimeHost/Services/ResourceService/StampService.cs
```diff
@@ -46,15 +46,16 @@ public StampService()
...
             var linuxTenantIds = await linuxTenantIdsTask;
 
-            PlatformType type = PlatformType.Windows;
             List<string> tenantIds = windowsTenantIds.Union(linuxTenantIds).ToList();
 
+            PlatformType type = PlatformType.Windows | PlatformType.HyperV;
             if (windowsTenantIds.Any() && linuxTenantIds.Any())
             {
-                type = PlatformType.Windows | PlatformType.Linux;
+                type = PlatformType.Windows | PlatformType.HyperV | PlatformType.Linux;
```
hforeste: Can platform be both HyperV and Linux?

src/Diagnostics.RuntimeHost/Services/ResourceService/StampService.cs
dagrooms52: That could result in `UnobservedTaskException` if the first `Task` throws.
https://stackoverflow.com/a/49952821/3524426

src/Diagnostics.RuntimeHost/Services/ResourceService/StampService.cs
hforeste: Ahh nevermind, It can be a megastamp of Linux and Windows machines is it?

dagrooms52: From the old diff, platform could be both Windows and Linux, so I don't think this code was meant to differentiate between the two.

dagrooms52: Right, I think we need more information about Stamp behavior with HyperV

ShekharGupta1988: Guys - We should only set HyperV platform if IsXenon is set to true. Am I missing anything ?

dagrooms52: I think this infers platform type of the stamp based on the presence of Windows or Linux tenant IDs. So this probably should not be set yet, and I'll get clarification on how to tell if the stamp runs Xenon.

src/Diagnostics.RuntimeHost/Services/ResourceService/StampService.cs
hforeste: Oh this is interesting, I haven't heard of this before. I was learning a little about it and I think you should be fine actually. Apparently as long as you have a reference to the task and you are not doing a fire and forget then you should be fine. 
https://stackoverflow.com/questions/3284137/taskscheduler-unobservedtaskexception-event-handler-never-being-triggered

dagrooms52: I can't reproduce it in .NET Core actually, so it's probably not an issue anymore

hforeste:
Your referring to https://github.com/Azure/Azure-AppServices-Diagnostics/blob/master2/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs#L502 when you say HostingEnvironment type is inferred from the app type.

dagrooms52:
> Your referring to https://github.com/Azure/Azure-AppServices-Diagnostics/blob/master2/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs#L502 when you say HostingEnvironment type is inferred from the app type.

That refers to this line: https://github.com/Azure/Azure-AppServices-Diagnostics/pull/400/files#diff-daf7bfd99599be41a31e9a2c497aa72aR502

src/Diagnostics.RuntimeHost/Models/PostBodyModel.cs
```diff
@@ -12,6 +12,7 @@ public class DiagnosticSiteData
         public string NamespaceDescriptor { get; set; }
         public bool IsDefaultContainer { get; set; }
         public bool? IsLinux { get; set; }
+        public bool IsXenon => !string.IsNullOrWhiteSpace(Kind) && Kind.ToLower().Contains("xenon");
```
ShekharGupta1988: This logic would need a change I believe after Joaquin's email?

dagrooms52: I pushed the latest update this morning

PR #406, opened 2020-02-04
Use RuntimeSiteSlotMap EndTime Query Parameter

dagrooms52:
- This allows Observer to send a cached slotmap response
- Detectors should be updated to pass the endtime or use param `cxt` to have endTime handled for them

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
@@ -62,24 +62,40 @@ namespace Diagnostics.DataProviders
             return JsonConvert.DeserializeObject(result);
         }
 
-        public override Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string stampName, string siteName)
+        public override Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string stampName, string siteName, DateTime? endTime = null)
         {
-            return GetRuntimeSiteSlotMap(stampName, siteName, null);
+            return GetRuntimeSiteSlotMap(stampName, siteName, null, endTime);
         }
 
-        public override async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string stampName, string siteName, string slotName = null)
+        public override async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(string stampName, string siteName, string slotName = null, DateTime? endTime = null)
         {
-            return await GetRuntimeSiteSlotMapInternal(stampName, siteName, slotName);
+            return await GetRuntimeSiteSlotMapInternal(stampName, siteName, slotName, endTime);
         }
 
-        private async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMapInternal(string stampName, string siteName, string slotName)
+        public override async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(OperationContext<App> cxt, string stampName = "", string siteName = "", string slotName = "", DateTime? endTime = null)
+        {
+            if (cxt?.Resource?.Stamp == null)
+                throw new ArgumentNullException(nameof(cxt));
+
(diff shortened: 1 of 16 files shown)


hforeste:
Hey Daniel, you may not need to make a change to the signature of the RuntimeSiteSlotMap methods. You can use the `endtime` supplied by the `DataProviderContext` object which is passed to the `SupportObserverDataProviderBase` constructor. 

https://github.com/Azure/Azure-AppServices-Diagnostics/blob/master2/src/Diagnostics.DataProviders/DataProviderContext.cs#L12

dagrooms52:
> Hey Daniel, you may not need to make a change to the signature of the RuntimeSiteSlotMap methods. You can use the `endtime` supplied by the `DataProviderContext` object which is passed to the `SupportObserverDataProviderBase` constructor.
> 
> https://github.com/Azure/Azure-AppServices-Diagnostics/blob/master2/src/Diagnostics.DataProviders/DataProviderContext.cs#L12

Thanks, that makes this a lot easier!

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -72,14 +72,30 @@ public override async Task<dynamic> GetCertificatesInResourceGroupAsync(string s
...
 
-        private async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMapInternal(string stampName, string siteName, string slotName)
+        public override async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMap(OperationContext<App> cxt, string stampName = null, string siteName = null, string slotName = null, DateTime? endTime = null)
+        {
+            if (cxt?.Resource?.Stamp == null)
+                throw new ArgumentNullException(nameof(cxt));
+
+            stampName = string.IsNullOrWhiteSpace(stampName) ? cxt.Resource.Stamp.InternalName : stampName;
+            siteName = string.IsNullOrWhiteSpace(siteName) ? cxt.Resource.Name : siteName;
+            endTime = endTime == null ? DateTime.SpecifyKind(DateTime.Parse(cxt.EndTime), DateTimeKind.Utc) : endTime;
```
hforeste: The OperationContext already has EndTime in UTC format

hforeste: The `endTime` parameter probably should probably be checked. Since you are accepting the OperationContext, you probably don't need these optional parameters right?

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
```diff
@@ -72,14 +72,30 @@ public override async Task<dynamic> GetCertificatesInResourceGroupAsync(string s
...
+        private async Task<Dictionary<string, List<RuntimeSitenameTimeRange>>> GetRuntimeSiteSlotMapInternal(string stampName, string siteName, string slotName, DateTime? endTime = null)
         {
             if (string.IsNullOrWhiteSpace(stampName))
                 throw new ArgumentNullException(nameof(stampName));
             if (string.IsNullOrWhiteSpace(siteName))
                 throw new ArgumentNullException(nameof(siteName));
 
-            var result = await GetObserverResource($"stamps/{stampName}/sites/{siteName}/runtimesiteslotmap");
+            endTime = endTime == null ? DataProviderContext.QueryEndTime : endTime;
+            var endTimeUtcString = $"{endTime.Value.ToUniversalTime():O}";
```
hforeste: Whether the endTime comes from the OperationContext or the DataProviderContext, it will already be in UTC so perhaps you don't need to do the ToUniversalTime conversion.

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
dagrooms52: cxt.EndTime was copypasta, but the optional parameters are there as overrides so you aren't thrown completely out of the `cxt` call if you want a different endTime, or the same siteName on a different stamp

src/Diagnostics.DataProviders/DataProviders/SupportObserverDataProvider.cs
dagrooms52: This has to be there for the overrides to work in case a detector passes in `DateTime.Now` or some other potentially-localized DateTime
```
$"{DateTime.Now.ToUniversalTime():O}"
"2020-02-04T21:46:33.9415616Z"
$"{DateTime.Now:O}"
"2020-02-04T13:46:39.4756676-08:00"
```

PR #414, opened 2020-02-11
Add port number to host name

puneetg1983:


src/Diagnostics.ModelsAndUtils/ScriptUtilities/Utilities.cs
```diff
@@ -116,18 +116,33 @@ public static string HostNamesFilterQuery(IEnumerable<string> hostnames, string
...
 
             return hostNameQuery;
         }
 
+        private static IEnumerable<string> AddPortNumberToHostNames(IEnumerable<string> hostnames)
+        {
+            var hostNamesWithPort = new List<string>();
+            foreach (var host in hostnames)
+            {
+                hostNamesWithPort.Add(host);
```
dagrooms52: What's the possibility of a "dirty" input, something with a port already on it? Maybe to clean this you can start with a base hostname of `host.Split(':').First()`

puneetg1983: most of the detectors are just passing cxt.Resource.Hostnames which should just be the hostnames configured on the app so the possibility of a dirty input is quite less. Worst case we will add the original input and another dirty input with the port twice in it but the overall logic of the KUSTO query should still be fine

PR #415, opened 2020-02-12
Self service kusto configuration backend

hforeste:
This feature enables partners to set up their own kusto cluster and database mappings and be automatically picked up by our backend.


src/Diagnostics.DataProviders/KustoMap.cs
```diff
@@ -0,0 +1,148 @@
...
+                            else if (pair.Key.ToLower().Contains("databasename"))
+                            {
+                                databaseRow[column] = pair.Value;
+                            }
+                        }
+                    }
+                }
+            }
+
+            //todo add logging of entries that didn't get added to the map
```
rekhaswaminathan: Are you going to add the logging?

src/Diagnostics.RuntimeHost/Controllers/Configuration/KustoClusterMappingsController.cs
```diff
@@ -0,0 +1,85 @@
...
+                return Ok(value);
+            }
+
+            return NotFound();
+        }
+
+        [HttpDelete]
+        public Task<IActionResult> DeleteMapping(string provider)
+        {
+            throw new NotImplementedException();
```
rekhaswaminathan: Are we going to support this in the future?

src/Diagnostics.DataProviders/KustoMap.cs
hforeste: Yes

src/Diagnostics.RuntimeHost/Controllers/Configuration/KustoClusterMappingsController.cs
hforeste: I might, so I will just keep it here.

src/Diagnostics.RuntimeHost/Controllers/Configuration/KustoClusterMappingsController.cs
```diff
@@ -0,0 +1,85 @@
...
+            if (_kustoMappingsCache.TryGetValue(cacheId, out List<Dictionary<string, string>> cacheValue) && cacheValue.Equals(kustoMappings))
+            {
+                return Ok();
+            }
+
+            var gitHubPackage = new GithubPackage(cacheId, "kustoClusterMappings", "json", JsonConvert.SerializeObject(kustoMappings));
+
+            try
+            {
+                await _sourceWatcherService.Watcher.CreateOrUpdatePackage(gitHubPackage);
```
rekhaswaminathan: Should we start the try in the beginning itself because we have `_sourceWatcherService` call before this

hforeste: Good idea

PR #417, opened 2020-02-13
National Cloud updates

hforeste:
- Add ARM template to Add certificate from keyvault
- Add Unit tests for `KustoMap` data structure
- Find certs for dependencies by subject name
- Fix `GetClusterNameFromStamp` bug where it return null in national clouds
- Dynamically modify Kusto queries to be cloud agnostic
- Extend dependency checker to other data providers and the `sourcewatcher`
- Create feature flag for each data provider with the `Enabled` switch


src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -54,19 +54,23 @@
     "HeartBeatTimeOutInSeconds": 10,
     "HeartBeatDelayInSeconds": 20,
     "AADAuthority": "https://login.windows.net/microsoft.com",
-    "AADKustoResource": "https://wawskustotest.kusto.windows.net"
+    "AADKustoResource": "https://wawskustotest.kusto.windows.net",
+    "Enabled": true
```
rekhaswaminathan: We have to update these new settings to our key vault right.. Will it cause any breaking change ?

rekhaswaminathan:
Question about health check, will it be applicable to dotnet process running inside DiagRole.

src/Diagnostics.RuntimeHost/appsettings.json
hforeste: No, whatever configuration is not in KeyVault is picked up from the AppSettings

PR #419, opened 2020-02-20
Remove console logging provider

cindylovescoding:


src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -90,6 +91,19 @@ public void ConfigureServices(IServiceCollection services)
             {
                 services.AddSingleton<ISearchService, SearchServiceDisabled>();
             }
+
+            services.AddLogging(config =>
+            {
+                config.ClearProviders();
+                config.AddConfiguration(Configuration.GetSection("Logging"));
+                config.AddDebug();
+                config.AddEventSourceLogger();
```
rekhaswaminathan: do you think we need this? `addEventSourceLogger()`

src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -90,6 +91,19 @@ public void ConfigureServices(IServiceCollection services)
             {
                 services.AddSingleton<ISearchService, SearchServiceDisabled>();
             }
+
+            services.AddLogging(config =>
+            {
+                config.ClearProviders();
```
hforeste: I've added some logging awhile ago for EventLog and App Service Logs. This was primarily because I wasn't getting logs in sovereign clouds in kusto so I had to find alternative way. 

https://github.com/Azure/Azure-AppServices-Diagnostics/blob/2d2be6a5d4592939a6852e865651a11332a7c6ae/src/Diagnostics.RuntimeHost/Program.cs#L56-L61

Will `conflig.ClearProviders()` remove the logging that I added in the `Program.cs` file.

PR #420, opened 2020-02-21
Remove console logging provider in master2

cindylovescoding:


src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -186,8 +187,22 @@ public void ConfigureServices(IServiceCollection services)
...
                 services.AddSingleton<ISearchService, SearchServiceDisabled>();
             }
-            // Initialize on startup
             servicesProvider.GetService<ISourceWatcherService>();
+
+            services.AddLogging(loggingConfig =>
+            {
+                loggingConfig.ClearProviders();
+                loggingConfig.AddConfiguration(Configuration.GetSection("Logging"));
+                loggingConfig.AddDebug();
```
hforeste: Can you add this inside of the if condition below so that we only do this when in development mode.

PR #426, opened 2020-02-25
Remove event log provider for public azure

cindylovescoding:
We got some logging file locked exceptions because it's a shared resource. Remove this for public azure

### DIFF
--- a/src/Diagnostics.RuntimeHost/Startup.cs
+++ b/src/Diagnostics.RuntimeHost/Startup.cs
@@ -194,10 +194,13 @@ namespace Diagnostics.RuntimeHost
                 loggingConfig.ClearProviders();
                 loggingConfig.AddConfiguration(Configuration.GetSection("Logging"));
                 loggingConfig.AddDebug();
-                loggingConfig.AddEventSourceLogger();
-                loggingConfig.AddEventLog();
-                loggingConfig.AddAzureWebAppDiagnostics();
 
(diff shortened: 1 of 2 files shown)


cindylovescoding:
Hi @hforeste , I am using this for comparison
string.Compare(Configuration.GetValue<string>("CloudDomain"), DataProviderConstants.AzureCloud, true)

src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -194,10 +194,13 @@ public void ConfigureServices(IServiceCollection services)
                 loggingConfig.ClearProviders();
                 loggingConfig.AddConfiguration(Configuration.GetSection("Logging"));
                 loggingConfig.AddDebug();
-                loggingConfig.AddEventSourceLogger();
```
puneetg1983: Do not remove AddEventSourceLogger() call. This is not event viewer logging. This is [logging for ETW events](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-3.1#event-source-provider). If we remove this, we will not get ETW emitted by ASP.NET Core. You can safely remove the AddEventLog() call which logs by default to EventLog.

Once we upgrade to .net 3.0 or above, we will need ETW to diagnose production issues via Profilers so lets keep the AddEventSourceLogger method.

PR #427, opened 2020-02-28
Enable partner teams to invoke runtimehost directly

rekhaswaminathan:
- Custom URL rewrite module that checks for /api/invoke path and rewrites the URL based on headers. (Currently in DiagRole there is a controller that has this route and forwards the request to runtimehost dotnet process ref:https://msazure.visualstudio.com/One/_git/AAPT-Antares-Websites?path=%2Fsrc%2FHosting%2FAzure%2FGeoRegionService%2FDiagnosticRole%2FControllers%2FDiagnosticResourceController.cs&version=GBdev) 

- Tested with few non Microsoft.Web RPs such as AKS, APIM, LogicApps, Microsoft.AppPlatform

- This will also need an update in KV to whitelist Partner team's ID in AllowedAppIds that runtimehost reads from for auth to work.


src/Diagnostics.RuntimeHost/Utilities/RewriteDiagnosticResource.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        public void ApplyRule(RewriteContext context)
+        {
+            var request = context.HttpContext.Request;
+            // Partner teams using Diagnose and Solve, make the request to api/invoke from their RP.
+            if (!request.Path.Value.Equals(UriElements.PassThroughAPIRoute, StringComparison.OrdinalIgnoreCase))
+            {
+                context.Result = RuleResult.ContinueRules;
+                return;
+            } else
```
hforeste: Put else on new line

src/Diagnostics.RuntimeHost/Utilities/RewriteDiagnosticResource.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                // Check for required headers and send 400 if not present.
+                if (!request.Headers.TryGetValue(HeaderConstants.ApiPathHeader, out var apiPaths) || !apiPaths.Any() || string.IsNullOrWhiteSpace(apiPaths.FirstOrDefault()))
+                {
+                    var response = context.HttpContext.Response;
+                    response.StatusCode = this.StatusCode;
+                    response.ContentType = "text/plain";
+                    response.WriteAsync($"Missing {HeaderConstants.ApiPathHeader} header").Wait();
```
hforeste: `Wait` would be blocking the request thread. Maybe instead you can write text to memory stream and assign that to `response.Body`

src/Diagnostics.RuntimeHost/Utilities/RewriteDiagnosticResource.cs
```diff
@@ -0,0 +1,54 @@
...
+                    context.Result = RuleResult.EndResponse; // Send response and do not continue the request.
+                    return;
+                }
+
+                if (!request.Headers.TryGetValue(HeaderConstants.ApiVerbHeader, out var apiVerbs) || !apiVerbs.Any() || string.IsNullOrWhiteSpace(apiVerbs.FirstOrDefault()))
+                {
+                    var response = context.HttpContext.Response;
+                    response.StatusCode = this.StatusCode;
+                    response.ContentType = "text/plain";
+                    response.WriteAsync($"Missing {HeaderConstants.ApiVerbHeader} header").Wait();
```
hforeste: Same idea here as above

PR #429, opened 2020-03-02
remove thumbprint from generic mdm data provider configuration

Sneezry:


src/Diagnostics.DataProviders/DataProviderConfigurations/GenericMdmDataProviderConfiguration.cs
```diff
@@ -15,11 +15,6 @@ public class GenericMdmDataProviderConfiguration : DataProviderConfigurationBase
         /// </summary>
         public string Endpoint { get; set; }
```
nmallick1: You actually do not need to even ask the detector author for a certificate name in the config. The system loads a default MDM cert by name. Just initialize the cert name with the same subject name that you see in the config.

In the detector code, the author can do a if(cxt.IsInternalCall) and either pass the subject name (for external) or not (for internal)

Sneezry: Because `CertificateName` is required to inherit `IMdmDataProviderConfiguration`, I create a new class `GenericMdmDataProviderConfigurationWrapper` to inherit `IMdmDataProviderConfiguration`, so that the config can be passed to `MdmLogDecorator`. No idea if I do things in the right way. Any suggestion?

PR #439, opened 2020-03-09
Ability to add datatable in AppInsightsOperationContext

puneetg1983:


src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/ApplicationInsight.cs
```diff
@@ -37,13 +37,18 @@ public class AppInsightsOperationContext
         /// <param name="renderingProperties">The rendering properties to render the application insight query result</param>
         /// <returns>AppInsightsOperationContext Object</returns>
         /// <example>
-        public AppInsightsOperationContext(string title, string description, string query, BladeInfo portalBladeInfo, Rendering renderingProperties)
+        public AppInsightsOperationContext(string title, string description, string query, BladeInfo portalBladeInfo, Rendering renderingProperties, DataTable dataTable = null)
```
ShekharGupta1988: This will break pre-compiled detectors

puneetg1983: Thanks, I updated it. good catch!

cindylovescoding: Good point I forgot this..

PR #443, opened 2020-03-10
Enable ETW events for runtime logs

michimune:
This PR adds ETW events ranging 5000 to 5003 for runtime event logs.

It also simplifies the runtime exception stack trace shown as a runtime exception insight

### DIFF
--- a/src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
+++ b/src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
@@ -596,5 +596,113 @@ namespace Diagnostics.Logger
         }
 
         #endregion Internal AI API Events (ID Range : 4000 - 4199)
+
+        #region Runtime Events (ID Range: 5000 - 5199)
+
+        /// <summary>
+        /// Log runtime host message (Error/Critical).
+        /// </summary>
+        /// <param name="RequestId">Request id.</param>
+        /// <param name="Source">The source.</param>
+        /// <param name="SubscriptionId">Subscription id.</param>
+        /// <param name="ResourceGroup">Resource group.</param>
+        /// <param name="Resource">The resource.</param>
+        /// <param name="ExceptionType">Exception type.</param>
+        /// <param name="ExceptionDetails">Exception details.</param>
+        /// <param name="Message">The message.</param>
+        [Event(5000, Level = EventLevel.Error, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogRuntimeMessage)]
+        public void LogRuntimeLogError(string RequestId, string Source, string SubscriptionId, string ResourceGroup, string Resource, string ExceptionType, string ExceptionDetails, string Message)
+        {
+            WriteDiagnosticsEvent(
+                5000,
+                RequestId,
+                Source,
+                SubscriptionId,
+                ResourceGroup,
+                Resource,
+                ExceptionType,
+                ExceptionDetails,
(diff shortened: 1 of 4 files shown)


src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -596,5 +596,113 @@ public void LogInternalAPITrainingSummary(string RequestId, string TrainingId, s
...
+        /// </summary>
+        /// <param name="RequestId">Request id.</param>
+        /// <param name="Source">The source.</param>
+        /// <param name="SubscriptionId">Subscription id.</param>
+        /// <param name="ResourceGroup">Resource group.</param>
+        /// <param name="Resource">The resource.</param>
+        /// <param name="ExceptionType">Exception type.</param>
+        /// <param name="ExceptionDetails">Exception details.</param>
+        /// <param name="Message">The message.</param>
+        [Event(5003, Level = EventLevel.Warning, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogRuntimeMessage)]
```
puneetg1983: Any specific reason you have the Level set to Warning for a Trace Message? If it is Trace, the level should be Verbose or Info.

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -596,5 +596,113 @@ public void LogInternalAPITrainingSummary(string RequestId, string TrainingId, s
...
+        /// <param name="RequestId">Request id.</param>
+        /// <param name="Source">The source.</param>
+        /// <param name="SubscriptionId">Subscription id.</param>
+        /// <param name="ResourceGroup">Resource group.</param>
+        /// <param name="Resource">The resource.</param>
+        /// <param name="ExceptionType">Exception type.</param>
+        /// <param name="ExceptionDetails">Exception details.</param>
+        /// <param name="Message">The message.</param>
+        [Event(5000, Level = EventLevel.Error, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogRuntimeMessage)]
+        public void LogRuntimeLogError(string RequestId, string Source, string SubscriptionId, string ResourceGroup, string Resource, string ExceptionType, string ExceptionDetails, string Message)
```
puneetg1983: My guess is that this tracing will be used widely by detector code. Do they have access to RequestId? Would it possible to add a DetectorId column to the table so that we can easily filter out events from Detectors specifically?

puneetg1983: Ignore my comment. I see the Source is DetectorId and I see how you are passing RequestId so cool!

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
michimune: Good catch! It was a mistake.

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
michimune: No problem!

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
michimune: Fix pushed.

PR #446, opened 2020-03-13
Move downloading Github files to GithubWatcher

rekhaswaminathan:
Having github client download files in base class can potentially cause race conditions between Gist and Detector SourceWatcher. Moved the github client download operation to GithubWatcher and specific SourceWatchers - Gist, Detector, Kusto will only read from file system and update their caches.

### DIFF
--- a/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/GitHubWatcher.cs
+++ b/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/GitHubWatcher.cs
@@ -303,20 +303,46 @@ namespace Diagnostics.RuntimeHost.Services.SourceWatcher
                 .Where(x => !string.IsNullOrWhiteSpace(x.Download_url) 
                 && !string.IsNullOrWhiteSpace(x.Name.Split(new char[] { '.' }).LastOrDefault()));
 
-            if (selectedGithubFiles.Any())
+            var assemblyName = Guid.NewGuid().ToString();
+            var expectedFiles = new string[] { "csx", "package.json", "metadata.json" , "kustoClusterMappings"};
+
+            // Identify files to be downloaded
+            if (!selectedGithubFiles.Any(x => expectedFiles.Any(y => x.Name.Contains(y, StringComparison.CurrentCultureIgnoreCase))))
             {
-                foreach (IGithubWorker githubWorker in GithubWorkers.Values)
+                return;
+            }
+
+            foreach (GithubEntry githubFile in selectedGithubFiles)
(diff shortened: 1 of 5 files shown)


src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/GitHubWatcher.cs
```diff
@@ -303,20 +303,46 @@ private async Task DownloadContentAndUpdateCache(GithubEntry parentGithubEntry,
...
 
-            if (selectedGithubFiles.Any())
+            var assemblyName = Guid.NewGuid().ToString();
+            var expectedFiles = new string[] { "csx", "package.json", "metadata.json" , "kustoClusterMappings"};
+
+            // Identify files to be downloaded
+            if (!selectedGithubFiles.Any(x => expectedFiles.Any(y => x.Name.Contains(y, StringComparison.CurrentCultureIgnoreCase))))
+            {
+                return;
+            }
```
hforeste: Is this necessary anymore? I had this logic in `GitHubScriptWorkerBase` class so that it would only download csx files and dlls but not the kusto mappings file but now that the downloading is singly handled in `GithubWatcher` class it should be okay to remove this. I think the only logic necessary is to make sure it doesn't download search files which is what it is doing in the first couple lines of this method.

hforeste:
Rekha one other change we can do. Now that `GithubWatcher` class is singly doing all the downloading, we don't need to pass the Github client to each individual source watcher worker class. https://github.com/Azure/Azure-AppServices-Diagnostics/pull/446/files#diff-03be911f54eeba8c1cf2de18b5fc8229L74-L76

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/GitHubWatcher.cs
hforeste: Unless your motive is to only download these set files

rekhaswaminathan: I still want this logic to be here just to make sure we are downloading the files we need...

rekhaswaminathan:
> Rekha one other change we can do. Now that `GithubWatcher` class is singly doing all the downloading, we don't need to pass the Github client to each individual source watcher worker class. https://github.com/Azure/Azure-AppServices-Diagnostics/pull/446/files#diff-03be911f54eeba8c1cf2de18b5fc8229L74-L76

Good point. I'll make this change

PR #448, opened 2020-03-16
Removing start up of more than one source watcher to avoid race condi…

rekhaswaminathan:
We can see from logs that there are 2 GithubWatchers running which is leading to race conditions in terms of downloading and IO operations


src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -189,7 +189,7 @@ public void ConfigureServices(IServiceCollection services)
             {
                 services.AddSingleton<ISearchService, SearchServiceDisabled>();
             }
-            servicesProvider.GetService<ISourceWatcherService>();
```
hforeste: When was this brought into the codebase?

PR #451, opened 2020-03-21
Training config for linux and query term sanity check

SharmaAjay19:


src/Diagnostics.AIProjects/SearchAPI/SearchModule/resourceConfig.json
```diff
@@ -2,16 +2,20 @@
...
       },
       "ApiApp": {
-        "Windows": "14748"
+        "Windows": "14748",
+        "Linux": "16170"
       },
       "GatewayApp": {
-        "Windows": "14748"
+        "Windows": "14748",
+        "Linux": "16170"
```
ShekharGupta1988: Hey Ajay, i think there are two PesId dor linux. One for web app linux and other for containers

SharmaAjay19: Okay will put both.

PR #457, opened 2020-03-30
ARM template for storage and create table

rekhaswaminathan:
The following arm template creates storage account and a blob container inside it. 
Script can be used to create table since table creation through ARM template is not supported yet.

### DIFF
--- a/infrastructure/ArmTemplates/storage-deploy.json
+++ b/infrastructure/ArmTemplates/storage-deploy.json
@@ -1,0 +1,85 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "storageAccount_name": {
+            "defaultValue": "diagsourcewatcher",
+            "type": "string"
+        },
+        "location": {
+            "type": "string"
+        },
+        "container_Name": {
+            "type": "string"
+        }
+    },
+    "variables": {},
+    "resources": [
+        {
+            "type": "Microsoft.Storage/storageAccounts",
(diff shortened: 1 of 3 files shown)


infrastructure/ArmTemplates/storage-deploy.json
```diff
@@ -0,0 +1,85 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "storageAccount_name": {
+            "defaultValue": "diagsourcewatcher",
+            "type": "string"
+        },
+        "location": {
+            "type": "string"
```
hforeste: Create default with `resourceGroup().location`. This will default to location of the resource group. https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-resource#resourcegroup

infrastructure/Scripts/storage-postdeployment.ps1
```diff
@@ -0,0 +1,15 @@
+param ($resourceGroup, $storageAccount)
+
```
hforeste: Doesn't the powershell script need to setup the azure environment with Get-AzAccount or this just automatically works because the deployment pipeline automatically loads the powershell environment?

rekhaswaminathan: I referred to some examples [here ](https://github.com/Azure/azure-docs-powershell-samples/tree/master/storage)and don't find the `Connect-AzAccount` at the beginning of each script. I guess it operates within the context of the powershell environment (Portal or deployment pipeline)

PR #461, opened 2020-04-01
Master2

hforeste:
- Use some ARM functions to simplify DeployKeyVault.json
- Created separate ARM template for deploying the web app
- Put KeyVault behind a flag for compilerhost as there is no "secrets" for compilerhost
     - TODO we need to figure out a way to only keep secrets in KeyVaults instead of using it as a central store for configuration as this has brought tons of trouble in other clouds
- Let environment variables/app settings have the last say for data provider configurations
- Stop using the built-in extensions.diagnostics HealthCheck library and use our own custom model. Main reason is we had limitation with the built-in library
     - .NET Core implementation of HealthCheckResult is a struct. There is not as much flexibility with structs as there are with classes.
     - .NET Core implementation of had this concept of HealthCheckContext which we didn't need and added unnecessary bloated code.
- Add dependency check for MDM
- Add dependency check for SourceWatcher
- Run health check only for data providers that are enabled
- Fix dependency checker failure bug where whole API fails if one of the dependencies fail.
     - Return health check data even in the event that 1+ data sources fail
- Return 500 in the event the whole API fails and return exception message in API response
- Fix content type bug. Partners would call into us using api/invoke without a ContentType header and they would get 400. It looks like they don't run into this with DiagnosticRole because the w3wp pass through automatically adds ContentType before sending request to the .NET Core process
- Fix rewrite rule query string bug. The query string parameters for changeanalysis were not flowing through. To fix, I simply reused the ApplyRule method of RewriteRule class instead of our own custom logic.


infrastructure/ArmTemplates/DeployWebApp.json
```diff
@@ -0,0 +1,374 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
+  "contentVersion": "1.0.0.0",
```
rekhaswaminathan, line 3: The existing one takes care for setting up geneva logging as well. Can you combine that here?

src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
```diff
@@ -206,5 +210,53 @@ public async Task<IEnumerable<DataTable>> GetMultipleTimeSeriesAsync(DateTime st
...
         }
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Reliability", "CA2007:Consider calling ConfigureAwait on the awaited task", Justification = "<Pending>")]
+        public override async Task<HealthCheckResult> CheckHealthAsync(CancellationToken cancellationToken = default(CancellationToken))
+        {
+            HealthCheckResult result;
+            if (_configuration.HealthCheckInputs != null && _configuration.HealthCheckInputs.Any())
+            {
+                //namespace=Microsoft/Web/WebApps;sampling=average;metric=PrivateBytes
+                _configuration.HealthCheckInputs.TryGetValue("namespace", out string mdmNamespace);
```
rekhaswaminathan, line 221: Is the `HealthCheckInputs` defined in config? I dont see it.

src/Diagnostics.RuntimeHost/Utilities/RewriteDiagnosticResource.cs
```diff
@@ -54,11 +60,21 @@ public void ApplyRule(RewriteContext context)
...
+                {
+                    request.Headers.Add(contentTypeHeader, new StringValues(contentTypeHeaderValue));
+                }
+                else if (!requestedContentTypes.Any(hv => hv.Equals("application/json")))
+                {
+                    request.Headers.Append(contentTypeHeader, new StringValues(contentTypeHeaderValue));
+                }
+
+                request.Method = apiVerbs.First().ToLower();
+                var rewriteRule = new RewriteRule(UriElements.PassThroughAPIRoute.Substring(1), apiPaths.First().ToLower(), true);
```
rekhaswaminathan, line 76: Thanks for fixing this

infrastructure/ArmTemplates/DeployWebApp.json
hforeste: I intentionally created this to remove the geneva logging in environments where we don't have geneva set up. I think it should be its own separate ARM template.

src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
hforeste: Yes it is in the DataProviderConfigurationBase class. https://github.com/Azure/Azure-AppServices-Diagnostics/blob/master2/src/Diagnostics.DataProviders/DataProviderConfigurations/DataProviderConfigurationBase.cs#L23-L48

hforeste: My bad, I think your talking about the app settings right?

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -684,5 +685,43 @@ private string ParseGeoMasterName(string geomasterHostName)
...
+                _configuration.HealthCheckInputs.TryGetValue("siteName", out string siteName);
+
+                var inputs = new string[] { subscription, resourceGroupName, siteName };
+
+                if (inputs.All(s => !string.IsNullOrWhiteSpace(s)))
+                {
+                    Exception geomasterException = null;
+                    try
+                    {
+                        var response = await this.GetAppSettings(subscription, resourceGroupName, siteName);
```
ShekharGupta1988, line 706: I feel its more necessary now to secure the Health check endpoint as the Stakes are high.
We can potentially cause high geomaster CPU if someone Ddos the health check endpoint

ShekharGupta1988: Also, The Appsettings call will go all the way to Stamp Database. So its very expensive.
for Health Check, maybe you want to consider a less expensive call which only involves geomaster.
I am little bit worried though about this call happening for every Health Check API request.

Another Design Model could be that you have a Service which pings these dependencies once every 5 mins (or X mins), logs the result in kusto and maintain the last Status Check result.
The API simply returns what was the last check result along with the timestamp.

src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
rekhaswaminathan: Sounds good. thanks

src/Diagnostics.RuntimeHost/Services/HealthCheckService/HealthCheckService.cs
```diff
@@ -61,24 +69,61 @@ public async Task RunHealthCheck()
...
 
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Reliability", "CA2007:Consider calling ConfigureAwait on the awaited task", Justification = "<Pending>")]
         public async Task<IEnumerable<HealthCheckResult>> RunDependencyCheck(DataProviders.DataProviders dataProviders)
         {
-            var healthCheckResultsTask = new List<Task<HealthCheckResult>>();
+            if (cache.TryGetValue(DEPENDENCYCHECK_CACHE_KEY, out object cachedDependencyChecks))
+            {
+                return (IEnumerable<HealthCheckResult>)cachedDependencyChecks;
+            }
+
```
ShekharGupta1988, line 79: nitpick : After the if block, How about adding a log here to know how frequent is the cache miss and the dependency call is actually running?

PR #464, opened 2020-04-03
Remove logic to run all detector when support topic is null or not found

SharmaAjay19:


src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -404,6 +404,11 @@ protected async Task<IActionResult> GetInsights(TResource resource, string pesId
             supportTopicId = ParseCorrectSupportTopicId(supportTopicId);
             var supportTopic = new SupportTopic() { Id = supportTopicId, PesId = pesId };
             RuntimeContext<TResource> cxt = PrepareContext(resource, startTimeUtc, endTimeUtc, true, supportTopic, null, postBody);
+            if (supportTopicId == null)
+            {
+                DiagnosticsETWProvider.Instance.LogRuntimeHostHandledException(cxt.OperationContext.RequestId, "GetInsights", cxt.OperationContext.Resource.SubscriptionId,
+                    cxt.OperationContext.Resource.ResourceGroup, cxt.OperationContext.Resource.Name, "ASCSupportTopicIdNull", "Support Topic Id is null or there is no mapping for the Support topic path provided.");
```
ShekharGupta1988: should we continue the API or return

SharmaAjay19: We still have to return the default insight, so I suppose letting it run is a good way.

PR #465, opened 2020-04-03
Adding DataSources Tab info for AI and MDM

puneetg1983:
With this PR, we are adding DataSource information for MDM provider and Application insights to Data Sources tab.

- For MDM, only a link to JARVIS portal with the right dashboard and timestamp will be provided (Unfortunately, for MDM, I cannot test my logic end to end because we dont have a private setup for MDM so there is a chance we have the URL which is slightly incorrect and might need another PR to get perfect) 
- For AppInsights, only the QueryText makes sense because we dont have the URL for customer's AI resource's Analytics blade. We are just pointing to documentation for Analytics for now.

After this gets deployed, slight changes would be required to the UI code because currently every link is coming as **Run In Kusto Web Explorer**

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
@@ -1,15 +1,24 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Linq;
 using System.Threading.Tasks;
 using Diagnostics.DataProviders.DataProviderConfigurations;
 using Diagnostics.DataProviders.Interfaces;
 using Diagnostics.Logger;
 using Diagnostics.ModelsAndUtils.Models;
 using Microsoft.AspNetCore.Http;
+using Newtonsoft.Json;
 
 namespace Diagnostics.DataProviders
 {
+    public class MdmQueryParameters
+    {
+        public string query { get; set; }
+        public string key { get; set; } = "value";
+        public string replacement { get; set; }
+    }
+
     /// <summary>
     /// Mdm data provider
     /// </summary>
@@ -29,6 +38,10 @@ namespace Diagnostics.DataProviders
         {
             _configuration = configuration;
             _mdmClient = MdmClientFactory.GetMdmClient(configuration, requestId);
+            Metadata = new DataProviderMetadata
+            {
+                ProviderName = "MDM"
+            };
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
```diff
@@ -206,5 +224,71 @@ public async Task<IEnumerable<DataTable>> GetMultipleTimeSeriesAsync(DateTime st
...
+                                                            x.Value.GetType() == typeof(DataProviderMetadataQuery) &&
+                                                            x.Value.CastTo<DataProviderMetadataQuery>().Url.Equals(finalUrl, StringComparison.OrdinalIgnoreCase));
+                if (!urlExists)
+                {
+                    Metadata.PropertyBag.Add(new KeyValuePair<string, object>("Query", new DataProviderMetadataQuery() { Text = finalUrl, Url = finalUrl }));
+                }
+            }
+        }
+
+        private double GetDateTimeInEpochMilliseconds(DateTime dateTimeUtc)
```
hforeste, line 289: You can use `DateTimeOffset` method of `ToUnixTimeMilliseconds`. https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset.tounixtimemilliseconds?view=netframework-4.8

src/Diagnostics.ModelsAndUtils/Models/DataProviderMetadataQuery.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Diagnostics.ModelsAndUtils.Models
+{
+    public class DataProviderMetadataQuery
+    {
+        public string Text;
+        public string Url;
+        public string OperationName;
+    }
+}
```
hforeste, line 9: Should you use this class in KustoDataProvider too? https://github.com/Azure/Azure-AppServices-Diagnostics/blob/master2/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs#L113

src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
puneetg1983: Sweet! I didn't know this existed

src/Diagnostics.ModelsAndUtils/Models/DataProviderMetadataQuery.cs
puneetg1983: I wanted to but unfortunately that will break the complied detectors today that are using the `GetKustoQueryAsync `method.

PR #469, opened 2020-04-05
Checking for Func instead of Interface

puneetg1983:
Since MDM is a field of type `Func<MdmDataSource, IMdmDataProvider>` in DataProviders class, our default logic of casting the field to Interface is not working and we are not getting the dataProvider metadata. 

This change only handles for fields which are of type` Func<Enum, Interface>` so the below logic will not work for `Func<GenericMdmDataProviderConfiguration, IMdmDataProvider>`

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/AppInsightsDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/AppInsightsDataProvider.cs
@@ -24,6 +24,10 @@ namespace Diagnostics.DataProviders
         {
             _configuration = configuration;
             _appInsightsClient = new AppInsightsClient(_configuration);
+            Metadata = new DataProviderMetadata
+            {
+                ProviderName = "AppInsights"
+            };
         }
 
         public Task<bool> SetAppInsightsKey(string appId, string apiKey)
@@ -110,10 +114,10 @@ namespace Diagnostics.DataProviders
 
         public async Task<DataTable> ExecuteAppInsightsQuery(string query)
         {
-            return await ExecuteAppInsightsQuery(query, "AppInsightsQuery");
+            return await ExecuteAppInsightsQuery(query, "");
         }
 
-        private void AddQueryInformationToMetadata(string query, string operationName = "AppInsightsQuery")
+        private void AddQueryInformationToMetadata(string query, string operationName = "")
         {
             bool queryExists = Metadata.PropertyBag.Any(x => x.Key == "Query" &&
                                                             x.Value.GetType() == typeof(DataProviderMetadataQuery) &&
--- a/src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
@@ -274,7 +274,7 @@ namespace Diagnostics.DataProviders
 
                 if (dashboard == "WAWS Shoebox/Web Apps/Per Resource")
                 {
-                    AddRemainingParametersForWebApps(queryParameters, resourceId);
+                    AddRemainingParametersForWebApps(queryParameters, resourceId, seriesResolutionInMinutes);
                 }
 
             }
@@ -297,7 +297,7 @@ namespace Diagnostics.DataProviders
             }
         }
 
-        private void AddRemainingParametersForWebApps(List<MdmQueryParameters> queryParameters, string resourceId)
+        private void AddRemainingParametersForWebApps(List<MdmQueryParameters> queryParameters, string resourceId, int seriesResolutionInMinutes)
         {
             var hostArray = resourceId.Split(".");
             if (hostArray.Length < 3)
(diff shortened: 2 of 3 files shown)


src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -865,8 +865,7 @@ private List<DataProviderMetadata> GetDataProvidersMetadata(DataProviders.DataPr
             {
                 if (dataProvider.FieldType.IsInterface)
                 {
-                    var metadataProvider = dataProvider.GetValue(dataProviders) as IMetadataProvider;
-                    if (metadataProvider != null)
+                    if (dataProvider.GetValue(dataProviders) is IMetadataProvider metadataProvider)
```
hforeste, line 868: > Since MDM is a field of type Func<MdmDataSource, IMdmDataProvider> in DataProviders class, our default logic of casting the field to Interface is not working and we are not getting the dataProvider metadata.

Are you saying this never used to work?

src/Diagnostics.DataProviders/DataProviders/AppInsightsDataProvider.cs
```diff
@@ -110,10 +114,10 @@ public async Task<DataTable> ExecuteAppInsightsQuery(string query, string operat
 
         public async Task<DataTable> ExecuteAppInsightsQuery(string query)
```
hforeste, line 115: If there is no purpose for this method then maybe just have the above `ExecuteAppInsightsQuery(string query, string operationName = null)`. Not sure if that'll break detectors..

puneetg1983: Changing the signature to  `ExecuteAppInsightsQuery(string query, string operationName = null)` breaks existing detectors hence we always added new functions rather than modifying the signature of the existing ones.

src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
```diff
@@ -266,6 +271,12 @@ public async Task<IEnumerable<DataTable>> GetMultipleTimeSeriesAsync(DateTime st
                         queryParameters.Add(mdmParameter);
                     }
                 }
+
+                if (dashboard == "WAWS Shoebox/Web Apps/Per Resource")
+                {
+                    AddRemainingParametersForWebApps(queryParameters, resourceId, seriesResolutionInMinutes);
+                }
```
hforeste, line 278: Now that we have this GenericMdm logic, I can see multiple partners who want to also benefit by adding additional information to the metadata. If we continue with this style of logic then I can see this list growing in the future. Not that its an urgent problem but we should probably think of a dynamic way of doing this.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
hforeste, line 868: What changes between `as` and `is`?

puneetg1983: This is visual studio default refactoring suggestion :-)

Instead of two lines, it is suggesting to combine null check and typecasting to IMetadataProvider in one line

src/Diagnostics.DataProviders/DataProviders/MdmDataProvider.cs
puneetg1983: Agreed. For GenericMdm logic, we need to come up with a better way of handling this. Currently this change is not impacting the GenericMdm provider at all.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -875,6 +874,47 @@ private List<DataProviderMetadata> GetDataProvidersMetadata(DataProviders.DataPr
                         }
                     }
                 }
+                else if (dataProvider.FieldType.UnderlyingSystemType.Name.StartsWith("Func`2", StringComparison.OrdinalIgnoreCase))
```
hforeste, line 877: Partners will be using MdmGeneric. Can you find a way to also have this benefit them too?

puneetg1983: We need to come up with a better way of doing this for partners. Doing it the reflection way is trickier and dirtier as the MdmGeneric expects GenericMdmDataProviderConfiguration parameter and I cannot find an easy way via reflection to get that parameter. We need to think of that provider differently. Maybe as a separate PR altogether.

hforeste: Okay, lets just make sure we create a WI for this so this doesn't slip through the cracks

puneetg1983: Opened 1215 to track this.

PR #471, opened 2020-04-06
Scheduled func app to back fill data in storage

rekhaswaminathan:
- Function app is timer triggered to run once per day
- When running func app locally, config is loaded from `configuration.json`, when deployed in Azure, it will take environment variables
- Goes through each directory in github repo, loads .dll to blob container and detector metadata into storage table. (detectors and gists) 
- Table is a schema-less, we can add any columns we want in the future. Columns stored right now are defined in `DetectorEntity.cs`


src/SourceWatcherFuncApp/.gitignore
```diff
@@ -0,0 +1,265 @@
+## Ignore Visual Studio temporary files, build results, and
+## files generated by popular Visual Studio add-ons.
+
+# Azure Functions localsettings file
+local.settings.json
+configuration.json
```
hforeste, line 6: Maybe you can just add the the function specific things that need to be ignored in the .gitignore file in the root directory.

src/SourceWatcherFuncApp/DiagSourceWatcher.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        [FunctionName("DiagSourceWatcher")]
+        public async Task Run([TimerTrigger("0 0 0 * * *")]TimerInfo timerInfo, ILogger log)
+        {
+            log.LogInformation($"C# Timer trigger function executed at: {DateTime.Now}");
+
+            log.LogInformation($"Loading directories from github branch  {config["Github:Branch"]}");
+
+            var githubDirectories = await githubService.DownloadGithubDirectories(config["Github:Branch"]);
```
hforeste, line 38: Check `Github:Branch` for null or pass `branchdownloadUrl`

src/SourceWatcherFuncApp/DiagSourceWatcher.cs
```diff
@@ -0,0 +1,87 @@
...
+                    var existingDetectorEntity = await storageService.GetEntityFromTable(configFileData.PartitionKey, configFileData.RowKey);
+
+                    // If there is no entry in table or github last modifed date has been changed, upload to blob
+                    if (existingDetectorEntity == null || existingDetectorEntity.GithubLastModified != configFileData.GithubLastModified)
+                    {                        
+                        var assemblyLastModified = await githubService.GetCommitDate(assemblyFile.Path);
+                        blobService.LoadBlobToContainer(assemblyFile.Path, assemblyData);
+                    }
+
+                    var result = await storageService.LoadDataToTable(configFileData);
```
hforeste, line 75: You only need to do this within the above if body right?

rekhaswaminathan: Nope. This should be done in case of any of changes to the detector entity. For example, if `supportTopicList `is updated or `InternalOnly `flag is updated in the publish flow. Instead of comparing and checking which fields of `DetectorEntity` has changed and updating those, I am just passing the model so that table storage can handle that better

src/SourceWatcherFuncApp/DiagSourceWatcher.cs
rekhaswaminathan: `Github:Branch` is checked for null within the method

src/SourceWatcherFuncApp/DiagSourceWatcher.cs
```diff
@@ -0,0 +1,87 @@
...
+
+                if (assemblyFile != null && scriptFile != null && configFile != null)
+                {
+
+                    log.LogInformation($"Getting content for Assembly file : {assemblyFile.Path}");
+                    var assemblyData = await githubService.GetFileContentStream(assemblyFile.Download_url);
+
+                    log.LogInformation("Reading detector metadata");
+                    var configFileData = await githubService.GetFileContentByType<DetectorEntity>(configFile.Download_url);
+                    var scriptData = await githubService.GetFileContentByType<string>(scriptFile.Download_url);
```
ShekharGupta1988, line 59: do you need to download csx file?

src/SourceWatcherFuncApp/Services/IGithubService.cs
```diff
@@ -0,0 +1,147 @@
...
+            octokitClient = new GitHubClient(new Octokit.ProductHeaderValue(githubUserName))
+            {
+                Credentials = new Credentials(AccessToken)
+            };
+
+            // Initialize HttpClient
+
+            httpClient.DefaultRequestHeaders.Add("Authorization", $@"token {AccessToken}");
+            httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+            httpClient.DefaultRequestHeaders.Add("User-Agent", "applensv3");
```
ShekharGupta1988, line 56: nit: you may want to have a different user-agent?

src/SourceWatcherFuncApp/DiagSourceWatcher.cs
rekhaswaminathan: Yes its needed in the `EntityInvoker` part of `EntityMetadata`

ShekharGupta1988: Yes, but since we are not compiling the scripts, you can leave the script part empty and save 1/3rd of the download calls (for csx files) every day.

rekhaswaminathan: I see. Sure I will make that change

PR #473, opened 2020-04-07
Fix gist publishing issue

michimune:
It is a regression caused by my change 06348ee2cdd84cdc4d3c11a34e3eaf71c58534c8.

### DIFF
--- a/src/Diagnostics.Scripts/EntityInvoker.cs
+++ b/src/Diagnostics.Scripts/EntityInvoker.cs
@@ -178,8 +178,9 @@ namespace Diagnostics.Scripts
 
             if (EntityMetadata.Type == EntityType.Gist)
             {
-                // This is a gist. We cannot invoke it. Let's return an empty response object.
-                return new Response();
+                // This is a gist. We cannot invoke it. Let's return the Response object that was given.
+                var response = parameters.FirstOrDefault(arg => arg is Response) ?? new Response();
+                return response;
             }
 
             var methodInfo = memberInfo as MethodInfo;


src/Diagnostics.Scripts/EntityInvoker.cs
```diff
@@ -178,8 +178,9 @@ public async Task<object> Invoke(object[] parameters)
 
             if (EntityMetadata.Type == EntityType.Gist)
             {
-                // This is a gist. We cannot invoke it. Let's return an empty response object.
-                return new Response();
+                // This is a gist. We cannot invoke it. Let's return the Response object that was given.
+                var response = parameters.FirstOrDefault(arg => arg is Response) ?? new Response();
```
rekhaswaminathan, line 182: Do you mind explaining how this is fixing the issue?

michimune: When you publish, it comes to this code path (DiagnosticsControllerBase.cs):

protected async Task<IActionResult> PublishPackage(Package pkg)
        {
            if (pkg == null || string.IsNullOrWhiteSpace(pkg.Id))
            {
                return BadRequest();
            }

So Package.Id must be a non-null value. The pkg is created from a Response object, which is returned by Invoke(). My previous change returned an empty Response object whose metadata wasn't initialized. This caused the Id property to be null and the issue occurred. Moreover, not only id but also author and attributes were empty.

My fix is to return the Response object created by the caller, which contains all the necessary metadata.

PR #483, opened 2020-04-10
Sourcewatcher set up function app updates

rekhaswaminathan:
- Store resource level info - Resource Provider ,Resource Type, PlatForm, AppType, Stack
- Fixing snat problem
- Log exception if operation fails.


src/SourceWatcherFuncApp/DiagSourceWatcher.cs
```diff
@@ -7,79 +7,82 @@
...
-
-                    var result = await storageService.LoadDataToTable(configFileData);
-
-                }
-                else
+                } catch (Exception ex)
                 {
-                    log.LogInformation($"One or more files were not found in {githubdir.Name}, skipped storage entry");
-                }
+                    log.LogInformation($"Exception occured while processing {githubdir.Name}: {ex.ToString()} ");
```
hforeste, line 81: Use the error level event here

src/SourceWatcherFuncApp/Entities/DetectorEntity.cs
```diff
@@ -65,5 +65,15 @@ public List<string> AnalysisTypes
...
+
+        public string PlatForm { get; set; }
+
+        public string AppType { get; set; }
+
+        public string StackType { get; set; }
+
+        public string HostingEnvironmentType { get; set; }
+
+        public string ResourceProvider { get; set; }
```
hforeste, line 77: I dont think you need `StackType` or `HostingEnvironmentType`. You have plans for these?

src/SourceWatcherFuncApp/Services/IStorageService.cs
```diff
@@ -35,15 +44,52 @@ public TableStorageService(IConfigurationRoot configuration, ILogger<TableStorag
...
+                            var directory = (CloudBlobDirectory)blobItem;
+                            var name = directory.Prefix.Replace("/", "");
+                            existingDetectors.Add(name);
+                        } 
+                    }
+                }
+                return existingDetectors.Contains(currentDetector);
+            } catch(Exception ex)
+            {
+                return false;
```
hforeste, line 72: Did you want to log anything?

src/SourceWatcherFuncApp/Utilities/EntityHelper.cs
```diff
@@ -27,16 +29,90 @@ public static DetectorEntity PrepareEntityForLoad(Stream streamAssemblyData, str
...
+                        detectorPackage.PlatForm = string.Join(",", platformTypesList);
+
+                        StackType stackType = resourceInfo.StackType;
+                        var stackTypesList = Enum.GetValues(typeof(StackType)).Cast<StackType>().Where(s => stackType.HasFlag(s)).Select(x => Enum.GetName(typeof(StackType), x));
+                        detectorPackage.StackType = string.Join(",", stackTypesList);
+
+                        detectorPackage.ResourceProvider = "Microsoft.Web";
+                        detectorPackage.ResourceType = "sites";
+                    } 
+                    else if (invoker.ResourceFilter is ApiManagementServiceFilter)
```
hforeste, line 57: Since this function app is throw-away code you don't have to do this suggestion but if you like one thing you can do to avoid this huge if else body and save time in maintaining this code is reference the Diagnostics.ModelAndUtils project and use my example code below. This code is simply reusing the `IsApplicable` method in the IResource classes to determine if that `IResourceFilter` is applicable. This is similar to what we do today for list detectors call. https://github.com/Azure/Azure-AppServices-Diagnostics/blob/7b4bd51ad8b0c228e2c7a85421e755714c328556/src/Diagnostics.RuntimeHost/Services/CacheService/InvokerCacheService.cs#L49-L59

                var services = Assembly.GetExecutingAssembly().GetTypes().Where(t => t.GetInterfaces().Contains(typeof(IResource)) && t.IsAssignableFrom(typeof(Attribute)));
                foreach (var item in services)
                {
                    IResource resource = (IResource)Activator.CreateInstance(item);
                    if (resource.IsApplicable(invoker.ResourceFilter))
                    {
                        detectorPackage.ResourceProvider = resource.Provider;
                        detectorPackage.ResourceType = resource.ResourceType.ToString();
                    }
                }

PR #484, opened 2020-04-17
SourceWatcher Redesign - Part 1 : ListDetectors, ListGists from AzureStorage

rekhaswaminathan:
- Added feature flag for exposing new system. Disabled by default. Can be enabled by overriding the value in `appsettings.Development.json`
- ListDetectors, ListGists retrieved from storage table based on feature flag 
- Added unit tests for Azure Storage using [Storage Emulator](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-emulator)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,7 +25,8 @@ steps:
 - task: CmdLine@2
   displayName: 'Run Azure Storage Emulator'
   inputs:
-    script: "%ProgramFiles(x86)%\Microsoft SDKs\Azure\Storage Emulator\AzureStorageEmulator.exe" start
+    script: |
+    "%ProgramFiles(x86)%\Microsoft SDKs\Azure\Storage Emulator\AzureStorageEmulator.exe" start
 
 - task: DotNetCoreCLI@2
   inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,6 +21,12 @@ steps:
     projects: '$(solution)'
     configuration: '$(buildConfiguration)'
 
+
+- task: CmdLine@2
+  displayName: 'Run Azure Storage Emulator'
+  inputs:
+    script: "%ProgramFiles(x86)%\Microsoft SDKs\Azure\Storage Emulator\AzureStorageEmulator.exe" start
+
 - task: DotNetCoreCLI@2
   inputs:
     command: 'test'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,13 +21,6 @@ steps:
     projects: '$(solution)'
     configuration: '$(buildConfiguration)'
 
-
-# - task: CmdLine@2
-#   displayName: 'Run Azure Storage Emulator'
-#   inputs:
-#     script: '"%ProgramFiles(x86)%\Microsoft SDKs\Azure\Storage Emulator\AzureStorageEmulator.exe" start'
-
-
 - task: DotNetCoreCLI@2
   inputs:
     command: 'test'
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,8 +25,8 @@ steps:
 - task: CmdLine@2
   displayName: 'Run Azure Storage Emulator'
   inputs:
-    script: |
-    "%ProgramFiles(x86)%\Microsoft SDKs\Azure\Storage Emulator\AzureStorageEmulator.exe" start
+    script: '"%ProgramFiles(x86)%\Microsoft SDKs\Azure\Storage Emulator\AzureStorageEmulator.exe" start'
+
 
 - task: DotNetCoreCLI@2
   inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -22,10 +22,10 @@ steps:
     configuration: '$(buildConfiguration)'
 
 
-- task: CmdLine@2
-  displayName: 'Run Azure Storage Emulator'
-  inputs:
-    script: '"%ProgramFiles(x86)%\Microsoft SDKs\Azure\Storage Emulator\AzureStorageEmulator.exe" start'
+# - task: CmdLine@2
+#   displayName: 'Run Azure Storage Emulator'
+#   inputs:
+#     script: '"%ProgramFiles(x86)%\Microsoft SDKs\Azure\Storage Emulator\AzureStorageEmulator.exe" start'
 
 
 - task: DotNetCoreCLI@2
--- a/src/Diagnostics.ModelsAndUtils/Diagnostics.ModelsAndUtils.csproj
+++ b/src/Diagnostics.ModelsAndUtils/Diagnostics.ModelsAndUtils.csproj
@@ -18,6 +18,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
     <PackageReference Include="System.Collections.Immutable" Version="1.3.1" />
+    <PackageReference Include="WindowsAzure.Storage" Version="9.3.3" />
   </ItemGroup>
   
   <ItemGroup>
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -21,6 +21,7 @@ using Diagnostics.RuntimeHost.Services;
 using Diagnostics.RuntimeHost.Services.CacheService;
 using Diagnostics.RuntimeHost.Services.CacheService.Interfaces;
 using Diagnostics.RuntimeHost.Services.SourceWatcher;
+using Diagnostics.RuntimeHost.Services.StorageService;
 using Diagnostics.RuntimeHost.Utilities;
 using Diagnostics.Scripts;
 using Diagnostics.Scripts.Models;
@@ -50,6 +51,7 @@ namespace Diagnostics.RuntimeHost.Controllers
         protected IRuntimeLoggerProvider _loggerProvider;
 
         private InternalAPIHelper _internalApiHelper;
+        private ITableCacheService tableCacheService;
 
         public DiagnosticControllerBase(IServiceProvider services, IRuntimeContext<TResource> runtimeContext)
         {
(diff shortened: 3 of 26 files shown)


src/Diagnostics.ModelsAndUtils/Diagnostics.ModelsAndUtils.csproj
```diff
@@ -18,6 +18,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="2.2.0" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.2" />
     <PackageReference Include="System.Collections.Immutable" Version="1.3.1" />
+    <PackageReference Include="WindowsAzure.Storage" Version="9.3.3" />
```
puneetg1983, line 21: >WindowsAzure.Storage [](start = 31, length = 20)

This version of Storage SDK is not actively maintained any more. 

> NOTE: As of version 9.4.0, this library has been split into multiple parts and replaced:  See Microsoft.Azure.Storage.Blob, Microsoft.Azure.Storage.File, Microsoft.Azure.Storage.Queue, and Microsoft.Azure.Storage.Common.

You can switch to the newer libraries that are active maintained. The classes and the functions are pretty much the same.

rekhaswaminathan: @puneetg1983  Yes, I tried using the newer library for table operations but I kept running into exceptions:
`Microsoft.Azure.Cosmos.Table.StorageException: Not Implemented
   at Microsoft.Azure.Cosmos.Table.RestExecutor.TableCommand.Executor.ExecuteAsync[T](RESTCommand`1 cmd, IRetryPolicy policy, OperationContext operationContext, CancellationToken token)`

`ErrorMessage:The requested operation is not implemented on the specified resource.`

azure-pipelines.yml
```diff
@@ -21,6 +21,13 @@ steps:
     projects: '$(solution)'
     configuration: '$(buildConfiguration)'
 
+
+- task: CmdLine@2
+  displayName: 'Run Azure Storage Emulator'
+  inputs:
+    script: '"%ProgramFiles(x86)%\Microsoft SDKs\Azure\Storage Emulator\AzureStorageEmulator.exe" start'
```
hforeste, line 28: Did this end up working?

rekhaswaminathan: The .yml file is valid. But seems like the test cases are failing in CI/CD. I am looking into it

hforeste:
Hey Rekha, have you given any thought to creating a `StorageWatcher` similar to how we have a `GitHubWatcher` and a `LocalFileSystemWatcher`?

rekhaswaminathan:
> Hey Rekha, have you given any thought to creating a `StorageWatcher` similar to how we have a `GitHubWatcher` and a `LocalFileSystemWatcher`?

Yes @hforeste, that is going to be next part that I am going to work on. The `StorageWatcher` will act in a similar way like `GithubWatcher` and update the caches when there is a change to blob detected. 

I'd like to do the SourceWatcher redesign in parts of small & complete functionality code, instead of a huge PR that will be difficult to review and also debug in case of regressions. Thats why I created this PR for ListDetectors and ListGists only.

src/Diagnostics.RuntimeHost/Services/CacheService/TableCacheService.cs
```diff
@@ -0,0 +1,99 @@
...
+using Diagnostics.RuntimeHost.Services.StorageService;
+using System.Threading.Tasks;
+using Diagnostics.ModelsAndUtils.Models.Storage;
+using Diagnostics.RuntimeHost.Models;
+using System.Linq;
+using System;
+
+namespace Diagnostics.RuntimeHost.Services.CacheService
+{
+    public class TableCacheService : ITableCacheService
```
ShekharGupta1988, line 14: nit: Looks like this Table Cache service is for DiagEntitites only. Should we rename this service class to be a very specific instead of generic name. (prob something like DiagEntityTableCacheService)

src/Diagnostics.RuntimeHost/Services/CacheService/TableCacheService.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        public async Task<List<DiagEntity>> GetEntityListByType<TResource>(RuntimeContext<TResource> context, string entityType = null)
+          where TResource : IResource
+        {
+            entityType = entityType == null ? "Detector" : entityType;
+
+            var result = new List<DiagEntity>();
+
+            while (!TryGetValue(entityType, out result))
```
ShekharGupta1988, line 57: does this needs to be "if" instead of "while"

ShekharGupta1988: Oh I see what you are doing. One other way to save little bit of computation is update line 59 to say : 
result = await storageService.RetieveEntitiesByPartitionkey(entityType);

src/Diagnostics.RuntimeHost/Services/CacheService/TableCacheService.cs
```diff
@@ -0,0 +1,99 @@
...
+            return result;
+        }
+
+
+        public bool IsStorageAsSourceEnabled()
+        {
+            return storageService.GetStorageFlag();
+        }
+
+        public List<DiagEntity> ApplySearchEngineFiltering(SearchResults searchResults, List<DiagEntity> currentEntities)
```
ShekharGupta1988, line 73: Feel like this method doesnt belong here. Nothing to do with Table cache. no?

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -0,0 +1,142 @@
...
+            }
+
+            CheckTable();
+        }
+
+
+        private void CheckTable()
+        {
+            CloudTable table = tableClient.GetTableReference(tableName);
+            table.CreateIfNotExistsAsync();
```
ShekharGupta1988, line 63: This looks like an awaitable task inside a void method. Two possible issues here:
- This method will run sequentially (Prob not that big impact as this is only called once in Service Ctor)
- Since we are not awaiting the task, there is no guarantee that the table is created when this method exits.

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -0,0 +1,142 @@
...
+        }
+
+        public async Task<List<DiagEntity>> RetieveEntitiesByPartitionkey(string partitionKey = null)
```
ShekharGupta1988, line 66: nitpick : spell mistake :) (Retrieve).
Just looking at the methods generally exposed by SDKs, calling it 'GetEntitiesByPartitionkey' might be simple

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            try
+            {
+                CloudTable table = tableClient.GetTableReference(tableName);
```
ShekharGupta1988, line 70: should we create the Table reference also once in initialization instead of every method call?

src/Diagnostics.RuntimeHost/Services/CacheService/TableCacheService.cs
rekhaswaminathan: The reason I have this as while is, first time we are querying for list of entities, there won't be any in the cache. So it can get the entities from storage and add to cache. When the while loop executes second time,  the condition will be met so it can proceed to the next statement.

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
rekhaswaminathan: Good idea. I will do that

src/Diagnostics.RuntimeHost/Services/CacheService/TableCacheService.cs
rekhaswaminathan: Right, I will move to Controller then

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -0,0 +1,142 @@
...
+
+                DiagnosticsETWProvider.Instance.LogSourceWatcherMessage(nameof(StorageService), $"Retrieving detectors from table");
```
ShekharGupta1988, line 72: Is it  better to not use Source Watcher Message event for Storage Service Logs. This class is not source watcher and going forward would be hard to differentiate.

rekhaswaminathan: Okay, I can create new EventIds for Storage Service

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -708,6 +704,63 @@ private async Task<IEnumerable<DiagnosticApiResponse>> ListDetectorsInternal(Run
...
+                    // Assign the score to detector if it exists in search results, else default to 0
+                    allDetectorsFromStorage.ForEach(entity =>
+                    {
+                        var detectorWithScore = (searchResults != null) ? searchResults.Results.FirstOrDefault(x => x.Detector == entity.RowKey) : null;
+                        entity.Score = detectorWithScore != null ? detectorWithScore.Score : 0;
+                    });
+
+                    // Filter only postive score detectors.
+                    allDetectorsFromStorage = allDetectorsFromStorage.Where(x => x.Score > 0).ToList();
+                }
```
SharmaAjay19, line 724: Should we add the logging on search results at the end of if block here - Just like we are doing below for sourcewatcher results.

var logMessage = new { InsightName = "SearchResultsPublic", InsightData = allDetectorsFromStorage.Select(p => new { Id = p.RowKey, Score = p.Score }) };
                DiagnosticsETWProvider.Instance.LogInternalAPIInsights(context.OperationContext.RequestId, JsonConvert.SerializeObject(logMessage));

rekhaswaminathan: Sure, I missed it. I will add it

PR #491, opened 2020-04-23
Store detector type from definition attribute

rekhaswaminathan:
- Store the DetectorType column from Definition Attribute to table
- Add unit tests cases for loading analysis type detectors
- If `loadOnlyPublicDetectors` is set to true, apply the filter. Otherwise return all detectors
- Minor fix for loading ASE detectors


src/SourceWatcherFuncApp/Utilities/EntityHelper.cs
```diff
@@ -31,7 +30,7 @@ public static DetectorEntity PrepareEntityForLoad(Stream streamAssemblyData, str
                 detectorPackage.IsInternal = resourceFilter != null? resourceFilter.InternalOnly : false;
                 detectorPackage.SupportTopicList = invoker.EntryPointDefinitionAttribute != null ? invoker.EntryPointDefinitionAttribute.SupportTopicList : new List<SupportTopic>() ;
                 detectorPackage.AnalysisTypes = invoker.EntryPointDefinitionAttribute != null ? invoker.EntryPointDefinitionAttribute.AnalysisTypes : new List<string>();
-
+                detectorPackage.DetectorType = invoker.EntryPointDefinitionAttribute != null ? invoker.EntryPointDefinitionAttribute.Type.ToString() : "Detector";
```
hforeste, line 33: instead of static string of "Detector" you can do `default(DetectorType)`

rekhaswaminathan: Good point

rekhaswaminathan: But how do we know the default of `DetectorType `is "Detector"?

PR #494, opened 2020-04-25
Set weak consistency for queries running against Antares Kusto clusters.

nmallick1:
More info on weak consistency 
https://kusto.azurewebsites.net/docs/concepts/queryconsistency.html?q=queryconsistency 

This change will help alleviate some pressure from the admin nodes making the cluster more responsive. The queries however "might" have stale data for up to 5 minutes only from the time of ingestion. Will not add to query latency.


nmallick1:
@ShekharGupta1988 , I was planning on monitoring for any negative impact post this setting before updating master.

PR #495, opened 2020-04-27
Allow Kusto heart beat task to run in Prod env

rekhaswaminathan:


src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -50,9 +51,10 @@ public void Dispose()
             _cancellationToken.Cancel();
         }
 
-        public KustoHeartBeatService(KustoDataProviderConfiguration configuration)
+        public KustoHeartBeatService(KustoDataProviderConfiguration configuration, IHostingEnvironment hostingEnvironment)
         {
             _configuration = configuration;
+            runHeartBeatQuery = hostingEnvironment.IsProduction() || hostingEnvironment.IsEnvironment("UnitTest");
```
hforeste, line 57: Can you instead put this logic in the startup.cs class? I think the design pattern for creating services in dotnetcore is that if you simply want the service then we should add the service in the startup class or conditionally add. The services themselves are unaware of some kind of feature flag.

ShekharGupta1988, line 57: should we control this via settings Rekha?
It will give us flexibility to turn it on/off on mulitple environments in future.

rekhaswaminathan: Cant move the logic to startup because the interface has a method `GetClusterNameFromStamp` which is used in other places. So the class has to be initialized with the requried config

rekhaswaminathan: Okay sure

hforeste: Ah I see makes sense. Can we instead then pass a `boolean  enableHeartbeats` which equal the setting that Shekhar suggested? 

> "should we control this via settings Rekha?
It will give us flexibility to turn it on/off on mulitple environments in future."

rekhaswaminathan: Yes, I am going to do that

rekhaswaminathan:
> maybe EnableHeartBeatQuery should be set to false by default? and enabled in prod keyvault?

If we enable it prod keyvault, since staging slot reference it, they will get it value as true right?

PR #498, opened 2020-04-28
Set weak consistency for queries runing against Antares kusto clusters.

nmallick1:
More info on weak consistency
https://kusto.azurewebsites.net/docs/concepts/queryconsistency.html?q=queryconsistency 

This change will help alleviate some pressure from the admin nodes making the cluster more responsive. The queries however "might" have stale data for up to 5 minutes only from the time of ingestion. Will not add to query latency.


rekhaswaminathan:
Hey Nazeef, Just read through the documentation. I know in liberation we use the Kusto SDK but in DiagRole we are using HTTPClient. According to [this ](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/api/connection-strings/kusto)doc, it says the consistency is set part of connection string properties, but over here it seems like you are setting in request payload. Is it the same?

rekhaswaminathan:
please ignore that, saw the REST API equivalent.

PR #503, opened 2020-05-04
Publish flow

rekhaswaminathan:
- In publish flow, upload dll to blob container and metadata to table
- Logging improvements to include time taken
- Unit tests for load blob and get blob


src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            if (pkg == null)
+            {
+                throw new ArgumentNullException(nameof(pkg));
+            }
+            try
+            {
+                var blobName = $"{pkg.Id.ToLower()}/{pkg.Id.ToLower()}.dll";
+                var etag = await storageService.LoadBlobToContainer(blobName, pkg.DllBytes);
+                if (string.IsNullOrEmpty(etag))
```
ShekharGupta1988, line 46: nit: IsNullorWhitespace

src/Diagnostics.RuntimeHost/Services/GithubClient/GithubClient.cs
```diff
@@ -178,6 +180,13 @@ public async Task CreateOrUpdateFiles(IEnumerable<CommitContent> commits, string
             await _octokitClient.Git.Reference.Update(_userName, _repoName, headRef, new ReferenceUpdate(commit.Sha));
         }
 
+        public async Task<GitHubCommit> GetCommitByPath(string filePath)
+        {
+            var request = new CommitRequest { Path = filePath, Sha = _branch };
+            var commitLists = await _octokitClient.Repository.Commit.GetAll(_userName, _repoName, request);
+            return commitLists != null ? commitLists[0] : null;
```
ShekharGupta1988, line 187: nit: you'll get an OutOfRange exception if CommitLists is Empty.
instead of returning commitLists[0], you can return commitLists.FirstOrDefault()

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
```diff
@@ -0,0 +1,82 @@
...
+            hostingEnvironment = env;
+            configuration = config;
+            gitHubClient = new GithubClient(env, config);
+        }
+        public async Task<HealthCheckResult> CheckHealthAsync(CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
+        }
+
+        public async Task CreateOrUpdatePackage(Package pkg)
```
ShekharGupta1988, line 36: Hi Rekha, just thinking about this approach little bit. We are first updating the blob with new dll and then the table storage. Both these are not one atomic operation. 
How should we handle a situation where the blob gets updated but the table storage operation failed due to any reason?

(However most of the update operations are only dll updates and definition updates, but something to think about here)

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
```diff
@@ -0,0 +1,82 @@
...
+                throw new ArgumentNullException(nameof(pkg));
+            }
+            try
+            {
+                var blobName = $"{pkg.Id.ToLower()}/{pkg.Id.ToLower()}.dll";
+                var etag = await storageService.LoadBlobToContainer(blobName, pkg.DllBytes);
+                if (string.IsNullOrWhiteSpace(etag))
+                {
+                    DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(StorageWatcher), $"Uploading {pkg.Id} to blob failed, not proceeding further");
+                    return;
```
ShekharGupta1988, line 49: should we throw a specific exception here instead of returning success?

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
rekhaswaminathan: I have few ideas here, may be we can chat offline about this

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -362,6 +365,7 @@ protected async Task<IActionResult> PublishPackage(Package pkg)
             }
 
             await _sourceWatcherService.Watcher.CreateOrUpdatePackage(pkg);
+            await storageWatcher.CreateOrUpdatePackage(pkg);
```
hforeste, line 368: Can we make these two calls concurrent? Also what is your intention in the case of failure from storage watcher? Will it silently fail until we flip the switch or will you fail the API even if git hub watcher succeeds?

rekhaswaminathan: So we cannot make the calls concurrent because we want to store the GitHub SHA in the table so storage watcher should execute after GitHub watcher finishes. 

Also, I have now coded it to be silently failing while logging the exceptions. May be once we are fully on new system we can fail the API

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
rekhaswaminathan: Since we are still in beta mode for this, I thought it would be okay to just log the exception than fail the API by throwing exception

ShekharGupta1988: Oh I see.. this gets called along with GitHub package update.
Then before switching to the new source watcher completely Rekha, we would have to run your function app to make sure that blobs are up-to date.

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -126,6 +140,49 @@ public async Task<DiagEntity> LoadDataToTable(DiagEntity detectorEntity)
...
+                using (var uploadStream = new MemoryStream(Convert.FromBase64String(contents)))
+                {
+                    await cloudBlob.UploadFromStreamAsync(uploadStream);               
+                }
+                await cloudBlob.FetchAttributesAsync();
+                timeTakenStopWatch.Stop();
+                var uploadResult = cloudBlob.Properties;  
+                DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(StorageService), $"Loaded {blobname}, etag {uploadResult.ETag}, time taken {timeTakenStopWatch.ElapsedMilliseconds}");
+                return uploadResult.ETag;
+            } catch (Exception ex)
```
hforeste, line 160: Are you going to do any retries?

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
rekhaswaminathan: Yes, we can do that

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
rekhaswaminathan: The storage client library has built-in retry policies. https://docs.microsoft.com/en-us/azure/architecture/best-practices/retry-service-specific#azure-storage

hforeste: Is it built in or do you have to configure it? When I read the docs you sent me it looks like you have to configure it with some `BlobRequestOptions` class

rekhaswaminathan: So we can use `BlobRequestOptions` if we want to configure it manually. These are default values:

image

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
hforeste: What happens in the case when it is national clouds?

rekhaswaminathan: @hforeste I have added an empty class that gets injected for national clouds. Please review the changes again.

src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -214,7 +214,13 @@ public void ConfigureServices(IServiceCollection services)
             }
             services.AddSingleton<IStorageService, StorageService>();
             services.AddSingleton<IDiagEntityTableCacheService, DiagEntityTableCacheService>();
-            services.AddSingleton<ISourceWatcher, StorageWatcher>();
+            if(IsPublicAzure())
+            {
+                services.AddSingleton<ISourceWatcher, StorageWatcher>();
+            } else
+            {
+                services.AddSingleton<ISourceWatcher, NationalCloudStorageWatcher>();
```
hforeste, line 222: 👍

src/Diagnostics.ModelsAndUtils/Models/Storage/DiagEntity.cs
```diff
@@ -148,5 +148,36 @@ public List<string> AnalysisTypes
         /// Detector type - Detector, Analysis, CategoryOverview
         /// </summary>
         public string DetectorType { get; set; }
+
+        /// <summary>
+        /// List of dependencies for a detector
+        /// </summary>
+        [IgnoreProperty]
+        public Dictionary<string, string> Dependencies
```
SharmaAjay19, line 156: (y)

PR #504, opened 2020-05-05
The basic framework for integration tests.

craigmaher:
This adds very basic integration tests and the ability to run them in dev ops.


tests/Diagnostics.Tests.Integration.CompilerHost/UnitTest1.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using Xunit;
+using Microsoft.AspNetCore.Mvc.Testing;
+using Diagnostics;
+using System.Threading.Tasks;
+
+namespace Diagnostics.Tests.Integration.CompilerHost
+{
+    public class UnitTest1 : IClassFixture<WebApplicationFactory<Diagnostics.CompilerHost.Startup>>
```
rekhaswaminathan, line 9: Can we give a specific name for class and method?

rekhaswaminathan:
@craigmaher How can we run the integration test in our machines locally?

craigmaher:
These tests use the xUnit framework so you simply can run them from within visual studio (make sure you are signed in with you Microsoft account). Or you can run them outside of visual studio just make sure to do the “az login” command on Azure CLI beforehand, let me know if the tests don’t work for you locally.

PR #507, opened 2020-05-13
Arm template update

craigmaher:
This aims to update the arm templates so we can deploy all 6 prod app services as well as deploy the front door configurations separate from the 4 new prod instances (and use the same parameter file for both app service deployment and front door deployment)


infrastructure/ArmTemplates/DeployAll.parameters.json
```diff
@@ -10,37 +10,28 @@
...
         },
-        "RuntimeHost2":{
-            "reference": {
-              "keyVault": {
-                "id": "/subscriptions/0e7b3183-4428-4c1f-93cd-cb2a26616a8a/resourcegroups/AppServiceDiagnosticsResources/providers/Microsoft.KeyVault/vaults/DevOpsDeploymentKV"
-              },
-              "secretName": "DiagRuntimehostNameWUS"
-            }
-        }
+        "RuntimeHostProd1": "todo1",
```
craigmaher, line 29: These names need to be in a key vault so we don't expose ourselves to a denial of service attack. But I don't have access to the key vault

infrastructure/ArmTemplates/KeyVaultDeploy.json
```diff
@@ -26,38 +26,46 @@
...
-                "enabledForDeployment": false,
                 "copy" : [
                     {
                         "name": "accessPolicies",
-                        "count": "[mul(length(parameters('sites')),2)]",
+                        "count": "[length(parameters('sites'))]",
                         "input": {
                             "tenantId": "[variables('tenantId')]",
-                            "objectId": "[reference(concat(resourceId(parameters('subscriptionId'), parameters('ResourceGroup'), 'Microsoft.Web/sites', parameters('sites')[div(copyIndex('accessPolicies'),2)].siteName),if(equals(mod(copyIndex('accessPolicies'),2),1),'/providers/Microsoft.ManagedIdentity/Identities/default','/slots/staging/providers/Microsoft.ManagedIdentity/Identities/default')), '2015-08-31-PREVIEW').principalId]",
+                            "objectId": "[reference(concat(resourceId(parameters('subscriptionId'), parameters('ResourceGroup'), 'Microsoft.Web/sites', parameters('sites')[copyIndex('accessPolicies')].siteName),'/providers/Microsoft.ManagedIdentity/Identities/default'), '2015-08-31-PREVIEW').principalId]",
```
rekhaswaminathan, line 68: Craig, how would this work if I pass in array of sites each belonging to different resource group?

infrastructure/ArmTemplates/KeyVaultDeploy.json
```diff
@@ -26,38 +26,46 @@
...
         {
-            "type": "Microsoft.KeyVault/vaults",
-            "name": "[parameters('KeyVaultName')]",
-            "apiVersion": "2015-06-01",
-            "location": "[parameters('KeyVaultLocation')]",
+            "type": "Microsoft.KeyVault/vaults/accessPolicies",
+            "name": "[concat(parameters('keyVaultName'), '/add')]",
+            "apiVersion": "2019-09-01",
+            "properties": {
+                "copy" : [
```
rekhaswaminathan, line 40: also, what does "copy" here do?

cindylovescoding:
Hey @craigmaher, Since we are enabling application insights for our runtimehost. Do we need to update the arm template to include the app insight settings?

infrastructure/ArmTemplates/KeyVaultDeploy.json
rekhaswaminathan: It can be different according to region. Example EUAP runtimehosts belong a specific resource group, and am2 in other

hforeste:
In Ev2 I'm not using the DeployAllAppServices, DeployAllParameters, and DeployAllFrontDoors template as we'd be doing more of a batch by batch rollout as oppose to a single deployment to deploy them all. I chose to use the most minimal unit of a template that we can use anywhere so I am using a variation of the antareskeyvaulttemplate https://github.com/Azure/Azure-AppServices-Diagnostics/blob/master2/infrastructure/ArmTemplates/antareskvtemplate.json and the keyvaulttemplate. We can also use the same approach in ADO in the meantime before we make the transition to Ev2. Whatever we do in the antareskevaulttemplate and the keyvaulttemplate should translate easily over to the new system since Ev2 is a superset of ARM.

cindylovescoding:
> > Hey @craigmaher, Since we are enabling application insights for our runtimehost. Do we need to update the arm template to include the app insight settings?
> 
> Yes we do need to include those settings, If you can let me know what those settings are I can work to add them.

Thank you Craig!
So here is the app settings section to add:

`  "ApplicationInsights": {
    "InstrumentationKey": ""
  }`

rekhaswaminathan:
Hey Craig, I have one request:
Can you add the below app setting to staging slot 
`Kusto:EnableHeartBeatQuery = false`

PR #508, opened 2020-05-13
Adding some more integration tests

craigmaher:
adding a few more integration tests to make sure list detectors and get detectors are working.


tests/Diagnostics.Tests.Integration.RuntimeHost/IntegrationTests.cs
```diff
@@ -30,5 +33,56 @@ public async Task HealthPingTest(string url)
             Assert.Equal("application/json; charset=utf-8",
                 response.Content.Headers.ContentType.ToString());
         }
+
+        [Theory]
+        [InlineData("/subscriptions/ef90e930-9d7f-4a60-8a99-748e0eea69de/resourceGroups/detector-development/providers/Microsoft.Web/sites/buggybakery/detectors")]
```
ShekharGupta1988, line 38: Is it possible to feed this data via config instead of code?

ShekharGupta1988: Check this out: https://andrewlock.net/creating-a-custom-xunit-theory-test-dataattribute-to-load-data-from-json-files/

ShekharGupta1988: You can imagine using the same method to test List Detectors for multiple resource types by changing the resource and minimum expected detector count.
- Web App
- Function App
- Logic App
- AKS

tests/Diagnostics.Tests.Integration.RuntimeHost/IntegrationTests.cs
```diff
@@ -30,5 +33,56 @@ public async Task HealthPingTest(string url)
...
+            // Act
+            string responseString = await response.Content.ReadAsStringAsync();
+            var detectors = JsonConvert.DeserializeObject<DiagnosticApiResponse[]>(responseString);
+
+            // Assert
+            Assert.True(detectors.Length > 5);
+        }
+
+        [Theory]
+        [InlineData("/subscriptions/ef90e930-9d7f-4a60-8a99-748e0eea69de/resourceGroups/detector-development/providers/Microsoft.Web/sites/buggybakery/detectors/httpservererrors")]
```
ShekharGupta1988, line 63: same here

tests/Diagnostics.Tests.Integration.RuntimeHost/TestData.json
```diff
@@ -0,0 +1,8 @@
+{
+  "ListDetector": [
```
ShekharGupta1988, line 2: possible to checkin the empty file and not have actual resources (with subid etc) here ?

ShekharGupta1988: have the real values during test runtime

PR #514, opened 2020-06-03
Kusto Retry Logic : Retries on specific exception types

ShekharGupta1988:
This PR Contains:
- Kusto Retry Logic for specific Exception Types (governed by configuration)
- Using Backup cluster for Last Retry attempt (governed by configuration)
           - For instance, if the original query is targeted towards follower cluster, the final retry will happen against main cluster and vice-versa
           - for Partners who doesnt have backup clusters, the final retry will be against the same cluster.
- This logic will be automatically applied to `ExecuteQuery` and `ExecuteClusterQuery` but not to `Heartbeat Logic`
- To begin with, we are only retrying in case of :
              - Exceptions of type `WeakConsistencyEntityNotFoundException` or `InternalServiceError (520-UnknownError)` . In Last 7 Days, there are 3085 exceptions with these errors and P90 is 8.5 sec.
              - Only those kusto requests that failed within 10 seconds

- please note : I was not able to use Generic Retry logic we have since there were special handling required for kusto retry like change the cluster for last retry and so on.

### DIFF
--- a/src/Diagnostics.DataProviders/KustoSDKClient.cs
+++ b/src/Diagnostics.DataProviders/KustoSDKClient.cs
@@ -15,18 +15,20 @@ using Diagnostics.ModelsAndUtils.Models;
 using Newtonsoft.Json;
 using Diagnostics.DataProviders.Exceptions;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Diagnostics.DataProviders
 {
     class KustoSDKClient : IKustoClient
     {
+        private readonly KustoDataProviderConfiguration _config;
         private string _requestId;
         private string _kustoApiQueryEndpoint;
         private string _appKey;
         private string _clientId;
         private string _aadAuthority;
         private static ConcurrentDictionary<Tuple<string, string>, ICslQueryProvider> QueryProviderMapping;
-        private static List<string> exceptionsToRetryFor = new List<string>() { "" };
+        private static List<string> exceptionsToRetryFor = new List<string>();
 
         /// <summary>
         /// Failover cluster mapping
@@ -39,12 +41,19 @@ namespace Diagnostics.DataProviders
             {
                 QueryProviderMapping = new ConcurrentDictionary<Tuple<string, string>, ICslQueryProvider>();
             }
+
+            _config = config;
             _requestId = requestId;
             _kustoApiQueryEndpoint = config.KustoApiEndpoint + ":443";
             _appKey = config.AppKey;
             _clientId = config.ClientId;
             _aadAuthority = config.AADAuthority;
             FailoverClusterMapping = config.FailoverClusterNameCollection;
+
+            if (!string.IsNullOrWhiteSpace(_config.ExceptionsToRetryFor))
+            {
+                exceptionsToRetryFor = _config.ExceptionsToRetryFor.Split(DataProviderConstants.CommonSeparationChars).Select(p => p.Trim()).ToList();
+            }
         }
 
         private ICslQueryProvider Client(string cluster, string database)
@@ -120,12 +129,7 @@ namespace Diagnostics.DataProviders
 
         public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, string database, string requestId = null, string operationName = null)
         {
-            bool useBackupClusterForLastAttempt = true;
             int retryCount = 0;
-            int retryDelayInSeconds = 10;
-            int maxRetries = 5;
-            double maxFailureResponseTimeInSecondsForRetry = 10.0;
-
             string source = string.IsNullOrWhiteSpace(operationName) ? "KustoSDKClient_ExecuteQueryAsync" : operationName;
             
             DateTime invocationStartTime = default;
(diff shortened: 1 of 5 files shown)


src/Diagnostics.DataProviders/KustoSDKClient.cs
```diff
@@ -117,7 +129,86 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
...
+                    }
+
+                    attempt++;
+                }
+
+                if (attempt <= _config.MaxRetryCount) await Task.Delay(_config.RetryDelayInSeconds * 1000);
+
+            } while (attempt <= _config.MaxRetryCount);
+
+            if(executeQueryTask.IsCompleted && !executeQueryTask.IsFaulted && !executeQueryTask.IsCanceled)
```
puneetg1983, line 206: >executeQueryTask.IsCompleted && !executeQueryTask.IsFaulted && !executeQueryTask.IsCanceled [](start = 15, length = 91)

You can use Task.IsCompletedSuccessfully
[https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.iscompletedsuccessfully?view=netcore-3.1](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.iscompletedsuccessfully?view=netcore-3.1)

src/Diagnostics.DataProviders/KustoSDKClient.cs
```diff
@@ -117,7 +129,86 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
...
+                            exceptionDetails
+                            );
+
+                    // Logic to check if retry needs to continue
+                    if(totalResponseTime.TotalSeconds > _config.MaxFailureResponseTimeInSecondsForRetry || !IsExceptionRetryable(attemptException))
+                    {
+                        DiagnosticsETWProvider.Instance.LogRetryAttemptMessage(
+                        requestId ?? string.Empty,
+                        source,
+                        $"Not continuing Retries after Attempt : {attempt}, Cluster: {cluster}, Query : {query}"
```
puneetg1983, line 193: >Not continuing Retries [](start = 26, length = 22)

Should we also log totalResponseTime.TotalSeconds ?

src/Diagnostics.DataProviders/KustoSDKClient.cs
```diff
@@ -15,48 +14,61 @@
...
         private string _clientId;
         private string _aadAuthority;
         private static ConcurrentDictionary<Tuple<string, string>, ICslQueryProvider> QueryProviderMapping;
+        private static List<string> exceptionsToRetryFor = new List<string>();
 
         /// <summary>
         /// Failover cluster mapping
         /// </summary>
-        public ConcurrentDictionary<string, string> FailoverClusterMapping { get; set; }
+        private ConcurrentDictionary<string, string> FailoverClusterMapping { get; set; }
```
puneetg1983, line 36: >private [](start = 8, length = 7)

Is this property not used anywhere else in the project?

src/Diagnostics.DataProviders/KustoSDKClient.cs
ShekharGupta1988: Its already logged as part of attempt summary event just above.  But i can add it to this event as well

puneetg1983: Never mind, I see it is logged above...

---
In reply to: [434590128](https://github.com/Azure/Azure-AppServices-Diagnostics/pull/514#discussion_r434590128) [](ancestors = 434590128)

src/Diagnostics.DataProviders/KustoSDKClient.cs
ShekharGupta1988: Yup. This shouldnt be public

src/Diagnostics.DataProviders/KustoSDKClient.cs
```diff
@@ -15,48 +14,61 @@
...
             _requestId = requestId;
             _kustoApiQueryEndpoint = config.KustoApiEndpoint + ":443";
             _appKey = config.AppKey;
             _clientId = config.ClientId;
             _aadAuthority = config.AADAuthority;
             FailoverClusterMapping = config.FailoverClusterNameCollection;
+
+            if (!string.IsNullOrWhiteSpace(_config.ExceptionsToRetryFor))
+            {
+                exceptionsToRetryFor = _config.ExceptionsToRetryFor.Split(DataProviderConstants.CommonSeparationChars).Select(p => p.Trim()).ToList();
```
SharmaAjay19, line 55: I like this split.

src/Diagnostics.DataProviders/KustoSDKClient.cs
```diff
@@ -117,7 +129,86 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
...
+                            $"Attempt : {attempt},  IsSuccessful : {attemptException == null}, Cluster: {cluster}, Query : {query}",
+                            Convert.ToInt64(totalResponseTime.TotalMilliseconds),
+                            invocationStartTime.ToString("HH:mm:ss.fff"),
+                            invocationEndTime.ToString("HH:mm:ss.fff"),
+                            exceptionType,
+                            exceptionDetails
+                            );
+
+                    // Logic to check if retry needs to continue
+                    if(totalResponseTime.TotalSeconds > _config.MaxFailureResponseTimeInSecondsForRetry || !IsExceptionRetryable(attemptException))
```
cindylovescoding, line 188: Do we want to check if "IsSuccessful" here too?  If the retry succeed, we don't need to continue here as well right?

ShekharGupta1988: Yup. The IsExceptionRetryable() takes care of that :)

cindylovescoding: Oh I see. It's checking null exception. Got it!

PR #516, opened 2020-06-04
Heuristic string manipulation for search

SharmaAjay19:
Sanitize search string before query.
Break long strings into smaller chunks to get better hits across different possible domains.

### DIFF
--- a/src/Diagnostics.AIProjects/SearchAPI/SearchModule/TextSearchModule.py
+++ b/src/Diagnostics.AIProjects/SearchAPI/SearchModule/TextSearchModule.py
@@ -1,4 +1,4 @@
-import os, shutil, json
+import os, shutil, json, re
 from SearchModule.Logger import loggerInstance
 from SearchModule.ModelInfo import ModelInfo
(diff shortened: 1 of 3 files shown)


src/Diagnostics.AIProjects/SearchAPI/SearchModule/TextSearchModule.py
```diff
@@ -23,7 +38,14 @@ def __init__(self, modelpackagepath):
             self.model = WmdSearchModel(modelpackagepath)
 
     def queryDetectors(self, query=None):
-        return self.model.queryDetectors(query)
+        cleansed_query = " ".join(re.sub(r'[^(0-9a-zA-Z )]+', " ", query).split())
+        # Break the query into smaller chunks if number of word is greater than 6
+        if len(cleansed_query.split())>6:
```
ShekharGupta1988, line 43: So this means that the original query has more than 6 words?

SharmaAjay19: Yup, if the original query has more than 6 words, we break it down.

PR #517, opened 2020-06-04
Update NRP entities for calling geomaster's CollectVnetConfiguration

dejo-msft:


src/Diagnostics.ModelsAndUtils/Models/GeoMaster/NetworkResourceProviderEntities.cs
```diff
@@ -59,64 +127,123 @@ public int CompareTo(NetworkResourceProviderNsgRule other)
...
-        public IEnumerable<string> Actions { get; set; }
+        public EffectiveRoute[] EffectiveRoutes { get; set; }
     }
 
-    public class NetworkResourceProviderSubnet
+    /// <summary>
+    /// Message envelope that contains the common Azure resource manager properties and the resource provider specific content.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class ResponseMessageEnvelope<T>
```
hforeste, line 184: Instead of using this class you can use [GeomasterResponse.cs](https://github.com/Azure/Azure-AppServices-Diagnostics/blob/master2/src/Diagnostics.ModelsAndUtils/Models/GeoMaster/GeoMasterResponse.cs) class

PR #522, opened 2020-06-10
Rekha/cleanup code

rekhaswaminathan:
Cleaning up some old code which was used in v1 of Application Change Analysis. We have now migrated to use their blade so this code is not being used anymore


src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -790,40 +790,7 @@ private async Task<IEnumerable<DiagnosticApiResponse>> ListDetectorsInternal(Run
             if (context.ClientIsInternal)
             {
                 dataProvidersMetadata = GetDataProvidersMetadata(dataProviders);
-            }
-
-            var changeSetId = queryParams.Where(s => s.Key.Equals("changeSetId")).Select(pair => pair.Value);
```
ShekharGupta1988, line 795: QQ Rekha - will applens change analysis work after this ?

rekhaswaminathan: Yes Shekhar, because we are using forms instead of these named query string parameters

PR #523, opened 2020-06-11
Changes to enable ListDetectors call completely

rekhaswaminathan:
- With this change we can enable ListDetectors call in prod. If detector gets published, the cache service will poll storage every 60secs and keep the cache upto date

### DIFF
--- a/src/Diagnostics.RuntimeHost/Services/CacheService/DiagEntityTableCacheService.cs
+++ b/src/Diagnostics.RuntimeHost/Services/CacheService/DiagEntityTableCacheService.cs
@@ -16,11 +16,33 @@ namespace Diagnostics.RuntimeHost.Services.CacheService
 
         private IStorageService storageService;
 
+        int cacheExpirationTimeInSecs = 60;
+
         public DiagEntityTableCacheService(IStorageService service)
         {
             cache = new ConcurrentDictionary<string, List<DiagEntity>>();
             storageService = service;
+            StartPolling();
         }
+
+        private async void StartPolling()
+        {
+            do
+            {
+                await Task.Delay(cacheExpirationTimeInSecs * 1000);
+                var detectorResult = await storageService.GetEntitiesByPartitionkey("Detector");
+                if (detectorResult != null)
+                {
+                    AddOrUpdate("Detector", detectorResult);
+                }
+                var gistResult = await storageService.GetEntitiesByPartitionkey("Gist");
+                if(gistResult != null)
+                {
+                    AddOrUpdate("Gist", gistResult);
+                }
+            } while (true);
+        }
+
         public void AddOrUpdate(string key, List<DiagEntity> value)
         {
             cache.AddOrUpdate(key, value, (existingKey, existingValue) => value);
@@ -70,5 +92,7 @@ namespace Diagnostics.RuntimeHost.Services.CacheService
         {
             return storageService.GetStorageFlag();
         }
+
+      
     }
 }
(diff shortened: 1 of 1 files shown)


src/Diagnostics.RuntimeHost/Services/CacheService/DiagEntityTableCacheService.cs
```diff
@@ -16,11 +16,33 @@ public class DiagEntityTableCacheService : IDiagEntityTableCacheService
...
+        {
+            do
+            {
+                await Task.Delay(cacheExpirationTimeInSecs * 1000);
+                var detectorResult = await storageService.GetEntitiesByPartitionkey("Detector");
+                if (detectorResult != null)
+                {
+                    AddOrUpdate("Detector", detectorResult);
+                }
+                var gistResult = await storageService.GetEntitiesByPartitionkey("Gist");
```
ShekharGupta1988, line 38: You can potentially call these two tasks (detector task and gist task) in parallel to speed up things.

src/Diagnostics.RuntimeHost/Services/CacheService/DiagEntityTableCacheService.cs
```diff
@@ -16,11 +17,34 @@ public class DiagEntityTableCacheService : IDiagEntityTableCacheService
...
             cache = new ConcurrentDictionary<string, List<DiagEntity>>();
             storageService = service;
+            StartPolling();
         }
+
+        private async void StartPolling()
+        {
+            do
+            {
+                await Task.Delay(cacheExpirationTimeInSecs * 1000);
```
ShekharGupta1988, line 33: QQ Rekha - What is the expected behavior if there was an exception in the logic below. 
Should we log the exception and still continue to poll. (and maybe add an email alert to begin with to see if there are exceptions here in PROD)

src/Diagnostics.RuntimeHost/Services/CacheService/DiagEntityTableCacheService.cs
```diff
@@ -16,11 +17,34 @@ public class DiagEntityTableCacheService : IDiagEntityTableCacheService
...
+                var detectorResult = await storageService.GetEntitiesByPartitionkey("Detector");
+                if (detectorResult != null)
+                {
+                    AddOrUpdate("Detector", detectorResult);
+                }
+                var gistResult = await storageService.GetEntitiesByPartitionkey("Gist");
+                if(gistResult != null)
+                {
+                    AddOrUpdate("Gist", gistResult);
+                }
```
ShekharGupta1988, line 44: I would also add a log statement capturing the completion of the call and how much time taken.

src/Diagnostics.RuntimeHost/Services/CacheService/DiagEntityTableCacheService.cs
rekhaswaminathan: If there is an exception while retrieving data from storage, the exception is logged and null is returned. We check for null before adding to cache. So cache wont be updated and polling continues.

ShekharGupta1988: One more point to think about : Today our detector changes (or new detectors) light up in 10 sec. Correct me if I am wrong here but it looks like now it will take 60 sec to see the new detector after publish. While I think 10 sec is too aggressive for polling, maybe starting with 30 sec might not be a bad idea.

PR #524, opened 2020-06-12
IsMarkedForDeletion Column for detectors deleted in github

rekhaswaminathan:
My plan is to run the function app once so it will mark any detectors not present in github as deleted in storage table. While retrieving the detectors in ListDetectors call, we filter this column.

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/Storage/DiagEntity.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/Storage/DiagEntity.cs
@@ -183,6 +183,6 @@ namespace Diagnostics.ModelsAndUtils.Models.Storage
         /// <summary>
         /// Is detector marked for deletion in github.
         /// </summary>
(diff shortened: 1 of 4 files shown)


src/Diagnostics.ModelsAndUtils/Models/Storage/DiagEntity.cs
```diff
@@ -179,5 +179,10 @@ public List<string> AnalysisTypes
         /// Gist References stored as raw string in table.
         /// </summary>
         public string GistReferencesRaw { get; set; } = string.Empty;
+
+        /// <summary>
+        /// Is detector marked for deletion in github.
+        /// </summary>
+        public bool IsMarkedForDeletion { get; set; } = false;
```
ShekharGupta1988, line 186: nit: How about renaming this to be little bit for generic like 'IsDisabled'
You can think you using this column in future as well for any other potential scenarios (like temporary disabling detectors) including permanent detection of detectors.

PR #550, opened 2020-07-21
Move Kusto Cluster Mapping to Storage Table

rekhaswaminathan:
### Code Changes
- Table **detectorruntimeconfig** stores Kusto Cluster Mapping
- Table is polled every 30s to refresh cache 
- Removed some flags/conditions which I added initially to make StorageWatcher work for detectors/gist and have Github work for KustoCluster
- Sample data in table:
image

### Next steps

- Enable StorageWatcher in staging slot and republish kustoClusterMappings.json file for the following to backfill table: 
image
- Enable in Prod for euap alone

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/Storage/DiagConfiguration.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/Storage/DiagConfiguration.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.WindowsAzure.Storage.Table;
+
+namespace Diagnostics.ModelsAndUtils.Models.Storage
+{
+    public class DiagConfiguration: TableEntity
+    {
+        public DiagConfiguration()
+        {
+
+        }
+
+        public DiagConfiguration(string entityType, string resourceProvider)
+        {
+            PartitionKey = entityType;
+            RowKey = resourceProvider;
+        }
+
+        /// <summary>
+        /// Kusto Cluster Mapping 
+        /// </summary>
+        public string KustoClusterMapping;
+
+        /// <summary>
+        /// Github Sha of the json file
+        /// </summary>
+        public string GithubSha;
+    }
+}
--- a/src/Diagnostics.ModelsAndUtils/Models/Storage/DiagConfiguration.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/Storage/DiagConfiguration.cs
@@ -21,11 +21,16 @@ namespace Diagnostics.ModelsAndUtils.Models.Storage
         /// <summary>
         /// Kusto Cluster Mapping 
         /// </summary>
-        public string KustoClusterMapping;
+        public string KustoClusterMapping { get; set; } = string.Empty;
 
         /// <summary>
         /// Github Sha of the json file
         /// </summary>
-        public string GithubSha;
+        public string GithubSha { get; set; } = string.Empty;
+
+        /// <summary>
+        /// Flag to indicate if config is disabled
+        /// </summary>
+        public bool IsDisabled { get; set; } = false;
     }
 }
--- a/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
+++ b/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
@@ -35,6 +35,7 @@ namespace Diagnostics.RuntimeHost.Services.SourceWatcher.Watchers
         private Dictionary<EntityType, ICache<string, EntityInvoker>> _invokerDictionary;
         private int _pollingIntervalInSeconds = 30;
         private DateTime cacheLastModifiedTime;
+        private IKustoMappingsCacheService kustoCacheService;
 
         private bool LoadOnlyPublicDetectors
         {
@@ -50,7 +51,7 @@ namespace Diagnostics.RuntimeHost.Services.SourceWatcher.Watchers
         }
           
 
-        public StorageWatcher(IHostingEnvironment env, IConfiguration config, IStorageService service, IInvokerCacheService invokerCache, IGistCacheService gistCache)
+        public StorageWatcher(IHostingEnvironment env, IConfiguration config, IStorageService service, IInvokerCacheService invokerCache, IGistCacheService gistCache, IKustoMappingsCacheService kustoCache)
         {
             storageService = service;
             hostingEnvironment = env;
@@ -62,6 +63,7 @@ namespace Diagnostics.RuntimeHost.Services.SourceWatcher.Watchers
                 { EntityType.Signal, invokerCache},
                 { EntityType.Gist, gistCache}
             };
+            kustoCacheService = kustoCache;
             Start();
         }
 
@@ -93,6 +95,18 @@ namespace Diagnostics.RuntimeHost.Services.SourceWatcher.Watchers
             try
             {
                 await gitHubClient.CreateOrUpdateFiles(pkg.GetCommitContents(), pkg.GetCommitMessage());
+                if(pkg.GetCommitContents().Any(content => content.FilePath.Contains("kustoClusterMappings", StringComparison.CurrentCultureIgnoreCase)))
+                {
+                    var kustoMappingPackage = pkg.GetCommitContents().Where(c => c.FilePath.Contains("kustoClusterMappings", StringComparison.CurrentCultureIgnoreCase)).FirstOrDefault();
+                    var githubCommit = await gitHubClient.GetCommitByPath(kustoMappingPackage.FilePath);
+                    // store kusto cluster data
+                    var diagconfiguration = new DiagConfiguration{
+                        PartitionKey = "KustoClusterMapping",
+                        RowKey = kustoMappingPackage.FilePath.Split("/kustoClusterMapping.json")[0],
+                        GithubSha = githubCommit != null ? githubCommit.Commit.Tree.Sha : string.Empty,
+                        KustoClusterMapping = kustoMappingPackage.Content
+                    };
+                }
                 var blobName = $"{pkg.Id.ToLower()}/{pkg.Id.ToLower()}.dll";
                 var etag = await storageService.LoadBlobToContainer(blobName, pkg.DllBytes);
                 if (string.IsNullOrWhiteSpace(etag))
@@ -218,5 +232,16 @@ namespace Diagnostics.RuntimeHost.Services.SourceWatcher.Watchers
             
         }
 
+        private async Task StartKustoCacheRefresh(bool startup = false)
+        {
+            var diagConfigRows = await storageService.GetKustoConfiguration();
+            if(!startup)
+            {
+
+            } else
+            {
+
+            }
+        }
     }
 }
--- a/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
+++ b/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
@@ -35,8 +35,8 @@ namespace Diagnostics.RuntimeHost.Services.SourceWatcher.Watchers
         private Dictionary<EntityType, ICache<string, EntityInvoker>> _invokerDictionary;
         private int _pollingIntervalInSeconds = 30;
         private DateTime blobCacheLastModifiedTime;
-        private IKustoMappingsCacheService kustoCacheService;
-        private DateTime kustoCacheLastModifiedTime;
+        private IKustoMappingsCacheService kustoMappingsCacheService;
+        private DateTime kustoMappingsCacheLastModified;
         private Task kustoConfigDownloadTask;
 
         private bool LoadOnlyPublicDetectors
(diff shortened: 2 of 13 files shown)


src/Diagnostics.ModelsAndUtils/Models/Storage/DiagConfiguration.cs
```diff
@@ -0,0 +1,36 @@
...
+using Microsoft.WindowsAzure.Storage.Table;
+
+namespace Diagnostics.ModelsAndUtils.Models.Storage
+{
+    public class DiagConfiguration: TableEntity
+    {
+        public DiagConfiguration()
+        {
+
+        }
```
hforeste, line 13: Do you need this empty constructor?

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
```diff
@@ -50,7 +53,7 @@ private bool LoadOnlyPublicDetectors
         }
           
 
-        public StorageWatcher(IHostingEnvironment env, IConfiguration config, IStorageService service, IInvokerCacheService invokerCache, IGistCacheService gistCache)
+        public StorageWatcher(IHostingEnvironment env, IConfiguration config, IStorageService service, IInvokerCacheService invokerCache, IGistCacheService gistCache, IKustoMappingsCacheService kustoCache)
```
hforeste, line 56: Just a small nitpik to rename `kustoCache` to `kustoMappingsCache`. The naming could be misleading.

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
```diff
@@ -93,6 +97,21 @@ public virtual async Task CreateOrUpdatePackage(Package pkg)
...
                 await gitHubClient.CreateOrUpdateFiles(pkg.GetCommitContents(), pkg.GetCommitMessage());
+                // Insert Kusto Cluster Mapping to Configuration table.
+                if(pkg.GetCommitContents().Any(content => content.FilePath.Contains("kustoClusterMappings", StringComparison.CurrentCultureIgnoreCase)))
+                {
+                    var kustoMappingPackage = pkg.GetCommitContents().Where(c => c.FilePath.Contains("kustoClusterMappings", StringComparison.CurrentCultureIgnoreCase)).FirstOrDefault();
+                    var githubCommit = await gitHubClient.GetCommitByPath(kustoMappingPackage.FilePath);
+                    // store kusto cluster data
+                    var diagconfiguration = new DiagConfiguration{
+                        PartitionKey = "KustoClusterMapping",
+                        RowKey = kustoMappingPackage.FilePath.Split("/kustoClusterMappings.json")[0],
```
hforeste, line 108: You can also get the same value by using `pkg.Id.ToLower()` instead of `kustoMappingPackage.FilePath.Split("/kustoClusterMappings.json")[0]`

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
```diff
@@ -218,5 +251,47 @@ private async Task StartBlobDownload(bool startup = false)
...
+                if (configsToLoad.Count > 0)
+                {
+                    DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(StorageWatcher), $"Starting kusto config download to update cache, Number of configs to load: {configsToLoad.Count}, startup : {startup.ToString()} at {DateTime.UtcNow}");
+                }
+                foreach(var config in configsToLoad)
+                {
+                    var kustoMappingsStringContent = config.KustoClusterMapping;
+                    var kustoMappings = (List<Dictionary<string, string>>)JsonConvert.DeserializeObject(kustoMappingsStringContent, typeof(List<Dictionary<string, string>>));
+                    var resourceProvider = config.RowKey;
+                    if (!kustoCacheService.ContainsKey(resourceProvider) || (kustoCacheService.TryGetValue(resourceProvider, out List<Dictionary<string, string>> value) && !value.Equals(kustoMappings)))
```
hforeste, line 278: So for list of dictionaries this `value.Equals(kustoMappings)` doesn't work as we would like unfortunately. I ran a quick test on LINQPad just to confirm.

```
	var map1 = new List<Dictionary<string, string>>{
		new Dictionary<string, string>{
			{"hello", "world"}
		}
	};

	var map2 = new List<Dictionary<string, string>>{
		new Dictionary<string, string>{
			{"hello", "world"}
		}
	};

	Console.Write(map1.Equals(map2));

map1.Equals(map2) == false
```

hforeste: Apparently I tried doing something like this in the KustoMappinsController and I noticed it then too but never had the chance to address it. https://github.com/Azure/Azure-AppServices-Diagnostics/blob/main/src/Diagnostics.RuntimeHost/Controllers/Configuration/KustoClusterMappingsController.cs#L43-L47

hforeste: The kusto map doesn't get updated often so maybe for know just compare if the `Count` is the same. 

```
if (!kustoCacheService.ContainsKey(resourceProvider) || (kustoCacheService.TryGetValue(resourceProvider, out List<Dictionary<string, string>> value) && value.Count == kustoMappings.Count))
```

rekhaswaminathan: Oh ok. In that case, we can just update it without comparing actual strings. It wont make a difference

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
rekhaswaminathan: Does `pkg.Id.ToLower()` give something like `microsoftwebconfiguration` ?

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
hforeste: Here's something you can try `bool isEqual = isEqual = map1.All(m1 => m1.All(kv1 => map2.Any(m2 => m2.Contains(kv1)))) && map1.Count == map2.Count;`

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
hforeste: Yes. Judging by looking at https://github.com/Azure/Azure-AppServices-Diagnostics/blob/main/src/Diagnostics.RuntimeHost/Controllers/Configuration/KustoClusterMappingsController.cs#L38

rekhaswaminathan: I see it now! Thanks Hawk

src/Diagnostics.ModelsAndUtils/Models/Storage/DiagConfiguration.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.WindowsAzure.Storage.Table;
+
+namespace Diagnostics.ModelsAndUtils.Models.Storage
+{
+    public class DiagConfiguration: TableEntity
```
hforeste, line 8: Should we rename this something more accurate to the data its holding? When I see DiagConfiguration I'm thinking our platform configuration.

rekhaswaminathan: How about `DetectorConfiguration` ?

src/Diagnostics.ModelsAndUtils/Models/Storage/DiagConfiguration.cs
rekhaswaminathan: Yes, this is needed as `NullableStorageService` depends on it

src/Diagnostics.ModelsAndUtils/Models/Storage/DiagConfiguration.cs
hforeste: Are we planning to add different types of configurations into `DetectorConfiguration` otherwise maybe `KustoMappingConfiguration` is best?

rekhaswaminathan: I thought of keeping this generic and flexible to add anything in future if needed. That's why the partitionkey is set to "KustoClusterMapping". We can always add anything in future by setting value for partitionkey

src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs
```diff
@@ -218,5 +251,47 @@ private async Task StartBlobDownload(bool startup = false)
...
+                    {
+                        kustoCacheService.AddOrUpdate(resourceProvider, kustoMappings);
+                    }
+                }
```
hforeste: Should this `foreach` be within the aforementioned if clause `if (configsToLoad.Count > 0)`

hforeste: Technically it doesn't matter because foreach of an empty list yields nothing. So either way I'm fine with it 👍🏿

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -205,5 +210,72 @@ public async Task<int> ListBlobsInContainer()
...
+                    // Execute the operation.
+                    var diagConfigurations = await cloudTable.ExecuteQuerySegmentedAsync(tableQuery, tableContinuationToken);
+                    tableContinuationToken = diagConfigurations.ContinuationToken;
+                    if (diagConfigurations.Results != null)
+                    {
+                        diagConfigurationsResult.AddRange(diagConfigurations.Results);
+                    }
+                } while (tableContinuationToken != null);
+                timeTakenStopWatch.Stop();
+                DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(StorageService), $"GetConfiguration by Parition key {partitionkey} took {timeTakenStopWatch.ElapsedMilliseconds}");
```
hforeste, line 272: misspell check: Parition -> Partition

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -205,5 +210,72 @@ public async Task<int> ListBlobsInContainer()
...
+            try
+            {
+                CloudTable cloudTable = tableClient.GetTableReference(configurationTable);
+                await cloudTable.CreateIfNotExistsAsync();
+                var timeTakenStopWatch = new Stopwatch();
+                var partitionkey = "KustoClusterMapping";
+                var filterPartitionKey = TableQuery.GenerateFilterCondition(PartitionKey, QueryComparisons.Equal, partitionkey);
+                var tableQuery = new TableQuery<DiagConfiguration>();
+                tableQuery.Where(filterPartitionKey);
+                DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(StorageService), $"GetConfiguration by parition key {partitionkey}");
```
hforeste, line 257: misspell check: Parition -> Partition

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -39,11 +42,13 @@ public class StorageService : IStorageService
         private bool loadOnlyPublicDetectors;
         private bool isStorageEnabled;
         private CloudTable cloudTable;
+        private string configurationTable;
 
         public StorageService(IConfiguration configuration, IHostingEnvironment hostingEnvironment)
         {
             tableName = configuration["SourceWatcher:TableName"];
             container = configuration["SourceWatcher:BlobContainerName"];
+            configurationTable = configuration["SourceWatcher:DiagConfiguration"];
```
hforeste, line 51: Could we add these to the appsettings? Just so that we have a one source of truth of all application settings needed for the runtimehost.

rekhaswaminathan: you mean to appsettings.json?  I thought of adding it directly to keyvault

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
rekhaswaminathan: Good catch

src/Diagnostics.ModelsAndUtils/Models/Storage/DiagConfiguration.cs
hforeste: Awesome

hforeste: Maybe not call it `DetectorConfiguration`. Just thinking about what kind of data this table holds, KustoMapping data can change anytime during the runtime and I assume any other type of data coming into this table in the future should also by some form of dynamic data. Anything that is static will just be configuration data in appsettings and keyvault. So maybe call it `DetectorRuntimeConfiguration` or any name you can think of that is configuration focused and dynamic.

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
hforeste: Having it in both I think is good

PR #553, opened 2020-07-28
Using geo redundancy to achieve high availability

rekhaswaminathan:
This gives the flexibility for us to point some endpoints to secondary storage account. As we look to expand storage feature to all endpoints, this is particularly useful for endpoints that are located faraway from the primary storage account and can face higher latency.

### DIFF
--- a/src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
+++ b/src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
@@ -43,6 +43,7 @@ namespace Diagnostics.RuntimeHost.Services.StorageService
         private bool isStorageEnabled;
         private CloudTable cloudTable;
         private string detectorRuntimeConfigTable;
+        private bool isSecondaryAccount = false;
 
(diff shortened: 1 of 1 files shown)


src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -57,11 +58,15 @@ public StorageService(IConfiguration configuration, IHostingEnvironment hostingE
             {
                 var accountname = configuration["SourceWatcher:DiagStorageAccount"];
                 var key = configuration["SourceWatcher:DiagStorageKey"];
-                var storageAccount = new CloudStorageAccount(new StorageCredentials(accountname, key), accountname, "core.windows.net", true);
+                var storageAccount = new CloudStorageAccount(new StorageCredentials(accountname, key), accountname, "core.windows.net", true);
```
ShekharGupta1988, line 61: will this work for any other cloud as we have the DNS suffix hardcoded here?

rekhaswaminathan: Valid point, It wont work. I will change it

PR #554, opened 2020-07-28
Add RebootWorker to geomaster data provider

michimune:
This PR adds a new API to the geomaster data provider named RebootWorker. This allows detectors to reboot a worker in the specified server farm.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -518,6 +518,26 @@ namespace Diagnostics.DataProviders
         /// </example>
         public Task<string> RebootWorker(string subscriptionId, string resourceGroup, string serverFarmName, string workerName, CancellationToken cancellationToken = default(CancellationToken))
         {
+            if (string.IsNullOrWhiteSpace(subscriptionId))
+            {
+                throw new ArgumentNullException(nameof(subscriptionId));
+            }
+
+            if (string.IsNullOrWhiteSpace(resourceGroup))
+            {
+                throw new ArgumentNullException(nameof(resourceGroup));
+            }
+
+            if (string.IsNullOrWhiteSpace(serverFarmName))
+            {
+                throw new ArgumentNullException(nameof(serverFarmName));
+            }
+
+            if (string.IsNullOrWhiteSpace(workerName))
+            {
+                throw new ArgumentNullException(nameof(workerName));
+            }
+
             var path = $"subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/Microsoft.Web/serverFarms/{serverFarmName}/workers/{workerName}/reboot";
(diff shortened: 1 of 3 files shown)


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -489,6 +489,39 @@ public Task<string> GetLinuxContainerLogs(string subscriptionId, string resource
...
+        ///     var serverFarmName = (string)observerData.server_farm_name;
+        ///     var machineName = "RD0123456789";
+        ///
+        ///     var results = await dp.GeoMaster.RebootWorker(subId, rg, serverFarmName, machineName);
+        /// }
+        /// </code>
+        /// </example>
+        public Task<string> RebootWorker(string subscriptionId, string resourceGroup, string serverFarmName, string workerName, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            var path = $"subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/Microsoft.Web/serverFarms/{serverFarmName}/workers/{workerName}/reboot";
```
ShekharGupta1988, line 521: nitpick : can you add basic checks for the arguments?

michimune: You are right. I will add null checking for the parameters.

michimune:
That's my concern, too. Do you have a good idea to control?

ShekharGupta1988:
> That's my concern, too. Do you have a good idea to control?

Lets discuss this. The real question is do we want to do this via Diag Backend --> Geomaster  (since this will open a can of worms) or should we let "users" issue this reboot via ARM --> Geomaster  (much safer, they can only reboot their own workers etc)

michimune:
The latter approach is the best. I believe we are already doing it with some features, correct? We have some diagnostic tools which are only allowed to access caller's resources

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
michimune: Added null checks

ShekharGupta1988:
> The latter approach is the best. I believe we are already doing it with some features, correct? We have some diagnostic tools which are only allowed to access caller's resources

correct. Basically if you are looking to provide a way to add this "Reboot worker" option in detectors, then we should look to add this as a solution. When this gets rendered :
- Internally in applens, it will have customer ready content that CSS can share with the customers.
- Externally in Case Submission(or Diag and Solve), customers will be able to reboot their workers via ARM call (and OAuth token)

michimune:
Got it, thank you. I will update my changes.

michimune:
It turned out that we don't need this PR at all. It is going to be replaced with this PR instead: https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/pull/854

PR #559, opened 2020-08-13
add log statements for debugging

rekhaswaminathan:


src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -143,7 +142,8 @@ public async Task<DiagEntity> LoadDataToTable(DiagEntity detectorEntity)
                 TableResult result = await table.ExecuteAsync(insertOrReplaceOperation);
                 timeTakenStopWatch.Stop();
                 DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(StorageService), $"InsertOrReplace result : {result.HttpStatusCode}, time taken {timeTakenStopWatch.ElapsedMilliseconds}");
-                DiagEntity insertedCustomer = result.Result as DiagEntity;
+                DiagEntity insertedEntity = result.Result as DiagEntity;
+                DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(StorageService), $"Inserted entity {insertedEntity.RowKey.ToString()} timestamp: {insertedEntity.Timestamp}");
```
hforeste, line 146: Is the `insertedEntity.Timestamp` approximately the same as the `PreciseTimeStamp` of the above log message that says "InsertOrReplace result: {result.HttpStatusCode}..."?

rekhaswaminathan: Thats what I want to clarify. I want to see timestamp stored in storage table because we compare that with when the cache was last modified in order to do updates

PR #560, opened 2020-08-18
Refresh cache changes and switch to secondary account for read path

rekhaswaminathan:
- Refresh cache with detectors published in the last 5 mins (Reason behind this: when multiple workers are polling storage, the last cache modified time may be different and out of sync, so on best-effort basis refreshing the cache with detectors modified in last 5 mins)
- Retry 3 times against primary storage account and move to secondary for read path
- table.CreateIfNotExists and container.CreateIfNotExists is causing a lot of noise in the application insight logs. It is seen as dependency failures. So removed those statements and move creation of container and table to script

### DIFF
--- a/src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
+++ b/src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
@@ -102,6 +102,7 @@ namespace Diagnostics.RuntimeHost.Services.StorageService
                     if(attempt == retryThreshold)
                     {
                         tableRequestOptions.LocationMode = LocationMode.SecondaryOnly;
+                        DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(StorageService), $"Retrying table against secondary account after {attempt} attempts");
                     }
                     do
                     {
@@ -205,7 +206,7 @@ namespace Diagnostics.RuntimeHost.Services.StorageService
                     if (attempt == retryThreshold)
                     {
                         options.LocationMode = LocationMode.SecondaryOnly;
-                        DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(StorageService), $"Retrying against secondary account after {attempt} attempts");
+                        DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(StorageService), $"Retrying blob against secondary account after {attempt} attempts");
                     }
                     var timeTakenStopWatch = new Stopwatch();
                     timeTakenStopWatch.Start();
--- a/src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
+++ b/src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
@@ -86,11 +86,15 @@ namespace Diagnostics.RuntimeHost.Services.StorageService
             int attempt = 0;
             do
             {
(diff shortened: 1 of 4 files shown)


hforeste:
> Refresh cache with detectors published in the last 5 mins (Reason behind this: when multiple workers are polling storage, the last cache modified time may be different and out of sync, so on best-effort basis refreshing the cache with detectors modified in last 5 mins

What is the most amount of time a cache will not have the latest updates for detectors?

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -81,37 +82,52 @@ public StorageService(IConfiguration configuration, IHostingEnvironment hostingE
 
         public async Task<List<DiagEntity>> GetEntitiesByPartitionkey(string partitionKey = null)
         {
-            try
+            int retryThreshold = 2;
+            int attempt = 0;
```
hforeste, line 86: Have you thought about using https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.table.tablerequestoptions.retrypolicy?view=azure-dotnet#Microsoft_Azure_Cosmos_Table_TableRequestOptions_RetryPolicy? I know were not using the cosmos library but I believe this is an older concept in the older libraries too.

rekhaswaminathan: So there are native implementations of retry policy already in SDK so I did not want to implement anything special unless I find a reason to. https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.storage.blob.blobrequestoptions.retrypolicy?view=azure-dotnet-legacy

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -81,37 +82,52 @@ public StorageService(IConfiguration configuration, IHostingEnvironment hostingE
...
+                catch (Exception ex)
+                {
+                    DiagnosticsETWProvider.Instance.LogAzureStorageException(nameof(StorageService), ex.Message, ex.GetType().ToString(), ex.ToString());
+                    return null;
+                }
+                finally
+                {
+                    attempt++;
+                }
+            } while (attempt <= retryThreshold);
```
hforeste, line 130: Do you want to put some delay between retries in case storage not responding?

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -81,37 +82,52 @@ public StorageService(IConfiguration configuration, IHostingEnvironment hostingE
...
-                do
+                try
                 {
-                    // Execute the operation.
-                    var detectorList = await table.ExecuteQuerySegmentedAsync(tableQuery, tableContinuationToken);
-                    tableContinuationToken = detectorList.ContinuationToken;
-                    if (detectorList.Results != null)
+                    CloudTable table = tableClient.GetTableReference(tableName);
+                    var timeTakenStopWatch = new Stopwatch();
+                    partitionKey = partitionKey == null ? "Detector" : partitionKey;
```
ShekharGupta1988, line 93: nit: use String.IsNullOrWhitespace

PR #566, opened 2020-08-24
Use new method to get 1000+ directories, clean up

rekhaswaminathan:
- Use the new Github API to fetch 1000+ directories
- Update metadata column

### DIFF
--- a/src/SourceWatcherFuncApp/DiagSourceWatcher.cs
+++ b/src/SourceWatcherFuncApp/DiagSourceWatcher.cs
@@ -60,7 +60,7 @@ namespace Diag.SourceWatcher
                     {
                         if(githubdir.Url.Contains("trees"))
                         {
-                            githubdir.Url = $"https://api.github.com/repos/{config["Github:UserName"]}/{config["Github:RepoName"]}/contents/{githubdir.Name}?ref={config["Github:Branch"]}";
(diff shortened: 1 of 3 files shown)


src/SourceWatcherFuncApp/DiagSourceWatcher.cs
```diff
@@ -48,10 +54,14 @@ public async Task Run([TimerTrigger("0 0 0 * * *")]TimerInfo timerInfo, ILogger
...
                 foreach (var githubdir in githubDirectories)
                 {
-                    if (!githubdir.Type.Equals("dir", StringComparison.OrdinalIgnoreCase)) continue;
+                    if (!githubdir.Type.Equals("tree", StringComparison.OrdinalIgnoreCase)) continue;
 
                     try
                     {
+                        if(githubdir.Url.Contains("trees"))
+                        {
+                            githubdir.Url = $"https://api.github.com/repos/{config["Github:UserName"]}/{config["Github:RepoName"]}/contents/{githubdir.Name}?ref={config["Github:Branch"]}";
```
hforeste, line 63: I added a new method to the `IGitHubService` that does this if you'll like to use it. https://github.com/Azure/Azure-AppServices-Diagnostics/blob/main/src/SourceWatcherFuncApp/Services/IGithubService.cs#L190-L193

PR #567, opened 2020-08-26
Nationalcloud

hforeste:


src/Diagnostics.RuntimeHost/Services/GithubClient/AzureStorageSourceCodeClient.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            foreach (var commit in commits)
+            {
+                if (!Equals(commit.EncodingType, EncodingType.Base64))
+                {
+                    commit.Content = InternalAPIHelper.Base64Encode(commit.Content);
+                }
+                await _storageService.LoadBlobToContainer(commit.FilePath, commit.Content).ConfigureAwait(false);
+            }
```
rekhaswaminathan, line 52: We should save the metadata to table here

rekhaswaminathan: https://github.com/Azure/Azure-AppServices-Diagnostics/blob/3e2c1c450407e27fa61f876ddd473e987c12d78d/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs#L135

hforeste: I think your code is already updating the table

rekhaswaminathan: So on top of the regular updates to blob and table, you are saving each file (like metadata.json, package.json) to the blob container?

hforeste: Yeah specifically the detector and gist file and package.json

rekhaswaminathan: You can filter out dll and pdb here. Dll is saved as part of StorageWatcher and pdb is not required

PR #573, opened 2020-09-16
Nationalcloud

hforeste:
- Update AI SDK to enable log ingestion into different clouds
- Add a few stats to NCloudSourceWatcher


src/SourceWatcherFuncApp/NCloudSourceWatcher.cs
```diff
@@ -142,17 +156,43 @@ public NCloudSourceWatcher(IConfigurationRoot configurationRoot, IGithubService
...
+                TotalUpdates = totalUpdates,
+                TotalSuccess = totalSuccess,
+                TotalFailures = totalFailures
+            };
+
+            if (!string.IsNullOrWhiteSpace(config["CallbackUrl"]))
+            {
+                log.LogInformation($"Send results to {config["CallbackUrl"]}");
+                var content = new StringContent(JsonConvert.SerializeObject(response), Encoding.UTF8, @"application/json");
+                await httpClient.PostAsync(config["CallbackUrl"], content).ConfigureAwait(false);
```
rekhaswaminathan, line 183: What is the callbackUrl here?

hforeste: So this is for sending stats to any HTTP service that want's to know about these periodic syncs. I have a LogicApp on the other end that is getting called by this FunctionApp via this callback URL.

rekhaswaminathan (approved):

PR #574, opened 2020-09-17
Storage throttling

rekhaswaminathan:
Last week we received throttling errors from storage. According to storage [documentation ](https://docs.microsoft.com/en-us/azure/storage/tables/storage-performance-checklist#targets-for-data-operations) we might we hitting entities per second(partition) limit because we are using [partition query](https://docs.microsoft.com/en-us/azure/storage/tables/storage-performance-checklist#querying) often

Fixes:

- When app starts up, get all detectors/gist based on partition key filter alone. For subsequent calls, use time range filter to get rows that have been modified in past 5 mins thereby reducing the number of entities need to be scanned each time

- Make StorageWatcher rely on cache service instead of querying storage directly. This leads to reduce number of outgoing calls to storage. Here is a graph that shows reduction in number of calls going to storage based on this change. This is data from a webapp with 3 instances:

image


src/Diagnostics.RuntimeHost/Extensions/MultiCloudExtensions.cs
```diff
@@ -26,5 +26,25 @@ public static bool IsAirGappedCloud(this IConfiguration configuration)
...
         }
+
+        /// <summary>
+        /// Determines if this cloud is mooncake
+        /// </summary>
+        /// <param name="configuration"></param>
+        /// <returns></returns>
+        public static bool IsMoonCakeCloud(this IConfiguration configuration)
+        {
+            return configuration.GetValue<string>("CloudDomain").Equals(DataProviderConstants.AzureChinaCloud, StringComparison.CurrentCultureIgnoreCase);
```
hforeste, line 37:
```suggestion
            return configuration.GetValue<string>("CloudDomain").Equals(DataProviderConstants.AzureChinaCloud, StringComparison.CurrentCultureIgnoreCase) || configuration.GetValue<string>("CloudDomain").Equals(DataProviderConstants.AzureChinaCloudCodename, StringComparison.CurrentCultureIgnoreCase);
```

src/Diagnostics.RuntimeHost/Extensions/MultiCloudExtensions.cs
```diff
@@ -26,5 +26,25 @@ public static bool IsAirGappedCloud(this IConfiguration configuration)
             return configuration.GetValue<string>("CloudDomain").Equals(DataProviderConstants.AzureUSSec, StringComparison.CurrentCultureIgnoreCase)
                 || configuration.GetValue<string>("CloudDomain").Equals(DataProviderConstants.AzureUSNat, StringComparison.CurrentCultureIgnoreCase);
         }
+
+        /// <summary>
+        /// Determines if this cloud is mooncake
+        /// </summary>
+        /// <param name="configuration"></param>
+        /// <returns></returns>
+        public static bool IsMoonCakeCloud(this IConfiguration configuration)
```
hforeste, line 35:
```suggestion
        public static bool IsAzureChinaCloud(this IConfiguration configuration)
```

src/Diagnostics.RuntimeHost/Extensions/MultiCloudExtensions.cs
```diff
@@ -26,5 +26,25 @@ public static bool IsAirGappedCloud(this IConfiguration configuration)
...
+        {
+            return configuration.GetValue<string>("CloudDomain").Equals(DataProviderConstants.AzureChinaCloud, StringComparison.CurrentCultureIgnoreCase);
+        }
+
+        /// <summary>
+        /// Determines if this cloud is US Gov (Fairfax)
+        /// </summary>
+        /// <param name="configuration"></param>
+        /// <returns></returns>
+        public static bool IsUSGovCloud(this IConfiguration configuration)
```
hforeste, line 45:
```suggestion
        public static bool IsAzureUSGovernment(this IConfiguration configuration)
```

src/Diagnostics.RuntimeHost/Extensions/MultiCloudExtensions.cs
```diff
@@ -26,5 +26,25 @@ public static bool IsAirGappedCloud(this IConfiguration configuration)
...
+        {
+            return configuration.GetValue<string>("CloudDomain").Equals(DataProviderConstants.AzureUSGovernment, StringComparison.CurrentCultureIgnoreCase);
```
hforeste, line 47:
```suggestion
            return configuration.GetValue<string>("CloudDomain").Equals(DataProviderConstants.AzureUSGovernment, StringComparison.CurrentCultureIgnoreCase) || configuration.GetValue<string>("CloudDomain").Equals(DataProviderConstants.AzureUSGovernmentCodename, StringComparison.CurrentCultureIgnoreCase);
```

ShekharGupta1988:
this is a great improvement @rekhaswaminathan . Very clear measurement of improvement.

ShekharGupta1988 (approved):

ShekharGupta1988:
@rekhaswaminathan - I think we should also set some sort of alerts if possible via portal or storage kusto to know about these throttling errors.

rekhaswaminathan:
> @rekhaswaminathan - I think we should also set some sort of alerts if possible via portal or storage kusto to know about these throttling errors.

Yup, I have a few already in place based on our kusto logs. But last time throttling happened, we didnt get any exception logged in DiagRole side. So I am planning to add more in AppInsights

PR #578, opened 2020-09-21
Deprecate observer api

hforeste:
Instead of erroring out the API with a 500 status code for using unsupported Observer APIs we will instead give a 400 bad request response. This is a more appropriate status code and will not impact our SLA.


src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -100,7 +100,15 @@ protected async Task<IActionResult> GetDetector(TResource resource, string detec
...
-            var detectorResponse = await GetDetectorInternal(detectorId, cxt);
+            Tuple<Response, List<DataProviderMetadata>> detectorResponse = null;
+            try
+            {
+                detectorResponse = await GetDetectorInternal(detectorId, cxt);
+            }
+            catch (Exception ex) when (ex.InnerException is ApiNotSupportedException)
+            {
+                return BadRequest(ex.Message);
+            }
```
ShekharGupta1988, line 111: I think the detector API should still return server error and not client error as client cannot do anything to correct the request. The issue is in the server using an unsupported dependent API call

hforeste: Hey Shekhar, yeah I can see what you mean but don't you think it is the job of the detector authors who should be concerned with that? As their clients are the end users and our clients are the detector authors themselves so it shouldn't impact platform SLA.

rekhaswaminathan (approved):

hforeste:
Hey @ShekharGupta1988 , looks like the 90,000+ calls that I mentioned earlier are actually calls made to Observer via browser

PR #589, opened 2020-10-26
add enums, gist and detector for WorkflowApp

milincjoshi7:
Onboarding WorkflowApp (Logic App V2) to AppLens as a new resource.


khaled-zayed:
@nmallick1 can you please validate this?

src/Diagnostics.RuntimeHost/Controllers/Site/SiteControllerBase.cs
```diff
@@ -79,6 +79,7 @@ private AppType GetApplicationType(string kind)
             else if (kindProperty.Contains("function")) return AppType.FunctionApp;
             else if (kindProperty.Contains("mobile")) return AppType.MobileApp;
             else if (kindProperty.Contains("gateway")) return AppType.GatewayApp;
+            else if (kindProperty.Contains("workflowapp")) return AppType.WorkflowApp;
```
cindylovescoding, line 82: Considering kindProperty for workflowApp could be something like "functionapp,workflowapp", which will fall into the logic at line 79. Should we move this condition before line 79?

nmallick1: @cindylovescoding , since we are allowing comma separated AppKind, do you think we should split the string by comma and then do a bitwise AND when returning the AppKind? The UI logic would have to be tweaked too.

Doing so will allow us to support different combinations in future without having to worry about the order in which the condition is  evaluated

nmallick1 (changes requested):

nmallick1 (approved):

cindylovescoding (approved):

milincjoshi7:
Closing PR as it is approved.

PR #598, opened 2020-11-05
Fetch app settings from keyvault and assign to site

rekhaswaminathan:

PR #600, opened 2020-11-06
Remove outbound calls to KeyVault for Prod and Staging

rekhaswaminathan:
Created separate branch for removing outbound keyvault calls. I will merge this branch to main once the devops pipeline is integrated with the [script ](https://github.com/Azure/Azure-AppServices-Diagnostics/pull/598)and tested. 

- For production and staging, skip outbound calls 

- @hforeste , Can you review this for N-clouds. From your email I learnt that N-clouds is using keyvault app setting reference so it shouldn't be affected.

### DIFF
--- a/src/Diagnostics.DataProviders/ConfigurationFactory.cs
+++ b/src/Diagnostics.DataProviders/ConfigurationFactory.cs
@@ -32,15 +32,19 @@ namespace Diagnostics.DataProviders
 
             var builtConfig = builder.Build();
 
-            var tokenProvider = new AzureServiceTokenProvider(azureAdInstance: builtConfig["Secrets:AzureAdInstance"]);
-            var keyVaultClient = new KeyVaultClient(
-                new KeyVaultClient.AuthenticationCallback(
-                    tokenProvider.KeyVaultTokenCallback
-                )
-            );
+            // For production and staging, skip outbound call to keyvault for AppSettings
+            if (env.IsDevelopment())
+            {
+                var tokenProvider = new AzureServiceTokenProvider(azureAdInstance: builtConfig["Secrets:AzureAdInstance"]);
+                var keyVaultClient = new KeyVaultClient(
+                    new KeyVaultClient.AuthenticationCallback(
+                        tokenProvider.KeyVaultTokenCallback
+                    )
+                );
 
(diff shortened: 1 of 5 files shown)


src/Diagnostics.DataProviders/ConfigurationFactory.cs
```diff
@@ -32,15 +32,19 @@ public AppSettingsDataProviderConfigurationFactory(IHostingEnvironment env)
...
-            var keyVaultClient = new KeyVaultClient(
-                new KeyVaultClient.AuthenticationCallback(
-                    tokenProvider.KeyVaultTokenCallback
-                )
-            );
-
-            string keyVaultConfig = Helpers.GetKeyvaultforEnvironment(env.EnvironmentName);
-            builder.AddAzureKeyVault(builtConfig[keyVaultConfig], keyVaultClient, new DefaultKeyVaultSecretManager());
+            // For production and staging, skip outbound call to keyvault for AppSettings
+            if (env.IsDevelopment())
```
hforeste, line 36: Should we also use the same boolean condition that you have in startup code? `builtConfig.GetValue<bool>("Secrets:KeyVaultEnabled", true) || hostingEnvironment.IsDevelopment()`

rekhaswaminathan: `Secrets:KeyVaultEnabled` - this config is present only in CompilerHost. So thats why I took this approach for runtimehost - If development include keyvault else skip the call.

hforeste: Okay. Just quick clarification for N-Clouds, I am using keyvault app settings reference specifically for observer in N-Clouds. Haven't transitioned to using that for runtimehost. You think we can include the `KeyVaultEnabled` flag as part of your condition? Then I'll just keep it enabled like the old way in N-clouds until I move to using keyvault appsetting references or the new approach we're taking in public.

rekhaswaminathan: Okay sounds good! Thanks for clarification @hforeste

hforeste (approved):

ShekharGupta1988 (approved):

PR #601, opened 2020-11-10
Keystone component

SharmaAjay19:
The addition of the Keystone component using res.AddKeystoneComponent() will add a Keystone Insight to the response of the detector.
image

The detector code to add keystone component looks like:
image

When a keystone insight is time sensitive, then we add the start date and expiry date like below:
image

Title and Summary are required attributes to provide in any keystone insight.
If any of them is missing or null, we provide the below exception message.
image

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/KeystoneComponent.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/KeystoneComponent.cs
@@ -6,28 +6,27 @@ using System.Text;
 
 namespace Diagnostics.ModelsAndUtils.Models.ResponseExtensions
 {
-    public class KeystoneInsightBase
-    {
-        public string LoggingName;
-        public string Title;
-        public string Summary;
-        public string Details;
-        public DateTime StartTime;
-        public DateTime EndTime;
-    }
-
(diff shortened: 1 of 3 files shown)


src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/KeystoneComponent.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    public static class ResponseKeystoneComponentExtension
+    {
+        public static DiagnosticData AddKeystoneComponent(this Response response, KeystoneInsight keystoneInsight)
+        {
+            try
+            {
+                if (!(keystoneInsight.Title != null && keystoneInsight.Summary != null))
```
ShekharGupta1988, line 33: perhaps check for IsNullOrWhitespace() ?

ShekharGupta1988 (approved):

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/KeystoneComponent.cs
SharmaAjay19: Yup, also will add a check for if expiry date is before current date.

PR #602, opened 2020-11-11
Create frontdoor-update.ps1

rekhaswaminathan:
Part of release pipeline, this script will be used to temporarily remove a backend host from receiving traffic to avoid any SLA dips. After swap is done, it will be reenabled. Currently devops does not have a frontdoor task, so using powershell script to do this. 

Proposed flow:

- Remove a backend from frontdoor
- Wait x secs for the change to be effective
- Update latest secrets from KeyVault as AppSettings to production slot
- Swap from staging to prod
- Wait for x secs for site to come back (May be send a /healthping call) 
- Add the backend back to frontdoor

### DIFF
--- a/infrastructure/Scripts/frontdoor-update.ps1
+++ b/infrastructure/Scripts/frontdoor-update.ps1
@@ -1,0 +1,51 @@
+# Parameters defined
+param(
+    [Parameter(Mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
+    [string]$Subscription,
+
+    [Parameter(Mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
+    [string]$FrontDoor,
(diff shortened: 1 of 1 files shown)


nmallick1:
Should we try updating our pipeline's swap task to use swap with preview instead of a direct Swap?

image

ShekharGupta1988 (approved):

hforeste (approved):

hforeste:
Along the lines with what @nmallick1  mentioned, should we hit the health check endpoints and dependencyChecks before swapping?

rekhaswaminathan:
> Along the lines with what @nmallick1 mentioned, should we hit the health check endpoints and dependencyChecks before swapping?

Yes, we can do that.

PR #610, opened 2020-12-01
Encrypt & Decrypt appsettings

rekhaswaminathan:
- In the release pipeline, we can create a file with encrypted values of the keyvault secrets and add to the build artifacts directory
- During preload, decrypt the values from the file and add to config. This removes the need for high number of environment variables (which prevents .NET processes to fail to startup due to known limitation) and avoids key vault dependency.
- Removed some duplicate logic in ConfigurationFactory for loading config.
- Added conditional check for PublicAzure for now. Development mode and N-clouds will follow the old flow.


infrastructure/Scripts/keyvaultsecrets-encrypted.ps1
```diff
@@ -0,0 +1,85 @@
...
+    [ValidateNotNullOrEmpty()]
+    [string]$Subscription,
+
+    [Parameter(Mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
+    [string]$VaultName,
+
+    [Parameter(Mandatory = $true)]
+    [ValidateNotNullOrEmpty()]
+    [string]$encryptionkey,
```
hforeste, line 12: How are we passing the key?

src/Diagnostics.RuntimeHost/Extensions/EncyptedConfigProviderExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class EncyptedConfigProviderExtensions
```
hforeste, line 12: Spell check: `Encrypted...`

src/Diagnostics.RuntimeHost/Extensions/EncyptedConfigProviderExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+            try
+            {
+                using (StreamReader reader = new StreamReader(EncryptedFile))
+                {
+                    string json = reader.ReadToEnd();
+                    var result = JsonConvert.DeserializeObject<Dictionary<string, string>>(json);
+
+                    foreach (KeyValuePair<string, string> item in result)
+                    {
+                        var plainText = DecryptSetting(item.Value);
```
hforeste, line 57: What happens if some weird failure comes while attempting to decrypt a bad value? Will we still add the other values to the configuration object?

infrastructure/Scripts/keyvaultsecrets-encrypted.ps1
rekhaswaminathan: Yes it will be a secret variable in ADO which we can pass as parameter to the script

src/Diagnostics.RuntimeHost/Extensions/EncyptedConfigProviderExtensions.cs
rekhaswaminathan: Good catch!

src/Diagnostics.RuntimeHost/Extensions/EncyptedConfigProviderExtensions.cs
rekhaswaminathan: Ideally it should skip the current setting and proceed to other good values. I will add code to handle this.

hforeste (approved):

src/Diagnostics.RuntimeHost/Extensions/EncryptedConfigProviderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+                        if (!string.IsNullOrEmpty(plainText))
+                        {
+                            config.Add(item.Key, plainText);
+                        }
+                    }
+
+                }
+            } catch (Exception ex)
+            {
+                DiagnosticsETWProvider.Instance.LogRuntimeHostMessage($"Exception in {nameof(EncryptedConfigProvider)} : {ex.ToString()}");
```
ShekharGupta1988, line 67: if Decryption of the setting fails, do we want to return gracefully? It might cause failures in downward streams.
Should we look to :
a) throw the exception here with exception details getting logged into kusto
b) let the process not start in this condition, so that health check/ front door etc can kick in and save our PROD SLA
c) put an alert on this above exception with TSG?

src/Diagnostics.RuntimeHost/Extensions/EncryptedConfigProviderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+                    aes.Key = Convert.FromBase64String(EncryptionKey);
+                    aes.IV = iv;
+                    aes.Padding = PaddingMode.PKCS7;
+                    ICryptoTransform decryptor = aes.CreateDecryptor(aes.Key, aes.IV);
+                    var unencryptedBytes = decryptor.TransformFinalBlock(buffer, 16, buffer.Length - 16);
+                    plainText = Encoding.UTF8.GetString(unencryptedBytes);
+                }
+            } catch (Exception ex)
+            {
+                DiagnosticsETWProvider.Instance.LogRuntimeHostMessage($"Exception while decrypting setting: {ex.ToString()}");
```
ShekharGupta1988, line 91: same here

src/Diagnostics.RuntimeHost/Extensions/EncryptedConfigProviderExtensions.cs
rekhaswaminathan: Yes, those steps make sense. We can implement those.

ShekharGupta1988 (approved):

PR #614, opened 2020-12-08
Redirect partner queries using ExecuteClusterQuery directly to the target cluster.

nmallick1:
Extract cluster and database name from the query and if it is AKS, connect to the AKS cluster instead of using the Antares cluster as a proxy. Once we ensure this is working, we can apply the logic for all partners wherever we find a cluster and database mentioned in the query.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
@@ -65,10 +65,8 @@ namespace Diagnostics.DataProviders
                     string targetCluster = element.Groups["cluster"].Value.Trim(new char[] { '\'', '\"' });
                     string targetDatabase = element.Groups["database"].Value.Trim(new char[] { '\'', '\"' });                    
 
-                    if (!string.IsNullOrWhiteSpace(targetCluster) && !string.IsNullOrWhiteSpace(targetDatabase)
-                        && targetCluster.Equals("Aks", StringComparison.OrdinalIgnoreCase) && targetDatabase.Equals("AKSprod", StringComparison.OrdinalIgnoreCase))
+                    if (!string.IsNullOrWhiteSpace(targetCluster) && !string.IsNullOrWhiteSpace(targetDatabase))
                     {
-                        //Apply this only for Aks at the moment.
                         return await ExecuteClusterQuery(query, targetCluster, targetCluster, requestId, operationName);
                     }
                 }
--- a/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
@@ -57,6 +57,23 @@ namespace Diagnostics.DataProviders
 
         public async Task<DataTable> ExecuteClusterQuery(string query, string requestId = null, string operationName = null)
         {
+            var matches = Regex.Matches(query, @"cluster\((?<cluster>([^\)]+))\).database\((?<database>([^\)]+))\)\.");
+            if (matches.Any())
+            {
+                foreach (Match element in matches)
+                {
+                    string targetCluster = element.Groups["cluster"].Value.Trim(new char[] { '\'', '\"' });
+                    string targetDatabase = element.Groups["database"].Value.Trim(new char[] { '\'', '\"' });                    
+
+                    if (!string.IsNullOrWhiteSpace(targetCluster) && !string.IsNullOrWhiteSpace(targetDatabase)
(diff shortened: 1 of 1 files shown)


src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -57,6 +57,23 @@ public KustoDataProvider(OperationDataCache cache, KustoDataProviderConfiguratio
...
+                foreach (Match element in matches)
+                {
+                    string targetCluster = element.Groups["cluster"].Value.Trim(new char[] { '\'', '\"' });
+                    string targetDatabase = element.Groups["database"].Value.Trim(new char[] { '\'', '\"' });                    
+
+                    if (!string.IsNullOrWhiteSpace(targetCluster) && !string.IsNullOrWhiteSpace(targetDatabase)
+                        && targetCluster.Equals("Aks", StringComparison.OrdinalIgnoreCase) && targetDatabase.Equals("AKSprod", StringComparison.OrdinalIgnoreCase))
+                    {
+                        //Apply this only for Aks at the moment.
+                        return await ExecuteClusterQuery(query, targetCluster, targetCluster, requestId, operationName);
```
hforeste, line 72: So if the query has a join where there are two statements with cluster and database and table information eg, <cluster>.<database>.<table> then we are executing the query twice?

nmallick1: Not really, the return along with the wait will cause the function to terminate as soon as the first match is found.

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -57,6 +57,23 @@ public KustoDataProvider(OperationDataCache cache, KustoDataProviderConfiguratio
...
+            var matches = Regex.Matches(query, @"cluster\((?<cluster>([^\)]+))\).database\((?<database>([^\)]+))\)\.");
+            if (matches.Any())
+            {
+                foreach (Match element in matches)
+                {
+                    string targetCluster = element.Groups["cluster"].Value.Trim(new char[] { '\'', '\"' });
+                    string targetDatabase = element.Groups["database"].Value.Trim(new char[] { '\'', '\"' });                    
+
+                    if (!string.IsNullOrWhiteSpace(targetCluster) && !string.IsNullOrWhiteSpace(targetDatabase)
+                        && targetCluster.Equals("Aks", StringComparison.OrdinalIgnoreCase) && targetDatabase.Equals("AKSprod", StringComparison.OrdinalIgnoreCase))
```
rekhaswaminathan, line 69: Will the cluster name for N-clouds?

nmallick1: Not really, people write query with their public cluster name and then we make the query cloud agnostic by looking up the cluster name from Kusto mapping configuration defined by teams.

rekhaswaminathan: I see. Thanks for the clarification!

hforeste, line 69: My only suggestion here is to put the cluster and database information in our runtimehost app settings. We don't want to reveal kusto URL information here to prevent future issues.

nmallick1: Good point, this however is an interim test for AKS alone, once we confirm that this change did not cause any significant errors in any of the existing detectors for this partner, we will apply this extraction logic to all partners and we won't have to store any strings anywhere and the auth on kusto cluster will prevent anything bad.

I do however get your point, would you suggest I apply it to all partner queries (given that I have manually tested a few of the queries for this partner) or run with this logic for a week and then remove the strings?

hforeste: I think that really depends on how confident you feel about applying it to all partners but if you choose to keep this logic and run it for a week then I believe we need be more slightly more secure and sensitive with partner kusto information..

hforeste: I think take your call but I would like to prevent the potential for DDoS on their cluster.

nmallick1: Made the change to apply this for all partners. We will have partners test their detectors on our staging environment to confirm we do not end up breaking them.

hforeste (approved):

PR #629, opened 2021-01-08
Disable continuous heartbeats to main kusto clusters

ShekharGupta1988:
### Problem:
- Our Heartbeat query count is very high on main kusto clusters and is deemed expensive. It is affecting other operations like wawsalerting etc.

### Solution:
1. Stop continuous pings to main kusto clusters.
2. Whenever failover is required from follower to main, run a heartbeat query on main cluster and if successful, then perform a failover. (Currently, we dont check anything on the main cluster when doing failover :) )
3. Kusto Dashboards will be impacted by this. However not a big thing as we will not see any synthetic SLA for main clusters. We will still continue to see synthetic SLA and failover telemetry for follower clusters.

### DIFF
--- a/src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
+++ b/src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
@@ -232,18 +232,6 @@ namespace Diagnostics.DataProviders
                 
                 string activityId = Guid.NewGuid().ToString();
                 await RunHeartBeatPrimary(activityId);
-
-                //if (string.IsNullOrWhiteSpace(FailoverCluster))
-                //{
-                //    await RunHeartBeatPrimary(activityId);
-                //}
-                //else
-                //{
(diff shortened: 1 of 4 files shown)


src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -182,12 +182,20 @@ private async Task RunHeartBeatPrimary(string activityId)
                 _ConsecutiveSuccessCount = 0;
             }
 
-            // if not in failover state
-            //  should failover?
+            /*
+             * Logic of failing over to failover cluster:
```
rekhaswaminathan, line 186: By failover cluster - you mean the main cluster?

ShekharGupta1988: correct. 
The way this code is written is primary ==> follower, and Failover ==> main cluster

rekhaswaminathan (approved):

hforeste (approved):

PR #638, opened 2021-02-05
Remove the httpClientHandler to limit connections per server and inst…

nmallick1:
Removed the httpClientHandler maxConnectionsPerServer as it may lead to a deadlock and thread starvation. We can introduce it back once we move to .NET Core 3.

Based on my tests, I observed when timeout is enforced via cancellation tokens, it keeps the number of threads lower than when the timeout property is used. Cancellation token will also enforce the timeout but somehow it seems better and managing number of threads.

https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/HttpClient.cs

image

Also added a sleep of 1 second so that we do not create connections too aggressively, as we are timing out the connections after 3 second, a RST will be sent on the channel and SLB will be able to reclaim these connections a lot quicker. Taking a pause "might" just allow us to avoid SNAT. 

Although I wasn't able to repro the deadlock scenario we encountered, these changes should help.

### DIFF
--- a/src/Diagnostics.RuntimeHost/Services/HealthCheckService/HealthCheckService.cs
+++ b/src/Diagnostics.RuntimeHost/Services/HealthCheckService/HealthCheckService.cs
@@ -53,13 +53,13 @@ namespace Diagnostics.RuntimeHost.Services
             this.cache = cache;
         }
 
-        private Task<HttpResponseMessage> Get(HttpRequestMessage request)
+        private async Task<HttpResponseMessage> Get(HttpRequestMessage request)
         {
             //Sleep for a while so that we do not create outbound connections too aggressively causing SNAT port exhaustion.
-            System.Threading.Thread.Sleep(TimeSpan.FromSeconds(1));
+            await Task.Delay(TimeSpan.FromSeconds(1));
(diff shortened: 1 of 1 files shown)


src/Diagnostics.RuntimeHost/Services/HealthCheckService/HealthCheckService.cs
```diff
@@ -54,7 +55,12 @@ public HealthCheckService(IConfiguration Configuration, ISourceWatcherService so
 
         private Task<HttpResponseMessage> Get(HttpRequestMessage request)
         {
-            return _httpClient.SendAsync(request);
+            //Sleep for a while so that we do not create outbound connections too aggressively causing SNAT port exhaustion.
+            System.Threading.Thread.Sleep(TimeSpan.FromSeconds(1));
```
ShekharGupta1988, line 59: await Task.Delay ?

ShekharGupta1988 (approved):

src/Diagnostics.RuntimeHost/Services/HealthCheckService/HealthCheckService.cs
nmallick1: of course stupid me :)

PR #640, opened 2021-02-11
Onboarding static sites

khaled-zayed:


data/templates/Detector_StaticSites.csx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+
+[ArmResourceFilter(provider: "Microsoft.Web", resourceTypeName: "staticSites")]
+[Definition(Id = "YOUR_DETECTOR_ID", Name = "", Author = "YOUR_ALIAS", Description = "")]
+public async static Task<Response> Run(DataProviders dp, OperationContext<ArmResource> cxt, Response res)
+{
+    res.Dataset.Add(new DiagnosticData()
+    {
+        Table = await dp.Kusto.ExecuteClusterQuery(GetQuery(cxt), "KUSTO_CLUSTER_NAME", "KUSTO_DB_NAME", null, "GetQuery"),
```
nmallick1, line 23: static apps is Antares, shouldn't they use ExecuteQuery syntax?

khaled-zayed: Good catch, updated 😊

nmallick1 (approved):

data/templates/Detector_StaticSites.csx
ShekharGupta1988: But one thing to note here is - this is onboarded as an ARM Resource. So the Operation Context will not have the stamp in it. Hence the detector authors will have to figure out the stamp of the site (using observer) before invoking ExecuteQuery(..) method.

data/templates/Detector_StaticSites.csx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+
+[ArmResourceFilter(provider: "Microsoft.Web", resourceTypeName: "staticSites")]
+[Definition(Id = "YOUR_DETECTOR_ID", Name = "", Author = "YOUR_ALIAS", Description = "")]
+public async static Task<Response> Run(DataProviders dp, OperationContext<ArmResource> cxt, Response res)
+{
+    res.Dataset.Add(new DiagnosticData()
+    {
+        Table = await dp.Kusto.ExecuteQuery(GetQuery(cxt), cxt.Resource.Stamp.Name, null, "GetQuery"),
```
ShekharGupta1988, line 23: cxt.Resource will not have Stamp property here

ShekharGupta1988: I suggest in the template, we can add a line to fetch the stamp name via observer

hforeste: From what I learned, staticwebsites does not have a concept of stamp. Our infrastructure where we determine kusto cluster by stampname will not apply for staticwebsites.

khaled-zayed: Let's confirm with Mitch then first, thank you all.

PR #642, opened 2021-02-18
Add notification rendering for risk alert and notification

cindylovescoding:


nmallick1 (approved):

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Notification.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    public static class ResponseNotificationsExtension
+    {
+        public static DiagnosticData AddNotification(this Response response, Notification notification)
+        {
+            try
+            {
+                if (notification.Status == null || string.IsNullOrWhiteSpace(notification.Title))
+                {
+                    throw new Exception("Required attributes Status and Title cannot be null or empty for Notification.");
```
ShekharGupta1988, line 83: Lets throw Argument Exception

ShekharGupta1988 (approved):

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Notification.cs
```diff
@@ -0,0 +1,135 @@
...
+        /// <param name="isExpanded">Whether a notification is expanded to begin with.</param>
+        /// <param name="startDate">The startdate when the notification will be shown.</param>
+        /// <param name="expiryDate">The end date to stop showing the notification.</param>
+        public Notification(InsightStatus status, string title, string description = "", Solution solution = null, bool isExpanded = true, DateTime? startDate = null, DateTime? expiryDate = null)
+        {
+            Status = status;
+            Title = title;
+            Description = description;
+            Solution = solution;
+            IsExpanded = isExpanded;
```
ShekharGupta1988, line 69: also, how do you plan to use isExpanded for notification? My understanding is this will show as a notification bar on the top

cindylovescoding: This rendering is not for the message bar Shekhar. It's the rendering for the detailed notification in the side panel:
I am still debating about whether we should have this option or not:

image

XiaoxuHU (approved):

PR #645, opened 2021-02-24
Add retry for Geomaster

XiaoxuHU:
I am trying to add retry for GeoMaster as for [WI 1657](https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/1657), for now I add a similar logic as we implemented for Kusto.

But didn't find a pretty generic solution to extend across different data providers yet, since some DPs might have edge cases like for last retry or when to stop retry.   

Maybe folks can raise some ideas about how to extend retry logic for different DPs in a more generic way in this PR

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
+++ b/src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
@@ -30,5 +30,11 @@
         /// </summary>
         [ConfigurationName("GeoCertSubjectName")]
         public string GeoCertSubjectName { get; set; }
+
+        [ConfigurationName("Retry:MaxRetryCount")]
+        public int MaxCountRetry { get; set; }
+
+        [ConfigurationName("Retry:RetryDelayInSeconds")]
+        public int RetryDelayInSeconds { get; set; }
     }
 }
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -623,63 +623,97 @@ namespace Diagnostics.DataProviders
             var query = SitePathUtility.CsmAnnotateQueryString(queryString, apiVersion);
             var uri = new Uri(_geoMasterClient.BaseUri, path + query);
             HttpResponseMessage response = null;
-
-            try
+            R value = default;
+            int attempt = 0;
+            var exceptions = new List<Exception>();
+            string source = "GeoMasterDataProvider";
+            Exception attemptException = null;
+            do
             {
-                using (var request = new HttpRequestMessage(method, uri))
+                try
                 {
-                    try
+                    using (var request = new HttpRequestMessage(method, uri))
                     {
-                        if (method == HttpMethod.Post || method == HttpMethod.Put)
+                        try
                         {
-                            request.Content = new StringContent(content, Encoding.UTF8, "application/json");
-                        }
+                            DiagnosticsETWProvider.Instance.LogRetryAttemptMessage(
+                                RequestId,
+                                source,
+                                $"Starting Attempt: {attempt}"
+                            );
 
-                        var token = _geoMasterClient.AuthenticationToken;
-                        if (!string.IsNullOrWhiteSpace(token))
-                        {
-                            request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", token);
+                            if (method == HttpMethod.Post || method == HttpMethod.Put)
+                            {
+                                request.Content = new StringContent(content, Encoding.UTF8, "application/json");
+                            }
+
+                            var token = _geoMasterClient.AuthenticationToken;
+                            if (!string.IsNullOrWhiteSpace(token))
+                            {
+                                request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", token);
+                            }
+                            DiagnosticsETWProvider.Instance.LogDataProviderMessage(RequestId, $"{nameof(GeoMasterDataProvider.PerformHttpRequest)}", $"Making HTTP call to GeoMaster URI: {uri.AbsoluteUri} Method: {method.Method} ");
+                            response = await _geoMasterClient.Client.SendAsync(request, cancellationToken).ConfigureAwait(false);
+                            if (!response.IsSuccessStatusCode && response.Content != null)
+                            {
+                                string responseMessage = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+                                DiagnosticsETWProvider.Instance.LogDataProviderMessage(RequestId, $"{nameof(GeoMasterDataProvider.PerformHttpRequest)}", $"HTTP call to GeoMaster failed with response: {responseMessage}");
+                            }
+                            response.EnsureSuccessStatusCode();
+                            attemptException = null;
                         }
-                        DiagnosticsETWProvider.Instance.LogDataProviderMessage(RequestId, $"{nameof(GeoMasterDataProvider.PerformHttpRequest)}", $"Making HTTP call to GeoMaster URI: {uri.AbsoluteUri} Method: {method.Method} ");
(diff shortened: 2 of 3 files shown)


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -623,63 +623,97 @@ private async Task<R> PerformHttpRequest<R>(HttpMethod method, string path, stri
...
             var uri = new Uri(_geoMasterClient.BaseUri, path + query);
             HttpResponseMessage response = null;
-
-            try
+            R value = default;
+            int attempt = 0;
+            var exceptions = new List<Exception>();
+            string source = "GeoMasterDataProvider";
+            Exception attemptException = null;
+            do
```
ShekharGupta1988, line 631: how about using :
https://github.com/Azure/Azure-AppServices-Diagnostics/blob/main/src/Diagnostics.RuntimeHost/Utilities/RetryHelper.cs

Unless the retry attempts have any special logic in between (like kusto) and are simply plain retries, this is a generic Retry Helper we have to retry a function.

ShekharGupta1988: For kusto, we couldn't use this as there was some special logic to switch to leader cluster before the last retry attempt, but see if you can use it here.

XiaoxuHU: Thank you, @ShekharGupta1988 ! Let me replace my logic with RetryHelper

XiaoxuHU: Do you think we can use default value of **maxRetry** and **retryDelayInMs** or we need to specify these two values in AppSetting for some DPs ?

ShekharGupta1988: Its good to have Data Provider specific retry settings as well (in case we need to adjust the retry limits based on every data provider)

XiaoxuHU: @ShekharGupta1988 I updated for using RetryHelper with retry setting, please check for the Geomaster change. If it's okay, I will move to other data providers.

BTW, which data providers we want to add retry logic for ?

src/Diagnostics.DataProviders/DataProviderConfigurations/GeoMasterDataProviderConfiguration.cs
```diff
@@ -30,5 +30,15 @@ public GeoMasterDataProviderConfiguration()
         /// </summary>
         [ConfigurationName("GeoCertSubjectName")]
         public string GeoCertSubjectName { get; set; }
+        /// <summary>
+        /// Max Retry Count
+        /// </summary>
+        [ConfigurationName("Retry:MaxRetryCount")]
+        public int MaxRetryCount { get; set; }
```
hforeste, line 37: Hey @XiaoxuHU , I'd recommend setting these retry configuration settings at the base class `DataProviderConfigurationBase` so that you don't have to have duplicate code everywhere and less prone to future errors for these configuration values. https://github.com/Azure/Azure-AppServices-Diagnostics/blob/main/src/Diagnostics.DataProviders/DataProviderConfigurations/DataProviderConfigurationBase.cs

XiaoxuHU: Thank you, @hforeste ! I  updated code to move common settings into base configuration

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -608,22 +611,35 @@ public Task<T> InvokeDaasExtension<T>(string subscriptionId, string resourceGrou
...
             var body = JsonConvert.SerializeObject(content);
-            return PerformHttpRequest<R>(HttpMethod.Post, path, queryString, body, apiVersion, cancellationToken);
+            //return PerformHttpRequest<R>(HttpMethod.Post, path, queryString, body, apiVersion, cancellationToken);
+            return PerformHttpRequestWithRetry<R>(HttpMethod.Post, path, queryString, body, apiVersion, cancellationToken);
         }
 
+        private Task<R> PerformHttpRequestWithRetry<R>(HttpMethod method, string path, string queryString, string content, string apiVersion, CancellationToken cancellationToken)
+        {
+            //Save params  into tuple temporarly to so to call PerformHttpRequest method without passing params for using RetryHelper
+            HttpRequestTuple = new Tuple<HttpMethod, string, string, string, string, CancellationToken>(method, path, queryString, content, apiVersion, cancellationToken);
```
ShekharGupta1988, line 628: is this multiple execution safe approach?
what would happen if in the same detector, I do the following at the same time:

<pre>
var task1 = dp.Geomaster.Execute..Reuest(method1, path1, .....);
var task2 = dp.Geomaster.Execute..Reuest(method2, path2, .....);

await Task.WhenAll(task1, task2);
</pre>

Since for one detector API call, we only create one object of each data provider (as far as I remember), isnt there a possibility that this HttpRequestTuple is not thread safe and will get overwritten

XiaoxuHU: Good point, do you think it will be more thread safe if I change the current logic to store task id with tuple into a concurrrent dictionary. Once task is done, remove entry from dictionary ?

hforeste: Hey @XiaoxuHU , if I am understanding this correctly. The reason we're going with the Tuple approach is because `RetryHelper` currently does not accept parameters for the `Func` that we pass?

XiaoxuHU: Yes, @hforeste. You are right.
Basically, I tried to use tuple is to make `Func` in `RetryHelper`  more generic since different data providers have different parmas for their method which pass into `Func`. 

But seems like it might be got overwritten since Tuple is not thread safe. I am thinking of using ConcurrentHashMap to store an id and tuple params for each func instead of tuple to make it safe. But one thing it is that which stuff can be the id. I have tried the task id where task is the return value after execute Func in RetryHelper. But this approach is not working since I am not able to get task id before func get invoked(I need to get id before func get invoke in RetryHelper so to put id and tuple into map) . Also I tried Func hashcode, but if call same function twice with different params, their hash code will be the same and map value will get overwritten.

Any ideas on how I can fix this issue, or do you think any other approaches I can implement ?

Thanks

XiaoxuHU: @ShekharGupta1988 and @hforeste , I updated code to use Concurrent Dictionary to store arguments from `Func` which goes to RetryHelper for thread safe

hforeste: > Any ideas on how I can fix this issue, or do you think any other approaches I can implement ?

Hey @XiaoxuHU , one approach is creating overloaded methods of that RetryHelper that take Func with parameters. .NET does this with Func https://docs.microsoft.com/en-us/dotnet/api/system.func-2?view=net-5.0 
image

XiaoxuHU: > > Any ideas on how I can fix this issue, or do you think any other approaches I can implement ?
> 
> Hey @XiaoxuHU , one approach is creating overloaded methods of that RetryHelper that take Func with parameters. .NET does this with Func https://docs.microsoft.com/en-us/dotnet/api/system.func-2?view=net-5.0
> image

@hforeste, the thing is for `Func` in RetryHelper is we only know its return type is `TResult` but we don't know exact types for **input prams** since different DPs will pass different signature of their functions got invoked in RetryHelper

XiaoxuHU:
Close this PR due to I unfolked repo and not able to push commit...

PR #647, opened 2021-02-26
Add ExecuteOnAllKustoCluster method

nmallick1:
Provide a way to execute a query across all of our kusto clusters as All has a heavy toll
throw a compiler error if we detect the use of All in any of the Antares detectors.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
@@ -101,7 +101,7 @@ namespace Diagnostics.DataProviders
             return await _kustoClient.ExecuteQueryAsync(Helpers.MakeQueryCloudAgnostic(_kustoMap, query), _kustoMap.MapCluster(cluster) ?? cluster, _kustoMap.MapDatabase(_configuration.DBName) ?? _configuration.DBName, requestId, operationName);
         }
 
-        public async Task<DataTable> ExecuteQueryOnAllAppAppServiceClusters(string query, string requestId = null, string operationName = null)
+        public async Task<DataTable> ExecuteQueryOnAllAppAppServiceClusters(string query, string operationName)
         {
             if (string.IsNullOrWhiteSpace(operationName))
             {
@@ -111,39 +111,38 @@ namespace Diagnostics.DataProviders
             {
                 throw new ArgumentNullException(nameof(query), "Query cannot be empty. Please supply a query to execute.");
             }
+
+            List<Task<DataTable>> queryTask = new List<Task<DataTable>>();
+
+            if(IsPublicCloud)
+            {
+                queryTask.Add(ExecuteQuery(query, "waws-prod-bay-153", null, operationName));
+                queryTask.Add(ExecuteQuery(query, "waws-prod-blu-189", null, operationName));
+                queryTask.Add(ExecuteQuery(query, "waws-prod-dm1-187", null, operationName));
+                queryTask.Add(ExecuteQuery(query, "waws-prod-am2-329", null, operationName));
+                queryTask.Add(ExecuteQuery(query, "waws-prod-db3-169", null, operationName));
+                queryTask.Add(ExecuteQuery(query, "waws-prod-hk1-029", null, operationName));
+            }
             else
             {
-                List<Task<DataTable>> queryTask = new List<Task<DataTable>>();
-
-                if(IsPublicCloud)
-                {
-                    queryTask.Add(ExecuteQuery(query, "waws-prod-bay-153", null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|WestUS"));
-                    queryTask.Add(ExecuteQuery(query, "waws-prod-blu-189", null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|EastUS"));
-                    queryTask.Add(ExecuteQuery(query, "waws-prod-dm1-187", null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|CentralUS"));
-                    queryTask.Add(ExecuteQuery(query, "waws-prod-am2-329", null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|WestEurope"));
-                    queryTask.Add(ExecuteQuery(query, "waws-prod-db3-169", null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|NorthEurope"));
-                    queryTask.Add(ExecuteQuery(query, "waws-prod-hk1-029", null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|EastAsia"));
-                }
-                else
-                {
-                    queryTask.Add(ExecuteQuery(query, DataProviderConstants.FakeStampForAnalyticsCluster, null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|{_configuration.CloudDomain}"));
-                }
-
-                var queryResult = await Task.WhenAll(queryTask.ToArray());
-                DataTable mergedTable = new DataTable();
-                foreach (DataTable dt in queryResult)
-                {
-                    if (dt.Rows.Count > 0)
-                    {
-                        if(dt.Rows.Count > 3000)
-                        {
-                            throw new Exception($"Query {operationName} returned more than 3000 rows. Please modify the query to fetch fewer rows while running across all app service clusters.");
-                        }
-                        mergedTable.Merge(dt, false, MissingSchemaAction.Add);
-                    }
-                }
-                return mergedTable;
+                queryTask.Add(ExecuteQuery(query, DataProviderConstants.FakeStampForAnalyticsCluster, null, operationName));
             }
+
+            var queryResult = await Task.WhenAll(queryTask.ToArray());
+            DataTable mergedTable = new DataTable();
+            foreach (DataTable dt in queryResult)
+            {
+                if (dt.Rows.Count > 0)
+                {
+                    if(dt.Rows.Count > 3000)
+                    {
+                        throw new Exception($"Query {operationName} returned more than 3000 rows. Please modify the query to fetch fewer rows while running across all app service clusters.");
+                    }
+                    mergedTable.Merge(dt, false, MissingSchemaAction.Add);
+                }
+            }
+            return mergedTable;
+
         }
 
         public Task<KustoQuery> GetKustoClusterQuery(string query)
--- a/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
@@ -55,7 +55,7 @@ namespace Diagnostics.DataProviders
             };
         }
 
-        private bool isPublicCloud
+        private bool IsPublicCloud
         {
             get {
                 var publicClouds = new string[] { DataProviderConstants.AzureCloud, DataProviderConstants.AzureCloudAlternativeName };
@@ -101,7 +101,7 @@ namespace Diagnostics.DataProviders
             return await _kustoClient.ExecuteQueryAsync(Helpers.MakeQueryCloudAgnostic(_kustoMap, query), _kustoMap.MapCluster(cluster) ?? cluster, _kustoMap.MapDatabase(_configuration.DBName) ?? _configuration.DBName, requestId, operationName);
         }
 
-        public async Task<DataTable> ExecuteQueryOnAllAntaresClusters(string query, string requestId = null, string operationName = null)
+        public async Task<DataTable> ExecuteQueryOnAllAppAppServiceClusters(string query, string requestId = null, string operationName = null)
         {
             if (string.IsNullOrWhiteSpace(operationName))
             {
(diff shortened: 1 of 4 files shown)


src/Diagnostics.Scripts/EntityInvoker.cs
```diff
@@ -369,6 +369,21 @@ private void Validate()
                 {
                     throw new ScriptCompilationException("Detector is marked with both SystemFilter and ResourceFilter. System Invoker should not include any ResourceFilter attribute.");
                 }
+
+                if(this._entityMetaData.ScriptText.Contains("All('") &&
```
puneetg1983, line 373: >All(' [](start = 61, length = 5)

What if someone uses All(" ... Its rare but can happen

src/Diagnostics.Scripts/EntityInvoker.cs
```diff
@@ -369,6 +369,21 @@ private void Validate()
...
+                    || this._resourceFilter.ResourceType == ResourceType.AppServiceCertificate
+                    || this._resourceFilter.ResourceType == ResourceType.AppServiceDomain))
+                {
+                    this.CompilationOutput = this.CompilationOutput.Concat(new string[]
+                           {
+                                "ERROR : Use of All('TableName') kusto function is not supported.",
+                                "Valid Options:",
+                                "- Use dp.Kusto.ExecuteQueryOnAllAntaresClusters(string query, string requestId = null, string operationName = null) instead."
+                           });
+                    throw new ScriptCompilationException(string.Empty);
```
puneetg1983, line 385: >string.Empty [](start = 57, length = 12)

Might want to give some message instead of empty

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -93,6 +101,47 @@ public async Task<DataTable> ExecuteQuery(string query, string stampName, string
...
+            {
+                throw new ArgumentNullException(nameof(query), "Query cannot be empty. Please supply a query to execute.");
+            }
+            else
+            {
+                List<Task<DataTable>> queryTask = new List<Task<DataTable>>();
+
+                if(isPublicCloud)
+                {
+                    queryTask.Add(ExecuteQuery(query, "waws-prod-bay-153", null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|WestUS"));
```
puneetg1983, line 120: >"waws-prod-bay-153 [](start = 54, length = 18)

Any reason for choosing these numbers or they are just random stamp names?

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -55,6 +55,14 @@ public KustoDataProvider(OperationDataCache cache, KustoDataProviderConfiguratio
             };
         }
 
+        private bool isPublicCloud
```
puneetg1983, line 58: >isPublicCloud [](start = 21, length = 13)

Should be IsPublicCloud per C# convention

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -93,6 +101,47 @@ public async Task<DataTable> ExecuteQuery(string query, string stampName, string
...
+                var queryResult = await Task.WhenAll(queryTask.ToArray());
+                DataTable mergedTable = new DataTable();
+                foreach (DataTable dt in queryResult)
+                {
+                    if (dt.Rows.Count > 0)
+                    {
+                        mergedTable.Merge(dt, false, MissingSchemaAction.Add);
+                    }
+                }
+                return mergedTable;
```
puneetg1983, line 141: >                 [](start = 35, length = 16)

nit:empty space

src/Diagnostics.DataProviders/Interfaces/IKustoDataProvider.cs
```diff
@@ -11,6 +11,8 @@ public interface IKustoDataProvider : IMetadataProvider
 
         Task<DataTable> ExecuteQuery(string query, string stampName, string requestId = null, string operationName = null);
 
+        Task<DataTable> ExecuteQueryOnAllAntaresClusters(string query, string requestId = null, string operationName = null);
```
puneetg1983, line 14: >ExecuteQueryOnAllAntaresClusters [](start = 24, length = 32)

Is ExecuteQueryOnAppAppServiceClusters better? This is also fine though. Just came to my mind so thought of mentioning.

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -93,6 +101,47 @@ public async Task<DataTable> ExecuteQuery(string query, string stampName, string
...
+                    queryTask.Add(ExecuteQuery(query, DataProviderConstants.FakeStampForAnalyticsCluster, null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|{_configuration.CloudDomain}"));
+                }
+
+                var queryResult = await Task.WhenAll(queryTask.ToArray());
+                DataTable mergedTable = new DataTable();
+                foreach (DataTable dt in queryResult)
+                {
+                    if (dt.Rows.Count > 0)
+                    {
+                        mergedTable.Merge(dt, false, MissingSchemaAction.Add);
```
puneetg1983, line 138: >mergedTable.Merge [](start = 24, length = 17)

We may end up spiking RuntimeHost CPU if someone writes a really bad query which has a lot of records. Do you want to put some hard limits here? Like 10K records or 100K records?

puneetg1983 (approved):
:shipit:

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
nmallick1: Does 3K rows per cluster sound reasonable or do you think the limit should be even lower. I am thinking 3K since internally we allow people to run query for a 3 day duration.

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
nmallick1: Removed

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
nmallick1: Totally random :)

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
puneetg1983: Yea sounds fine. We can bump it or lower it once we know how this behaves

nmallick1: @puneetg1983 , I looked up the statistics and I do not see any query in the past 2 days crossing 2K rows when using All, I believe 3K limit should be good to avoid queries from failing however like you said, we need to keep an eye on the CPU and adjust the limit accordingly.

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -93,6 +101,51 @@ public async Task<DataTable> ExecuteQuery(string query, string stampName, string
...
+                    queryTask.Add(ExecuteQuery(query, DataProviderConstants.FakeStampForAnalyticsCluster, null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|{_configuration.CloudDomain}"));
+                }
+
+                var queryResult = await Task.WhenAll(queryTask.ToArray());
+                DataTable mergedTable = new DataTable();
+                foreach (DataTable dt in queryResult)
+                {
+                    if (dt.Rows.Count > 0)
+                    {
+                        if(dt.Rows.Count > 3000)
```
hforeste, line 138: Hey @nmallick1 , how did you decide 3000 would be a threshold? Can we also add a configuration value for this and you can make a default value of 3000 in the app settings

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -93,6 +101,51 @@ public async Task<DataTable> ExecuteQuery(string query, string stampName, string
             return await _kustoClient.ExecuteQueryAsync(Helpers.MakeQueryCloudAgnostic(_kustoMap, query), _kustoMap.MapCluster(cluster) ?? cluster, _kustoMap.MapDatabase(_configuration.DBName) ?? _configuration.DBName, requestId, operationName);
         }
 
+        public async Task<DataTable> ExecuteQueryOnAllAppAppServiceClusters(string query, string requestId = null, string operationName = null)
+        {
+            if (string.IsNullOrWhiteSpace(operationName))
+            {
+                throw new ArgumentNullException(nameof(operationName), "OperationName cannot be empty. Please supply a name to identify the query.");
```
hforeste, line 108: Make operationName a required parameter since you will throw ArgumentNullException if it doesn't exist

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -93,6 +101,51 @@ public async Task<DataTable> ExecuteQuery(string query, string stampName, string
...
+            }
+            if (string.IsNullOrWhiteSpace(query))
+            {
+                throw new ArgumentNullException(nameof(query), "Query cannot be empty. Please supply a query to execute.");
+            }
+            else
```
hforeste, line 114: nitpik: this else body is unnecessary

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -93,6 +101,51 @@ public async Task<DataTable> ExecuteQuery(string query, string stampName, string
...
+            {
+                List<Task<DataTable>> queryTask = new List<Task<DataTable>>();
+
+                if(IsPublicCloud)
+                {
+                    queryTask.Add(ExecuteQuery(query, "waws-prod-bay-153", null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|WestUS"));
```
hforeste, line 120: Hey @nmallick1 , what do you think about iterating over our clusters using the Kusto configuration `KustoClusterNameGroupings` and running the `ExecuteClusterQuery` kind of like this:

```
foreach (var cluster in _configuration.KustoClusterNameGroupings.Split(new char[]{','})){
    queryTask.Add(ExecuteClusterQuery(query, cluster, "wawsprod", null, $"ExecuteQueryOnAllAntaresClusters|{operationName}|{cluster}"))
}
```

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
nmallick1: I did a show query analysis for all queries that contain All for the past 7 days, the max rows returned was around 2K. I'm thinking 3K should be fine and anything more is actually too much.

Do you think think we'd need to reconfigure this frequently? I know it's cleaner but do we really need a config setting for this? 

If we do decide to introduce a rows limit via config setting, it should apply to all queries executed against Kusto not just this specific code path.

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -93,6 +101,51 @@ public async Task<DataTable> ExecuteQuery(string query, string stampName, string
...
+            }
+            if (string.IsNullOrWhiteSpace(query))
+            {
+                throw new ArgumentNullException(nameof(query), "Query cannot be empty. Please supply a query to execute.");
+            }
+            else
+            {
+                List<Task<DataTable>> queryTask = new List<Task<DataTable>>();
+
+                if(IsPublicCloud)
```
hforeste, line 118: Instead of using the cloudenvironment as a configuration to determine if this should be enabled or not can we create an enable flag? Kind of like https://github.com/Azure/Azure-AppServices-Diagnostics/blob/b688b57942342d035a6c9bb3fa7a65c259159168/src/Diagnostics.RuntimeHost/appsettings.json#L60-L61

This will enable us to:
1. Disable in production in the case something weird happens.
2. Enable/Disable anywhere in sovereign clouds when we want.

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -55,6 +55,14 @@ public KustoDataProvider(OperationDataCache cache, KustoDataProviderConfiguratio
             };
         }
 
+        private bool IsPublicCloud
+        {
+            get {
+                var publicClouds = new string[] { DataProviderConstants.AzureCloud, DataProviderConstants.AzureCloudAlternativeName };
+                return publicClouds.Any(s => _configuration.CloudDomain.Equals(s, StringComparison.CurrentCultureIgnoreCase));
+            }
+        }
```
hforeste, line 64: You may be able to remove this, see comment on line 118

hforeste (changes requested):
Left some comments

hforeste (approved):

PR #651, opened 2021-03-16
Log Runtimehost Auto-heal settings

ShekharGupta1988:
User Story 1648: Alert when auto heal is turned off on our Runtimehost PROD endpoints

The PR contains:
- Check AppHost config at regular intervals (30 mins), read the auto heal settings and log it in kusto
- Added new Events for such monitoring requirements

### DIFF
--- a/src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
+++ b/src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
@@ -775,5 +775,50 @@ namespace Diagnostics.Logger
         }
 
         #endregion AzureStorage Events (ID Range : 5500 - 5599)
+
+        #region Generic Monitoring Events (ID : 5600 - 5609)
+
+        /// <summary>
+        /// Log Monitoring Event Message
+        /// </summary>
+        /// <param name="Source">Monitoring Source</param>
+        /// <param name="Message">Event Message</param>
+        /// <param name="DiagEnvironment">Diag Environment</param>
+        /// <param name="DiagWebsiteHostName">Diag Hostname</param>
+        [Event(5600, Level = EventLevel.Informational, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogMonitoringEventMessage)]
+        public void LogMonitoringEventMessage(string Source, string Message, string DiagEnvironment = null, string DiagWebsiteHostName = null)
+        {
+            WriteDiagnosticsEvent(5600, Source, Message, EnvironmentName, WebsiteHostName);
+        }
+
+        /// <summary>
+        /// Log Monitoring Event Warning
+        /// </summary>
+        /// <param name="Source">Monitoring Source</param>
+        /// <param name="Message">Event Message</param>
+        /// <param name="DiagEnvironment">Diag Environment</param>
+        /// <param name="DiagWebsiteHostName">Diag Hostname</param>
+        [Event(5601, Level = EventLevel.Warning, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogMonitoringEventWarning)]
+        public void LogMonitoringEventWarning(string Source, string Message, string DiagEnvironment = null, string DiagWebsiteHostName = null)
+        {
+            WriteDiagnosticsEvent(5601, Source, Message, EnvironmentName, WebsiteHostName);
+        }
+
+        /// <summary>
+        /// Log Monitoring Event Exception
+        /// </summary>
+        /// <param name="Source">Monitoring Source</param>
+        /// <param name="Message">Event Message</param>
+        /// <param name="ExceptionType">Exception Type</param>
+        /// <param name="ExceptionDetails">Exception Details</param>
+        /// <param name="DiagEnvironment">Diag Environment</param>
+        /// <param name="DiagWebsiteHostName">Diag hostname</param>
+        [Event(5602, Level = EventLevel.Error, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogMonitoringEventException)]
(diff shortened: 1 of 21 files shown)


src/Diagnostics.RuntimeHost/Utilities/CredentialTapper.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System;
+
+namespace Diagnostics.RuntimeHost.Utilities
+{
+    /// <summary>
+    /// Traps potential creds and obfuscate them.
+    /// </summary>
+    public static class CredentialTrapper
```
puneetg1983, line 8: >CredentialTrapper [](start = 24, length = 17)

This is a bit unrelated to this PR but we have similar code in GeoMasterDataProvider.cs. Can you update that class also to use CredentialTapper?

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -775,5 +775,50 @@ public void LogAzureStorageException(string Source, string Message, string Excep
...
+        /// Log Monitoring Event Exception
+        /// </summary>
+        /// <param name="Source">Monitoring Source</param>
+        /// <param name="Message">Event Message</param>
+        /// <param name="ExceptionType">Exception Type</param>
+        /// <param name="ExceptionDetails">Exception Details</param>
+        /// <param name="DiagEnvironment">Diag Environment</param>
+        /// <param name="DiagWebsiteHostName">Diag hostname</param>
+        [Event(5602, Level = EventLevel.Error, Channel = EventChannel.Admin, Message = ETWMessageTemplates.LogMonitoringEventException)]
+        public void LogMonitoringEventException(string Source, string Message, string ExceptionType, string ExceptionDetails, string DiagEnvironment = null, string DiagWebsiteHostName = null)
```
puneetg1983, line 817: >string DiagEnvironment = null, string DiagWebsiteHostName = null [](start = 126, length = 64)

This is true for the rest of this class too but are we using these parameters anywhere?

src/Diagnostics.RuntimeHost/Services/MonitoringService/AuoHealMonitoringService.cs
```diff
@@ -0,0 +1,106 @@
...
+        /// <param name="environment">Environment object</param>
+        public AuoHealMonitoringService(IConfiguration configuration, IHostingEnvironment environment)
+        {
+            isEnabled = configuration.GetValue("AutoHealMonitoringSettings:IsEnabled", true);
+            iterationDelayInSeconds = configuration.GetValue("AutoHealMonitoringSettings:IterationDelayInSeconds", 30 * 60);
+            appHostConfigFilePath = Environment.GetEnvironmentVariable("APP_POOL_CONFIG");
+
+            if (environment.IsProduction() && isEnabled)
+            {
+                LogAutoHealSettings();
```
puneetg1983, line 39: >LogAutoHealSettings [](start = 16, length = 19)

I don't know this works but reading this code looks like mixing sync code with Async. Also, does this mean that constructor is blocked because we have a While loop in LogAutoHealSettings? 

Maybe a better approach is to use a timer and initialize the timer in constructor and set its interval equal to IterationDelayInSeconds.

The recommended way as suggested by ASP.NET core is to use IHostedService - [https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services)

puneetg1983 (approved):

src/Diagnostics.RuntimeHost/Services/MonitoringService/AuoHealMonitoringService.cs
ShekharGupta1988: this method  "LogAutoHealSettings" is fire and forget Task (not sync code). And since there is no "await" statement, the constructor is not blocked.  (Verified during testing).

I dont see a need to over complicate the logic here.

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
ShekharGupta1988: The way this works is these parameters are treated as Event Definitions (which at the end of the day gets translated to columns in Kusto)

src/Diagnostics.RuntimeHost/Utilities/CredentialTapper.cs
ShekharGupta1988: I found the code in Geomaster Data Provider. You created this method as Public:
public string RemovePIIFromSettings(string content) {... }

Any reason for that? I can remove it but I am not sure if the intention was to use it in detectors as well (in addition to the gist) since the method is Public.

puneetg1983: Feel free to remove. I think the detectors are using the gist

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
puneetg1983: I see.

src/Diagnostics.RuntimeHost/Services/MonitoringService/AuoHealMonitoringService.cs
puneetg1983: okay. That sounds good!

src/Diagnostics.RuntimeHost/Utilities/CredentialTapper.cs
ShekharGupta1988: Updated the PR

rekhaswaminathan (approved):

PR #656, opened 2021-03-22
Create CosmosDB.csx

yunjchoi:


data/templates/Detector_CosmosDB.csx
```diff
@@ -0,0 +1,26 @@
+private static string GetQuery(OperationContext<AzureCosmosDB> cxt)
+{
+    return
+    $@"
+        let startTime = datetime({cxt.StartTime});
+        let endTime = datetime({cxt.EndTime});
+        cluster('Aks').database('AKSprod').<YOUR_TABLE_NAME>
```
khaled-zayed, line 7: I think we should update those to some cosmosdb ones

data/templates/Detector_CosmosDB.csx
```diff
@@ -0,0 +1,26 @@
...
+        | where Timestamp >= startTime and Timestamp <= endTime
+        | <YOUR_QUERY>";
+}
+
+[AzureCosmosDBFilter]
```
khaled-zayed, line 12: I believe this should be an ArmResourceFilterlike this:
[ArmResourceFilter(provider: "Microsoft.DocumentDB", resourceTypeName: "databaseAccounts")]

khaled-zayed (changes requested):

data/templates/Detector_CosmosDB.csx
yunjchoi: Thanks for catching this!

data/templates/Detector_CosmosDB.csx
yunjchoi: I updated the file with your suggestion.

khaled-zayed (approved):

PR #657, opened 2021-03-23
Add section data rendering

XiaoxuHU:
This is to add a new DataRendering for sectionalize detector 

PR for UI change,
https://github.com/Azure/Azure-AppServices-Diagnostics-Portal/pull/1083

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Section.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Section.cs
@@ -81,7 +81,7 @@ namespace Diagnostics.ModelsAndUtils.Models.ResponseExtensions
                     section.IsExpanded.ToString()
                 }); ;
             }
-            
+
             var diagData = new DiagnosticData()
             {
                 Table = table,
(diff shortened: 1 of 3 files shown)


src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Section.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// </summary>
+        public bool IsExpanded;
+
+        /// <summary>
+        /// create an instance of Section class
+        /// </summary>
+        /// <param name="data"> </param>
+        /// <param name="title">Section Title</param>
+        /// <param name="isExpanded">Default expand state of section</param>
+        public Section(Response data, string title = "", bool isExpanded = true)
```
ShekharGupta1988, line 32: nitpick - the logical order of the Section is the title first and then inner response. Can you update the order of parameters in the ctor() as well.

XiaoxuHU: Updated

ShekharGupta1988 (approved):

PR #662, opened 2021-03-31
Add retry for Change Analysis and Geomaster

XiaoxuHU:
This PR is a following up one with [PR 645](https://github.com/Azure/Azure-AppServices-Diagnostics/pull/645), due to I unfolked repo and not able to push the last commit. 

This PR is for [WI 1657](https://dev.azure.com/app-service-diagnostics-portal/app-service-diagnostics-portal/_workitems/edit/1657) to add retry for other data providers

The latest change is I tried to tweak **RetryAsync** method in **RetryHelper** to pass func with either one param or no param to solve multi-thread issue last time Shekhar pointed out

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -632,15 +632,15 @@ namespace Diagnostics.DataProviders
                 ApiVersion = apiVersion,
                 CancellationToken = cancellationToken
             };
-            
-            var task = RetryHelper.RetryAsync<R>(PerformHttpRequest<R>,param, "GeoMasterDataProvider", RequestId, _configuration.MaxRetryCount, _configuration.RetryDelayInSeconds * 1000);
+
+            var task = RetryHelper.RetryAsync<R>(PerformHttpRequest<R>, param, "GeoMasterDataProvider", RequestId, _configuration.MaxRetryCount, _configuration.RetryDelayInSeconds * 1000);
             return task;
         }
 
         private Task<R> PerformHttpRequest<R>(object obj)
         {
             var param = (GeoMasterRetryParam)obj;
-            return PerformHttpRequest<R>(param.HttpMethod,param.Path,param.QueryString,param.Content,param.ApiVersion,param.CancellationToken);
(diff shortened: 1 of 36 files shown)


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -568,22 +570,44 @@ public Task<T> InvokeDaasExtension<T>(string subscriptionId, string resourceGrou
...
+            {
+                HttpMethod = method,
+                Path = path,
+                QueryString = queryString,
+                Content = content,
+                ApiVersion = apiVersion,
+                CancellationToken = cancellationToken
+            };
+            
+            var task = RetryHelper.RetryAsync<R>(PerformHttpRequest<R>,param, "GeoMasterDataProvider", RequestId, _configuration.MaxRetryCount, _configuration.RetryDelayInSeconds * 1000);
```
ShekharGupta1988, line 596: nitpick : spacing between arguments. Just format the whole document :)

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -712,5 +736,14 @@ public override async Task<HealthCheckResult> CheckHealthAsync(CancellationToken
 
             return result ?? await base.CheckHealthAsync(cancellationToken);
         }
+
+        class GeoMasterRetryParam {
```
ShekharGupta1988, line 740: Does these needs to public classes or can they simply be internal or private?

ShekharGupta1988 (approved):

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
XiaoxuHU: Yes, will do

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
XiaoxuHU: I think for nested class default will be private(https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/nested-types)

image

Anyway, I can add private modifier for these class

PR #663, opened 2021-03-31
Origin/darreldonald/log analytics data provider

DarrelDonald:
I have added a data provider for K8SE Log Analytics. 

This data provider is specific to K8SE.
&nbsp;&nbsp;&nbsp;&nbsp; A generic Log Analytics data provider would require users to pass the workspace ID
&nbsp;&nbsp;&nbsp;&nbsp; This workspace is in AME tenant and wouldn't be able to use the same app id/secret as workspaces in Microsoft tenant.

The method to call in the detector code to use this data provider is `dp.K8SELogAnalytics.ExecuteQueryAsync` 
&nbsp;&nbsp;&nbsp;&nbsp;It takes a log analytics query as a string for its input and returns the result as a data table.

Example:
&nbsp;&nbsp;&nbsp;&nbsp;```var tableTask = dp.K8SELogAnalytics.ExecuteQueryAsync("union * | take 5");```


src/Diagnostics.RuntimeHost/Diagnostics.RuntimeHost.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
rekhaswaminathan, line 4: Was this change intended?

rekhaswaminathan:
Please consider writing a simple HealthCheck method that validates the connectivity to LogAnalytics endpoint. This is helpful in catching any exceptions before the impact reaches production and is also part of our devops process.

Eg: 
https://github.com/Azure/Azure-AppServices-Diagnostics/blob/524926bcb3607cba6ab0548d0317cce39738b430/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs#L196

rekhaswaminathan:
Can we rename K8SELogAnalytics to something more meaningful? The term 'K8SE' might not be clear for someone new/not familiar with the system.

src/Diagnostics.DataProviders/Interfaces/ILogAnalyticsClient.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Diagnostics.DataProviders.Interfaces
+{
+    public interface ILogAnalyticsClient
+    {
+    }
```
rekhaswaminathan, line 9: This class is empty. Declare the method here- `ExecuteQueryAsync(string query)`

src/Diagnostics.DataProviders/K8SELogAnalyticsClient.cs
```diff
@@ -0,0 +1,112 @@
...
+            _requestId = RequestId;
+        }
+
+        public async Task<DataTable> ExecuteQueryAsync(string query)
+        {
+            var timeTakenStopWatch = new Stopwatch();
+            QueryResults queryResults;
+            Exception dataProviderException = null;
+            //check query
+
```
rekhaswaminathan, line 38: You can add some validation for `query` parameter here.

src/Diagnostics.DataProviders/TokenService/K8SELogAnalyticsTokenService.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        public void Initialize(K8SELogAnalyticsDataProviderConfiguration k8SELogAnalyticsDataProviderConfiguration)
+        {
+            workspaceId = k8SELogAnalyticsDataProviderConfiguration.WorkspaceId;
+            clientId = k8SELogAnalyticsDataProviderConfiguration.ClientId;
+            clientSecret = k8SELogAnalyticsDataProviderConfiguration.ClientSecret;
+
+            domain = k8SELogAnalyticsDataProviderConfiguration.Domain;
+            authEndpoint = k8SELogAnalyticsDataProviderConfiguration.AuthEndpoint;
+            tokenAudience = k8SELogAnalyticsDataProviderConfiguration.TokenAudience;
```
hforeste, line 42: What do partners need to provide so that they can query their log analytics?

src/Diagnostics.RuntimeHost/Diagnostics.RuntimeHost.csproj
DarrelDonald: Yes. The packages needed to use the log analytics api require 2.1. The build fails with 2.2

src/Diagnostics.DataProviders/TokenService/K8SELogAnalyticsTokenService.cs
DarrelDonald: It only queries a specific workspace. To add other workspaces we would need the workspace ID, the domain it's hosted on, and a client ID and secret for an AAD app with access to the workspace.

src/Diagnostics.RuntimeHost/Diagnostics.RuntimeHost.csproj
rekhaswaminathan: Is there any workaround? Not sure what the impact of 2.1 would be and moreover we are trying to progress to latest versions (3.x).

ShekharGupta1988: I also had to do this change on my local proj to get the runtimehost working. @DarrelDonald - lets not push this change for now and keep it only on your box.

src/Diagnostics.DataProviders/K8SELogAnalyticsClient.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        //convert results to DataTable
+        private DataTable ResultAsDataTable(QueryResults results)
+        {
+            DataTable dataTable = new DataTable("results");
+            dataTable.Clear();
+
+            int cursor = 0;
+            foreach (var i in results.Tables[0].Columns)
```
ShekharGupta1988, line 86: why do you need cursor variable here. You can simply use i variable in foreach to access the columns[cursor].
Regardless, explore System.Linq in c# to achieve these things in a much more smaller and compact code.
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.selectmany?view=net-5.0

ShekharGupta1988: Also, since the structures of Data Table and results.Table[0] seems very similar, see if you can simply typecast the object (if that works)

src/Diagnostics.DataProviders/K8SELogAnalyticsClient.cs
```diff
@@ -0,0 +1,112 @@
...
+namespace Diagnostics.DataProviders
+{
+    class K8SELogAnalyticsClient : ILogAnalyticsClient
+    {
+        private Microsoft.Rest.ServiceClientCredentials creds;
+        private string workspaceId;
+        private static OperationalInsightsDataClient client;
+        private string _requestId;
+
+        public K8SELogAnalyticsClient(string RequestId)
```
ShekharGupta1988, line 23: nitpick: small r in requestId parameter

src/Diagnostics.DataProviders/LogAnalyticsClientBase.cs
```diff
@@ -0,0 +1,80 @@
...
+            finally
+            {
+                timeTakenStopWatch.Stop();
+                var StopTime = DateTime.Now;
+                var StartTime = StopTime.AddMilliseconds(-timeTakenStopWatch.ElapsedMilliseconds);
+
+                if (dataProviderException != null)
+                {
+                    DiagnosticsETWProvider.Instance.LogDataProviderException(_requestId,
+                        "K8SELogAnalyticsDataProvider", StartTime.ToString("HH:mm:ss.fff"),
```
ShekharGupta1988, line 45: move the data provider name - "K8SELogAnalyticsDataProvider" as well to the derived class.

src/Diagnostics.DataProviders/LogAnalyticsClientBase.cs
```diff
@@ -0,0 +1,80 @@
...
+                }
+
+            }
+
+            if (queryResults == null)
+            {
+                queryResults = new QueryResults();
+            }
+            var dataTable = ResultAsDataTable(queryResults);
+            //DataTable dataTable = Enumerable.Cast<DataTable>((Enumerable) queryResults.Tables[0]);
```
ShekharGupta1988, line 63: remove this comment

ShekharGupta1988 (approved):
Left few small comments. Rest of the code looks good to me.
Great work in abstracting out the common logic to base classes :)

PR #671, opened 2021-04-13
Adding methods to MDM client from SDK

puneetg1983:


src/Diagnostics.DataProviders/Diagnostics.DataProviders.csproj
```diff
@@ -23,6 +23,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.2.0" />
     <PackageReference Include="Microsoft.Azure.Kusto.Data" Version="7.1.1" />
     <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.0.3" />
+    <PackageReference Include="Microsoft.Cloud.Metrics.Client" Version="2.2021.223.1905" />
```
ShekharGupta1988, line 26: Once this is merged to main, can you also merge this to replat3.1main and verify it this works with .net core 3.1 version as well. (today .netcore 3.1 is only deployed into few endpoints to test in production)
We are in this transitionary phase from .net core 2.1 to 3.1 and we expect that within next 2 weeks, 3.1 branch will become the default main branch.

ShekharGupta1988 (approved):

src/Diagnostics.DataProviders/Diagnostics.DataProviders.csproj
puneetg1983: Yup, will do

PR #682, opened 2021-04-30
Support observer client errors

hforeste:
This addresses an issue where in Fairfax we're failing the API with 500 because Observer responds with 404 when it cannot find a table. Example detector

```
dp.Observer.ExecuteSqlQuery('select * from dbo.view_sometable')
```

In the example dbo.view_sometable does not exist so Observer responds with 404 and our diagnostics is failing with 500.


ShekharGupta1988:
Hey @hforeste - Isnt this a genuine http 500 for runtime host API? sending 404 from Observer API makes sense if the table doesnt exist, but for runtimehost, it should still be a server error so that the detector can be fixed?

hforeste:
> Hey @hforeste - Isnt this a genuine http 500 for runtime host API? sending 404 from Observer API makes sense if the table doesnt exist, but for runtimehost, it should still be a server error so that the detector can be fixed?

So instead of a 500 for runtimehost I'd like to give kind of a warning feedback so that they can fix their detector. Something more graceful.

ShekharGupta1988:
My worry is we will miss genuine scenarios by consuming 4xx status code. (How would this work lets say if there is a 401, 403, 429 etc from observer ExecuteSQLQueryAsync. )
I like the idea of giving a graceful error message to customer but I feel it shouldn't  be wrapped as Http 200 for API call. It should be something like 500.Sub_Status where this Sub_Status means a dependency (like observer) failed for a specific reason. Today we dont have sub status code so either we can add that (which would be nice :) or  the second option is we throw a specific exception like ObserverTableNotFoundException and we can include the error message in the Http 500 response. (show it in the UI)

hforeste:
All those are valid status codes, thank you for mentioning those. I'll make an edit to only consider 404 and 400. I think that should address the major concerns here. This way we can still error on the rest of 4xx status code (401,403,408) + 5xx.

In general, when we have an abnormal amount of 4xx class errors coming from any Observer API, that should be captured in our monitoring and in our service health metrics. So when something genuine is happening we wont miss it. Thankfully we don't get any 404 for this specific API as I checked for the past 7 days. This is really just a corner case of lack of parity for a specific set of tables in n-clouds.

For those detectors that have a dependency on those tables, the detector author should decide how they'd like to handle the situation ie., throw an exception, or show partial data or disable detector for that cloud. I think the platform is correct in not throwing an exception at least for 404 and 400 and put the ball in the detector authors court on what they need to do. Normally what I have seen is detector authors just simply check if the data is empty or not. So they should be able to determine next steps from there.

In general, I do love the idea of a feedback channel that the platform can give back to the UI perhaps some message that we can show up as an insight from the platform. Could be status code like 200.3006 to show some warning message from data providers ie., kusto throttling

ShekharGupta1988 (approved):

yifguoMSFT (approved):

PR #684, opened 2021-05-03
Add localization for runtime host

cindylovescoding:
1. Parse language query parameter for ListDetectors and GetDetector calls
2. Add DianosticTranslatorService to handle detector dataset and detector metadata localization
3. Add Translation Cache service to store cached <language, original strings> => < translated strings > mapping before we make outbound translator API calls
4. There will be another frontend PR to handle UI framework localization

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3 @@
+{
+  "lockfileVersion": 1
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,3 +1,0 @@
-{
-  "lockfileVersion": 1
-}
--- a/src/Diagnostics.RuntimeHost/Controllers/ApiManagement/ApiManagementController.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/ApiManagement/ApiManagementController.cs
@@ -28,15 +28,15 @@ namespace Diagnostics.RuntimeHost.Controllers
         }
 
         [HttpPost(UriElements.Detectors)]
-        public async Task<IActionResult> ListDetectors(string subscriptionId, string resourceGroupName, string serviceName, [FromBody] dynamic postBody, [FromQuery(Name = "text")] string text = null)
+        public async Task<IActionResult> ListDetectors(string subscriptionId, string resourceGroupName, string serviceName, [FromBody] dynamic postBody, [FromQuery(Name = "text")] string text = null, [FromQuery] string l = "")
         {
-            return await base.ListDetectors(GetResource(subscriptionId, resourceGroupName, serviceName), text);
+            return await base.ListDetectors(GetResource(subscriptionId, resourceGroupName, serviceName), text, language: l.ToLower());
         }
 
         [HttpPost(UriElements.Detectors + UriElements.DetectorResource)]
-        public async Task<IActionResult> GetDetector(string subscriptionId, string resourceGroupName, string serviceName, string detectorId, [FromBody] dynamic postBody, string startTime = null, string endTime = null, string timeGrain = null, [FromQuery][ModelBinder(typeof(FormModelBinder))] Form form = null)
+        public async Task<IActionResult> GetDetector(string subscriptionId, string resourceGroupName, string serviceName, string detectorId, [FromBody] dynamic postBody, string startTime = null, string endTime = null, string timeGrain = null, [FromQuery][ModelBinder(typeof(FormModelBinder))] Form form = null, [FromQuery] string l = "")
         {
-            return await base.GetDetector(GetResource(subscriptionId, resourceGroupName, serviceName), detectorId, startTime, endTime, timeGrain, form: form);
+            return await base.GetDetector(GetResource(subscriptionId, resourceGroupName, serviceName), detectorId, startTime, endTime, timeGrain, form: form, language: l.ToLower());
         }
 
         [HttpPost(UriElements.Detectors + UriElements.DetectorResource + UriElements.StatisticsQuery)]
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -100,14 +100,14 @@ namespace Diagnostics.RuntimeHost.Controllers
 
             try
             {
-                if (IsTranslationApplicable(language))
+                if (IsLocalizationApplicable(language))
                 {
-                    listDetectorsTranslatedResponse = await _diagnosticTranslator.GetMetadataTranslations(listDetectorsResponse, language);
+                    listDetectorsTranslatedResponse = await _diagnosticTranslator.GetMetadataTranslations(listDetectorsResponse, language).ConfigureAwait(false);
                 }
             }
             catch (Exception ex)
             {
-                // Log translation exceptions and return orgininal untranslated response
+                // Log translation exceptions and return original untranslated response
                 DiagnosticsETWProvider.Instance.LogRuntimeHostHandledException(cxt.OperationContext.RequestId, "ListDetectorsTranslations", cxt.OperationContext.Resource.SubscriptionId,
                     cxt.OperationContext.Resource.ResourceGroup, cxt.OperationContext.Resource.Name, ex.GetType().ToString(), ex.ToString());
                 listDetectorsTranslatedResponse = listDetectorsResponse;
@@ -130,7 +130,7 @@ namespace Diagnostics.RuntimeHost.Controllers
             try
             {
                 
-                if (IsTranslationApplicable(language) && detectorResponse != null)
+                if (IsLocalizationApplicable(language) && detectorResponse != null)
                 {
                     responseObject = await this._diagnosticTranslator.GetResponseTranslations(detectorResponse.Item1, language);
                 }
@@ -138,7 +138,7 @@ namespace Diagnostics.RuntimeHost.Controllers
             }
             catch (Exception ex)
             {
-                // Log translation exceptions and return orgininal untranslated resp
+                // Log translation exceptions and return original untranslated resp
                 DiagnosticsETWProvider.Instance.LogRuntimeHostHandledException(cxt.OperationContext.RequestId, "GetDetectorTranslations", cxt.OperationContext.Resource.SubscriptionId,
     cxt.OperationContext.Resource.ResourceGroup, cxt.OperationContext.Resource.Name, ex.GetType().ToString(), ex.ToString());
                 responseObject = detectorResponse.Item1;
@@ -544,7 +544,7 @@ namespace Diagnostics.RuntimeHost.Controllers
 
         #endregion API Response Methods
 
-        protected bool IsTranslationApplicable(string language)
+        protected bool IsLocalizationApplicable(string language)
         {
             return !String.IsNullOrWhiteSpace(language) && string.Compare(language, "en", StringComparison.Ordinal) != 0 && !language.StartsWith("en.", StringComparison.CurrentCulture);
         }
@@ -734,7 +734,7 @@ namespace Diagnostics.RuntimeHost.Controllers
             await _sourceWatcherService.Watcher.WaitForFirstCompletion();
             var invoker = this._invokerCache.GetEntityInvoker<TResource>(gistId, context);
 
-
+            if (invoker == null)
             {
                 return null;
             }
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -51,7 +51,6 @@ namespace Diagnostics.RuntimeHost.Controllers
         protected ISupportTopicService _supportTopicService;
         protected IKustoMappingsCacheService _kustoMappingCacheService;
         protected IRuntimeLoggerProvider _loggerProvider;
-        protected ITranslationCacheService _translationCacheService;
 
         private InternalAPIHelper _internalApiHelper;
         private IDiagEntityTableCacheService tableCacheService;
@@ -70,7 +69,6 @@ namespace Diagnostics.RuntimeHost.Controllers
             this._searchService = (ISearchService)services.GetService(typeof(ISearchService));
             this._supportTopicService = (ISupportTopicService)services.GetService(typeof(ISupportTopicService));
             this._kustoMappingCacheService = (IKustoMappingsCacheService)services.GetService(typeof(IKustoMappingsCacheService));
-            this._translationCacheService = (ITranslationCacheService)services.GetService(typeof(ITranslationCacheService));
             this._loggerProvider = (IRuntimeLoggerProvider)services.GetService(typeof(IRuntimeLoggerProvider));
             tableCacheService = (IDiagEntityTableCacheService)services.GetService(typeof(IDiagEntityTableCacheService));
             this._diagnosticTranslator = (IDiagnosticTranslatorService)services.GetService(typeof(IDiagnosticTranslatorService));
@@ -100,10 +98,12 @@ namespace Diagnostics.RuntimeHost.Controllers
 
             try
             {
-                if (IsLocalizationApplicable(language))
+                if (string.IsNullOrEmpty(language) && this.Request.Headers.ContainsKey(HeaderConstants.LocalizationHeader))
                 {
-                    listDetectorsTranslatedResponse = await _diagnosticTranslator.GetMetadataTranslations(listDetectorsResponse, language).ConfigureAwait(false);
+                    language = this.Request.Headers[HeaderConstants.LocalizationHeader];
                 }
+
+                listDetectorsTranslatedResponse = await _diagnosticTranslator.GetMetadataTranslations(listDetectorsResponse, language);               
             }
             catch (Exception ex)
             {
@@ -129,12 +129,12 @@ namespace Diagnostics.RuntimeHost.Controllers
 
             try
             {
-                
-                if (IsLocalizationApplicable(language) && detectorResponse != null)
+                if (string.IsNullOrEmpty(language) && this.Request.Headers.ContainsKey(HeaderConstants.LocalizationHeader))
                 {
-                    responseObject = await this._diagnosticTranslator.GetResponseTranslations(detectorResponse.Item1, language);
+                    language = this.Request.Headers[HeaderConstants.LocalizationHeader];
                 }
 
+                responseObject = await this._diagnosticTranslator.GetResponseTranslations(detectorResponse.Item1, language);
             }
             catch (Exception ex)
             {
@@ -544,10 +544,6 @@ namespace Diagnostics.RuntimeHost.Controllers
 
         #endregion API Response Methods
 
-        protected bool IsLocalizationApplicable(string language)
-        {
-            return !String.IsNullOrWhiteSpace(language) && string.Compare(language, "en", StringComparison.Ordinal) != 0 && !language.StartsWith("en.", StringComparison.CurrentCulture);
-        }
 
         protected TResource GetResource(string subscriptionId, string resourceGroup, string name)
         {
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -96,14 +96,24 @@ namespace Diagnostics.RuntimeHost.Controllers
             }
 
             IEnumerable<DiagnosticApiResponse> listDetectorsResponse = await this.ListDetectorsInternal(cxt, queryText);
+            IEnumerable<DiagnosticApiResponse> listDetectorsTranslatedResponse = listDetectorsResponse;
 
-            language = language.ToLower();
-            if (IsTranslationApplicable(language))
+            try
             {
-                listDetectorsResponse = await _diagnosticTranslator.GetMetadataTranslations(listDetectorsResponse, language);
+                if (IsTranslationApplicable(language))
+                {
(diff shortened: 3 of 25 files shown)


src/Diagnostics.RuntimeHost/Controllers/ApiManagement/ApiManagementController.cs
```diff
@@ -28,15 +28,15 @@ public async Task<IActionResult> ExecuteQuery(string subscriptionId, string reso
         }
 
         [HttpPost(UriElements.Detectors)]
-        public async Task<IActionResult> ListDetectors(string subscriptionId, string resourceGroupName, string serviceName, [FromBody] dynamic postBody, [FromQuery(Name = "text")] string text = null)
+        public async Task<IActionResult> ListDetectors(string subscriptionId, string resourceGroupName, string serviceName, [FromBody] dynamic postBody, [FromQuery(Name = "text")] string text = null, [FromQuery] string l = "")
```
ShekharGupta1988, line 31: nitpick : instead of 'l', can we call it 'locale'

cindylovescoding: I thought about this and then I found in all the ARM calls, they are using something like "**l=ja.ja-jp**". For example:
 
https://appservice-diagnostics.azurefd.net/?clientOptimizations=undefined&**l=ja.ja-jp**&trustedAuthority=https%3A%2F%2Fms.portal.azure.com&shellVersion=undefined

I am open to modify this if that's more understandable:)

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -500,6 +544,11 @@ protected async Task<IActionResult> GetInsights(TResource resource, string pesId
 
         #endregion API Response Methods
 
+        protected bool IsLocalizationApplicable(string language)
```
ShekharGupta1988, line 547: I think you should move this method as well to TranslatorService (In case you need to repurpose this in some other components in future)

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -500,6 +544,11 @@ protected async Task<IActionResult> GetInsights(TResource resource, string pesId
...
+        {
+            return !String.IsNullOrWhiteSpace(language) && string.Compare(language, "en", StringComparison.Ordinal) != 0 && !language.StartsWith("en.", StringComparison.CurrentCulture);
```
ShekharGupta1988, line 549: use OrdinalIgnoreCase to be on the safer side

src/Diagnostics.RuntimeHost/Controllers/ApiManagement/ApiManagementController.cs
ShekharGupta1988: ah ok.. nvm then

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
```diff
@@ -0,0 +1,288 @@
...
+        Task<Response> GetResponseTranslations(Response response, string language);
+        Task<IEnumerable<DiagnosticApiResponse>> GetMetadataTranslations(IEnumerable<DiagnosticApiResponse> listDetectorsResponse, string language);
+    }
+
+    public class DiagnosticTranslatorService : IDiagnosticTranslatorService
+    {
+        private IConfiguration _config;
+
+        private static string _translatorSubscriptionKey;
+        private static readonly string translatorBaseURL = "https://api.cognitive.microsofttranslator.com/";
```
ShekharGupta1988, line 30: I think this should go to App Setting file. (Thinking about other clouds as well)

src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -149,6 +149,9 @@
   },
   "AutoHealMonitoringSettings": {
     "IsEnabled": true,
-    "IterationDelayInSeconds" : 1800
+    "IterationDelayInSeconds": 1800
+  },
+  "DiagnosticTranslator": {
```
ShekharGupta1988, line 154: I think it would be helpful to add a On/Off flag for Translation. Two Benefits:
- In case things dont go as planned, we can quickly turn this feature off
- For other S-Clouds, we dont want to regress the APIs and we can keep this off to begin with there.

cindylovescoding: Very good point!!

rekhaswaminathan:
The only thing I am concerned here if all query parameters get passed for Partner team's RP which calls Diagnostics Service backend (For ListDetectors call). For GeoMaster (Microsoft.Web), we made a change to send all query parameters that the client sends. For the language parameter, should we send it as a request header instead of query param? What are your thoughts @cindylovescoding @ShekharGupta1988

src/Diagnostics.RuntimeHost/appsettings.json
rekhaswaminathan: Please also add these settings to Dev KeyVault to maintain consistency :) Thanks

cindylovescoding: Will do @rekhaswaminathan:)

ShekharGupta1988:
> The only thing I am concerned here if all query parameters get passed for Partner team's RP which calls Diagnostics Service backend (For ListDetectors call). For GeoMaster (Microsoft.Web), we made a change to send all query parameters that the client sends. For the language parameter, should we send it as a request header instead of query param? Could be something like "x-ms-diag-language" ([Ref ](https://github.com/ShekharGupta1988/AppServiceDiagnosticsAPISample/blob/ea05707831afb1bd8c161906be1768847c580902/src/SampleAPIServer/Services/AppServiceDiagnosticsClientService.cs#L65)) What are your thoughts @cindylovescoding @ShekharGupta1988

Thats a good point. So Rekha - today partner RP code passes all x-ms headers to us? Maybe we can check it from kusto

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
```diff
@@ -0,0 +1,288 @@
...
+        private ITranslationCacheService _translationCacheService;
+        private static readonly HttpClient _httpClient = new HttpClient();
+
+        private void InitializeHttpClient()
+        {
+            _httpClient.Timeout = TimeSpan.FromSeconds(30);
```
ShekharGupta1988, line 36: Hey Cindy, I think we should have a lower timeout value here (like 10 seconds)
The reason is we dont want to impact perf if for whatever reason translation service is down or is slow 
I noticed you are already consuming any exceptions from this service and falling back to current behavior which is perfect.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -89,10 +94,29 @@ protected async Task<IActionResult> ListDetectors(TResource resource, string que
...
-            return Ok(await this.ListDetectorsInternal(cxt, queryText));
+
+            IEnumerable<DiagnosticApiResponse> listDetectorsResponse = await this.ListDetectorsInternal(cxt, queryText);
+            IEnumerable<DiagnosticApiResponse> listDetectorsTranslatedResponse = listDetectorsResponse;
+
+            try
+            {
+                if (IsLocalizationApplicable(language))
+                {
+                    listDetectorsTranslatedResponse = await _diagnosticTranslator.GetMetadataTranslations(listDetectorsResponse, language).ConfigureAwait(false);
```
hforeste, line 105: You can remove the `ConfigureAwait(false)`

hforeste:
Hey @cindylovescoding , how will the language be selected? Are customer expected to choose a language or are we autodetecting base in the region were in?

src/Diagnostics.RuntimeHost/Controllers/ApiManagement/ApiManagementController.cs
hforeste: LOL I was thinking the same thing

hforeste:
@cindylovescoding , I can see this being extremely useful in AzureChinaCloud once we get Diagnose and Solve enabled there. Looking forward to using this in the near future.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -67,8 +70,10 @@ public DiagnosticControllerBase(IServiceProvider services, IRuntimeContext<TReso
             this._searchService = (ISearchService)services.GetService(typeof(ISearchService));
             this._supportTopicService = (ISupportTopicService)services.GetService(typeof(ISupportTopicService));
             this._kustoMappingCacheService = (IKustoMappingsCacheService)services.GetService(typeof(IKustoMappingsCacheService));
+            this._translationCacheService = (ITranslationCacheService)services.GetService(typeof(ITranslationCacheService));
```
hforeste, line 73: It doesn't look like your using this in this class.

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
```diff
@@ -0,0 +1,288 @@
...
+
+            return diagnosticResponse;
+        }
+
+        public async Task<IEnumerable<DiagnosticApiResponse>> GetMetadataTranslations(IEnumerable<DiagnosticApiResponse> listDetectorsResponse, string language)
+        {
+            IEnumerable<DiagnosticApiResponse> translatedResponse = listDetectorsResponse;
+      
+            List<Definition> allDetectorsDefinitions = listDetectorsResponse.Select(detectorResponse => detectorResponse.Metadata).ToList<Definition>();
+            List<Definition> metadataTranslations = await GetMetadataTranslations(allDetectorsDefinitions, language).ConfigureAwait(false);
```
hforeste, line 118: You can remove `ConfigureAwait(false)`

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
```diff
@@ -0,0 +1,288 @@
...
+        {
+            if (metadataList == null || metadataList.Count == 0)
+            {
+                return metadataList;
+            }
+            List<Definition> responseMetaDataList = metadataList;
+            List<string> textsToTraslate = new List<string>();
+            metadataList.ForEach((metadata) =>
+            {
+                textsToTraslate.AddRange(new List<string> { metadata.Name != null ? metadata.Name: "", metadata.Description != null? metadata.Description: ""});
```
hforeste, line 66:
```suggestion
                textsToTraslate.AddRange(new List<string> { metadata.Name != null ? metadata.Name : "", metadata.Description != null ? metadata.Description : ""});
```

hforeste: Small edit, makes it easier to read.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
cindylovescoding: > You can remove the `ConfigureAwait(false)`
Oh Hawk I thought using ConfigureAwait(false) is a good practice to not capturing current context and avoid deadlock?

cindylovescoding:
> Hey @cindylovescoding , how will the language be selected? Are customer expected to choose a language or are we autodetecting base in the region were in?

@hforeste Customers will be able to select the language and location from Azure Portal. Our Iframe will pass the language and we localize the response based on that:

image

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
```diff
@@ -0,0 +1,288 @@
...
+            }
+
+            // Always look for cache before calling translator API
+            Tuple<string, string> key = new Tuple<string, string>(languageToTranslate, string.Join("_", textsToTranslate));
+            if (this._translationCacheService.TryGetValue(key, out List<string> value))
+            {
+                return value;
+            }
+
+            string route = string.Format("/translate?api-version=3.0&from=en&to={0}", languageToTranslate);
```
hforeste, line 250: Could we add this route and querystring to appsettings? Similiar to what we do for ASC

```
"AzureSupportCenter": {
    "BaseUri": "https://api.diagnostics.msftcloudes.com",
    "ApiUri": "/api/diagnosis/",
    "ApiVersion": "2018-02-01",
    "UserAgent": "AppLensClient",
    "AADAuthority": "https://login.microsoftonline.com/microsoft.onmicrosoft.com",
    "TokenResource": "https://microsoft.onmicrosoft.com/azurediagnostic",
    "ClientId": "",
    "AppKey": "",
    "Enabled": true,
    "DiagAscHeader":  ""
  }
```

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
```diff
@@ -0,0 +1,288 @@
...
+
+            try
+            {
+                using (var request = new HttpRequestMessage())
+                {
+                    request.Method = HttpMethod.Post;
+                    request.RequestUri = new Uri(translatorBaseURL + route);
+                    request.Content = new StringContent(requestBody, Encoding.UTF8, "application/json");
+
+                    HttpResponseMessage response = await _httpClient.SendAsync(request).ConfigureAwait(false);
```
hforeste, line 262: You can remove `ConfigureAwait(false)` and anywhere else you might be doing this.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -89,10 +94,29 @@ protected async Task<IActionResult> ListDetectors(TResource resource, string que
...
+            {
+                if (IsLocalizationApplicable(language))
+                {
+                    listDetectorsTranslatedResponse = await _diagnosticTranslator.GetMetadataTranslations(listDetectorsResponse, language).ConfigureAwait(false);
+                }
+            }
+            catch (Exception ex)
+            {
+                // Log translation exceptions and return original untranslated response
+                DiagnosticsETWProvider.Instance.LogRuntimeHostHandledException(cxt.OperationContext.RequestId, "ListDetectorsTranslations", cxt.OperationContext.Resource.SubscriptionId,
```
hforeste, line 111: Is there something we're going to mention in the UI if the translation is not successful? Like still render results but have a small note somewhere saying `Sorry failed to translate response, please try again later.`

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
hforeste: This is a nice resource https://devblogs.microsoft.com/dotnet/configureawait-faq/#when-should-i-use-configureawaitfalse

But generally, since your not writing library code and these are all application code changes that your implementing and we know that you are already following async/await best practices when calling the Cognitive Service API then you can avoid calling `ConfigureAwait(false)`. 

Deadlock would be something to think about if we were calling some library code in which we had no idea how it was implemented (worst case it was not implemented with thread-safety in mind and broke best practices)

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
cindylovescoding: Good catch! Updated

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
```diff
@@ -0,0 +1,288 @@
...
+
+namespace Diagnostics.RuntimeHost.Services.DiagnosticsTranslator
+{
+    public interface IDiagnosticTranslatorService
+    {
+        Task<Response> GetResponseTranslations(Response response, string language);
+        Task<IEnumerable<DiagnosticApiResponse>> GetMetadataTranslations(IEnumerable<DiagnosticApiResponse> listDetectorsResponse, string language);
+    }
+
+    public class DiagnosticTranslatorService : IDiagnosticTranslatorService
```
SharmaAjay19, line 25: Couple things:
1. We should have a "enabled" switch so that this can be managed depending on situation and environment.
2. Since we are caching string translations received from the translation API - which is basically caching content which double the size of the detector response, we should find a way to limit memory usage of this cache i.e. implement it as a LRU cache maybe.

package-lock.json
```diff
@@ -0,0 +1,3 @@
+{
+  "lockfileVersion": 1
```
SharmaAjay19, line 2: If we indeed want to completely move to package-lock.json for a reproducible build, we should go with lockfileVersion 2 as it is also backwards compatible with older node cli versions that could work with only lockfileVersion 1.

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
SharmaAjay19: Could probably do a memory test of 20-30 detector calls on localhost and see the impact.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
ShekharGupta1988: I think in case of exception, Cindy is falling back to old behavior (non-localized response) which is the right behavior.

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
cindylovescoding: Updated:)

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
cindylovescoding: Thanks Hawk! I removed it accordingly

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
cindylovescoding: Added query string and Api version into appsettings

package-lock.json
cindylovescoding: Removed this file:)

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
```diff
@@ -0,0 +1,320 @@
...
+        }
+
+        #region Localization Helper Methods
+
+        // Scenarios to enable localization:
+        // 1. EnableLocalization is not set or set to be true in appsettings/keyvault
+        // 2. Language to translate to is not empty or English
+        private static bool IsLocalizationApplicable(string language)
+        {
+            return (string.IsNullOrWhiteSpace(_enableLocalization) || string.Compare(_enableLocalization, "true", StringComparison.OrdinalIgnoreCase) == 0) && !String.IsNullOrWhiteSpace(language) && string.Compare(language, "en", StringComparison.OrdinalIgnoreCase) != 0 && !language.StartsWith("en.", StringComparison.CurrentCulture);
```
ShekharGupta1988, line 125: shouldnt this check be _enableLocalization != null && string.Compare(_enableLocalization, "true", StringComparison.OrdinalIgnoreCase) == 0 ?

ShekharGupta1988: better to test out with Off Flag as well to make sure it works for S-Clouds :)

ShekharGupta1988 (approved):

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
cindylovescoding: I thought we are gonna enable this by default. But Shekhar you are right, the default behavior should be OFF instead. I will test this:)

src/Diagnostics.RuntimeHost/Services/DiagnosticTranslator/DiagnosticTranslatorService.cs
```diff
@@ -0,0 +1,320 @@
...
+
+        /// <summary>
+        /// Creates a new instance of DiagnosticTranslatorService
+        /// </summary>
+        /// <param name="configuration">Configuration object</param>
+        /// <param name="translationCacheService">Translation cache service</param>
+        public DiagnosticTranslatorService(IConfiguration configuration, ITranslationCacheService translationCacheService)
+        {
+            _config = configuration;
+            _enableLocalization = _config[$"DiagnosticTranslator:EnableLocalization"];
```
hforeste, line 54: Can you rename `DiagnosticTranslator:EnableLocalization` to `DiagnosticTranslator:Enabled`. Just to keep consistency with what we called the feature flag for other services.

cindylovescoding: Updated

hforeste (approved):

PR #689, opened 2021-05-08
Show storage account name with masked key from AzureWebJobsStorage setting

cindylovescoding:
This PR is to retrieve storage account name from AzureWebJobsStorage settings and return the masked value. This is what AzureWebJobsStorage value looks like:

DefaultEndpointsProtocol=https;AccountName=[name];AccountKey=[key]

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
@@ -101,6 +101,20 @@ namespace Diagnostics.DataProviders
                     string value = CredentialTrapper.Obfuscate(item.Value);
                     appSettings.Add(item.Key, value);
                 }
+                else if(string.Compare(item.Key, "AzureWebJobsStorage", StringComparison.CurrentCulture) == 0)
+                {
+                    var storageAccountMatch = Regex.Match(item.Value.ToString(), @".*AccountName=(?<StorageAccount>.+?);.*");
+                    if (storageAccountMatch.Success)
+                    {
+                        string value = storageAccountMatch.Groups["StorageAccount"].Value;
+                        value = string.Concat("****AccountName=", value, "*****");
+                        appSettings.Add(item.Key, value);
+                    }
+                    else
+                    {
+                        appSettings.Add(item.Key, "******");
+                    }
+                }
                 else
                 {
                     appSettings.Add(item.Key, "******");


src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
```diff
@@ -101,6 +101,20 @@ private IGeoMasterClient InitClient()
                     string value = CredentialTrapper.Obfuscate(item.Value);
                     appSettings.Add(item.Key, value);
                 }
+                else if(string.Compare(item.Key, "AzureWebJobsStorage", StringComparison.CurrentCulture) == 0)
+                {
+                    var storageAccountMatch = Regex.Match(item.Value.ToString(), @".*AccountName=(?<StorageAccount>.+?);.*");
```
puneetg1983, line 106: Instead of RegEx, do you want to use [CloudStorageAccount.Parse](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.storage.cloudstorageaccount.parse?view=azure-dotnet) method instead?

puneetg1983: Also, are you trying to hide the storage account name only? AFAIK, account names themselves are not a concern. Are the keys for AzureWebJobsStorage already masked?

cindylovescoding: No Puneet, I want to only show the storage account name and hide the key part in AzureWebJobsStorage. So for AzureWebJobsStorage in appsettings, we will return ***AccountName={StorageAccountName}**** instead of "******", which is what we are currently returning for AzureWebJobsStorage app settings. Currently there is no easy way to figure out which storage account the customer is using. Showing storage account name part in the value will be helpful in a lot of cases.

puneetg1983: I see. Just add `RegexOptions.IgnoreCase` to Regex.Match and the rest is fine

puneetg1983 (changes requested):

puneetg1983 (approved):

src/Diagnostics.DataProviders/DataProviders/GeoMasterDataProvider.cs
cindylovescoding: @puneetg1983 do you know if the connection string value is case sensitive or not? I know the app settings key is case sensitive but not quite sure about the value. If it's case insensitive, add IgnoreCase option definitely makes sense:)

PR #693, opened 2021-05-12
Nmallick1/support topic lookup parity

nmallick1:
Adding a support topic ID map in JSON config for ASC to AppLens integration to work in national clouds. The functionality in public clouds will remain unaffected. This will help us be compliant across NClouds.


src/Diagnostics.RuntimeHost/Services/SupportTopicService/SupportTopicService.cs
```diff
@@ -52,17 +57,58 @@ public SupportTopicService()
...
+                    throw new Exception("Support topic map is empty.");
+                }
+
+                DataTable supportTopicTable = new DataTable();
+                supportTopicTable.Columns.Add("SupportTopicId", typeof(string));
+                supportTopicTable.Columns.Add("ProductId", typeof(string));
+                supportTopicTable.Columns.Add("SupportTopicPath", typeof(string));
+                
+                var supportTopicList = _configuration.GetSection("SupportTopicMap").GetChildren();
+                if(supportTopicList?.Any() == true)
```
hforeste, line 75:
```suggestion
                if(supportTopicList.Any() == true)
```

hforeste: Since you already checking if its empty on line 64

hforeste (approved):
Just left one comment but everything looks great.

ShekharGupta1988 (approved):

PR #694, opened 2021-05-12
Darreldonald log analytics data sources

DarrelDonald:
Log Analytics queries will show in the Data Sources tab

### DIFF
--- a/src/Diagnostics.DataProviders/Interfaces/ILogAnalyticsClient.cs
+++ b/src/Diagnostics.DataProviders/Interfaces/ILogAnalyticsClient.cs
@@ -11,6 +11,6 @@ namespace Diagnostics.DataProviders.Interfaces
     {
         Task<DataTable> ExecuteQueryAsync(string query);
 
-        Task<LogAnalyticsQuery> GetLogAnalyticsQueryAsync(string query);
+        LogAnalyticsQuery GetLogAnalyticsQuery(string query);
     }
 }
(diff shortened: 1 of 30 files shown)


src/Diagnostics.DataProviders/Interfaces/ILogAnalyticsClient.cs
```diff
@@ -3,11 +3,14 @@
...
 using System.Threading.Tasks;
+using static Diagnostics.DataProviders.LogAnalyticsDataProvider;
 
 namespace Diagnostics.DataProviders.Interfaces
 {
     interface ILogAnalyticsClient
     {
         Task<DataTable> ExecuteQueryAsync(string query);
+
+        Task<LogAnalyticsQuery> GetLogAnalyticsQueryAsync(string query);
```
hforeste, line 14: Does this need to be wrapped in a task?

src/Diagnostics.DataProviders/LogAnalyticsClientBase.cs
```diff
@@ -65,6 +66,16 @@ public async Task<DataTable> ExecuteQueryAsync(string query)
             return dataTable;
         }
 
+        public async Task<LogAnalyticsQuery> GetLogAnalyticsQueryAsync(string query)
```
hforeste, line 69: Since your not doing any IO calls calls does the response need to be a task?

hforeste (approved):

PR #705, opened 2021-06-04
Supply default text to ASC to avoid NRE

nmallick1:
Supply a default text when NULL  value is passed while creating ASC insight. This avoids an NRE at ASC.

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Utilities/AzureSupportCenterInsightUtilities.cs
+++ b/src/Diagnostics.ModelsAndUtils/Utilities/AzureSupportCenterInsightUtilities.cs
@@ -10,18 +10,21 @@ namespace Diagnostics.ModelsAndUtils.Utilities
 {
     public static class AzureSupportCenterInsightUtilites
     {
-        private static readonly string DefaultDescription = "This insight was raised as part of the {0} detector.";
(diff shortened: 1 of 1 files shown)


src/Diagnostics.ModelsAndUtils/Utilities/AzureSupportCenterInsightUtilities.cs
```diff
@@ -52,14 +52,13 @@ internal static AzureSupportCenterInsight CreateInsight<TResource>(string title,
                 InsightFriendlyName = category,
                 IssueCategory = category,
                 IssueSubCategory = category,
-                Description = description,
+                Description = description?? new Text("Please submit a feedback for the detector author via AppLens to supply a description for this insight.", false),
                 RecommendedAction = new RecommendedAction()
                 {
                     Id = Guid.NewGuid(),
-                    Text = recommendedAction
+                    Text = recommendedAction?? new Text("Please submit a feedback for the detector author via AppLens to populate recommended action.", false)
```
hforeste, line 59: Create a const string maybe defined at the top to hold this string value.

hforeste (approved):

src/Diagnostics.ModelsAndUtils/Utilities/AzureSupportCenterInsightUtilities.cs
nmallick1: Good point, updated the PR.

hforeste (approved):

PR #710, opened 2021-06-09
Enable for different metric types

XiaoxuHU:
When is not hovering the graph, it will display `--` in metric now, ideally it will show number(by calculating Average/Sum/Max) for each line in graph 
This PR is to allow detector author to specify different ways they want to calculate for metric 
image

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/Rendering.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/Rendering.cs
@@ -294,7 +294,12 @@ namespace Diagnostics.ModelsAndUtils.Models
         /// <summary>
         /// Data Rendered as section
         /// </summary>
-        Section
+        Section,
+
+        /// <summary>
+        /// StepViews component of Network Troubleshooter
+        /// </summary>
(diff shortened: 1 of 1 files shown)


src/Diagnostics.ModelsAndUtils/Models/Rendering.cs
```diff
@@ -355,4 +355,27 @@ public class TableColumnOption
         /// </summary>
         public bool Visible { get; set; } = true;
     }
+
+    /// <summary>
+    /// Default metric value show below graph
```
cindylovescoding, line 360: Do we want to add a min value as well?

cindylovescoding (approved):

src/Diagnostics.ModelsAndUtils/Models/Rendering.cs
XiaoxuHU: Good idea ! Let me add it

cindylovescoding (approved):

ShekharGupta1988 (approved):

PR #712, opened 2021-06-18
Add Insight.IsBackgroundPainted and TableColumnOption.DefaultSelection

michimune:
This PR adds the following for improving Linux Log Viewer:
- Insight.IsBackgroundPainted
- TableColumnOption.DefaultSelection

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Models/Rendering.cs
+++ b/src/Diagnostics.ModelsAndUtils/Models/Rendering.cs
@@ -55,6 +55,9 @@ namespace Diagnostics.ModelsAndUtils.Models
         public dynamic TableOptions { get; set; }
 
         public bool AllowColumnSearch { get; set; }
+
+        public string SearchPlaceholder { get; set; }
+
         public IEnumerable<TableColumnOption> ColumnOptions { get; set; }
 
         public TableRendering() : base(RenderingType.Table)
@@ -64,6 +67,7 @@ namespace Diagnostics.ModelsAndUtils.Models
             DescriptionColumnName = null;
             AllowColumnSearch = false;
             ColumnOptions = null;
+            SearchPlaceholder = null;
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/Diagnostics.ModelsAndUtils/Models/Rendering.cs
```diff
@@ -346,10 +347,15 @@ public class TableColumnOption
...
         /// <summary>
-        /// Filter Selelction(Single, Multiple)
+        /// Filter Selection(Single, Multiple)
         /// </summary>
         public FileterSelectionOption SelectionOption { get; set; } = FileterSelectionOption.None;
 
+        /// <summary>
+        /// Default filter selection. Use '|' characters for multiple options (e.g. "option1 | option2")
+        /// </summary>
+        public string DefaultSelection { get; set; }
```
XiaoxuHU, line 357: Michi, do you think change `DefaultSelection` to an array of string will be better instead of using `|` to separate selection ?

XiaoxuHU:
Can you also add `SearchPlaceholder` property in `TableRendering` to customize table search place holder ?

src/Diagnostics.ModelsAndUtils/Models/Rendering.cs
michimune: Agreed, it is better. Which do you prefer, string[] or List<string>?

XiaoxuHU: I think list is slightly better, so it can add default filter words more dynamic

cindylovescoding (approved):

cindylovescoding:
Thanks Michi! Could you help resolve the build error before merging it?

src/Diagnostics.ModelsAndUtils/Models/Rendering.cs
michimune: Change pushed. I used IEnumerable<string> like other classes.

michimune:
Thank you, Cindy. I fixed the Rendering.AddInsight() extension which caused the test failure.

XiaoxuHU (approved):

PR #718, opened 2021-07-01
Darreldonald retry specialcases

DarrelDonald:
Exceptions with "Could not connect to net.tcp" now retry against the leader if they fail within 30 seconds instead of following the normal retry sequence

More exceptions can be added in the future by adding to the Retry:OverridableExceptionsToRetryAgainstLeaderCluster list in appsettings.json
&nbsp;&nbsp;&nbsp;&nbsp; The number of exceptions in the list must be updated
&nbsp;&nbsp;&nbsp;&nbsp; Requires exception string and max failure response time


src/Diagnostics.DataProviders/ConfigurationFactory.cs
```diff
@@ -174,18 +175,76 @@ private void LoadConfigurationValues(object dataProviderConfiguration)
...
                     }
 
-                    var value = GetValue(prefix, attribute.Name);
+                    dynamic value;
+                    string valueType;
+                  
+                    if (property.Name == "OverridableExceptions")
+                    {
+                        List<ITuple> overridableExceptionsList = new List<ITuple>();
+                        for (int i = 0; i < int.Parse(GetValue(prefix, "Retry:OverridableExceptionsToRetryAgainstLeaderCluster:AmountOfOverridableExceptions")); i++)
```
nmallick1, line 184: Consider using int.TryParse instead of int.Parse

nmallick1: Is it possible to make this work without the property AmountOfOverridableExceptions?

src/Diagnostics.DataProviders/ConfigurationFactory.cs
```diff
@@ -174,18 +175,76 @@ private void LoadConfigurationValues(object dataProviderConfiguration)
...
+                        {
+                            overridableExceptionsList.Add((
+                                GetValue(prefix, $"Retry:OverridableExceptionsToRetryAgainstLeaderCluster:Cases:{i.ToString()}:ExceptionString"),
+                                double.Parse(GetValue(prefix, $"Retry:OverridableExceptionsToRetryAgainstLeaderCluster:Cases:{i.ToString()}:MaxFailureResponseTimeInSeconds"))));
```
nmallick1, line 188: Consider using TryParse.

src/Diagnostics.DataProviders/ConfigurationFactory.cs
```diff
@@ -174,18 +175,76 @@ private void LoadConfigurationValues(object dataProviderConfiguration)
...
                 }
             }
         }
 
         protected abstract string GetValue(string prefix, string name);
 
+        protected void SetValue(object target, PropertyInfo property, object objectValue, object defaultValue)
+        {
+            object value = null;
+            if (property.PropertyType == typeof(string))
```
nmallick1, line 227: Is it possible for the property.PropertyType to be something other than string/List<ITuple>?

src/Diagnostics.DataProviders/ConfigurationFactory.cs
```diff
@@ -229,6 +288,18 @@ protected void SetValue(object target, PropertyInfo property, string stringValue
                     value = doubleValue;
                 }
             }
+            /*else if (property.PropertyType == typeof(List<(string, double)>))
```
nmallick1, line 291: nit: delete commented code please.

src/Diagnostics.DataProviders/DataProviderConfigurations/KustoDataProviderConfiguration.cs
```diff
@@ -222,6 +230,8 @@ public override void PostInitialize()
                     FailoverClusterNameCollection.TryAdd(clusterNameGroupingParts[iterator], clusterFailoverGroupingParts[iterator]);
                 }
             }
+
+            //foreach ()
```
nmallick1, line 234: nit: Remove these two lines please.

src/Diagnostics.DataProviders/KustoSDKClient.cs
```diff
@@ -29,6 +30,7 @@ public class KustoSDKClient : IKustoClient
         private string _aadAuthority;
         private static ConcurrentDictionary<Tuple<string, string>, ICslQueryProvider> QueryProviderMapping;
         private static List<string> exceptionsToRetryFor = new List<string>();
+
```
nmallick1, line 33: nit: Extra line here, remove please.

src/Diagnostics.DataProviders/KustoSDKClient.cs
```diff
@@ -184,9 +188,12 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
                             exceptionType,
                             exceptionDetails
                             );
-
+                    if (attemptException is not null && totalResponseTime.TotalSeconds <= (double) _config.OverridableExceptions.Single(x => x[0].ToString().ToLower() == attemptException.Message.ToLower())[1] && IsSpecialRetryException(attemptException))
```
nmallick1, line 191: Do you want to use contains here instead of comparing the exact exception message?

src/Diagnostics.DataProviders/ConfigurationFactory.cs
DarrelDonald: Should I put the for loop inside of an if block that executes if the try parse is successful?

I wasn't able to find a way to count how many objects were in the list or just grab the list and iterate through it. I did see a list of all the key value pairs from the config file in watcher but I couldn't find a way to access it in the code. Do you know of any ways to do this?

src/Diagnostics.DataProviders/ConfigurationFactory.cs
DarrelDonald: Before I started working on this change it was always string and anything else threw the error in the else block. I just added an option for List<ITuple> here because that was what I was using to get this piece to work.

src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -64,7 +64,16 @@
       "RetryDelayInSeconds": 3,
       "UseBackupClusterForLastAttempt": true,
       "MaxFailureResponseTimeInSecondsForRetry": 22,
-      "ExceptionsToRetryFor": "Kusto.Data.Exceptions.WeakConsistencyEntityNotFoundException | InternalServiceError (520-UnknownError) | Source: Kusto::CachedStorageObject | Kusto.DataNode.Exceptions.QueryThrottledException"
+      "ExceptionsToRetryFor": "Kusto.Data.Exceptions.WeakConsistencyEntityNotFoundException | InternalServiceError (520-UnknownError) | Source: Kusto::CachedStorageObject | Kusto.DataNode.Exceptions.QueryThrottledException",
+      "OverridableExceptionsToRetryAgainstLeaderCluster": {
+        "AmountOfOverridableExceptions": 1,
```
hforeste, line 69: To @nmallick1 point earlier, do we need this?

src/Diagnostics.DataProviders/ConfigurationFactory.cs
```diff
@@ -174,18 +175,76 @@ private void LoadConfigurationValues(object dataProviderConfiguration)
                         continue;
                     }
 
-                    var value = GetValue(prefix, attribute.Name);
+                    dynamic value;
+                    string valueType;
+                  
+                    if (property.Name == "OverridableExceptions")
```
hforeste, line 181: `OverridableExceptions` is really specific to Kusto, instead of constructing the overridableExceptionsList here, perhaps we can construct that in the `KustoDataProviderConfiguration` class?

src/Diagnostics.DataProviders/ConfigurationFactory.cs
nmallick1: Following code snippet can help read an array from config.

https://github.com/Azure/Azure-AppServices-Diagnostics-KustoIngestor/blob/b43309976067f879c4e7d0f56e8b5af089b9c7cc/Kusto.Data.Ingestor/ExecuteCustomLogAgregatorsFunction.cs#L74

src/Diagnostics.DataProviders/DataProviderConfigurations/KustoDataProviderConfiguration.cs
```diff
@@ -222,6 +230,19 @@ public override void PostInitialize()
                     FailoverClusterNameCollection.TryAdd(clusterNameGroupingParts[iterator], clusterFailoverGroupingParts[iterator]);
                 }
             }
+
+            int numOfOverridableExceptions = config != null? config.GetSection("Kusto").GetSection("Retry").GetSection("OverridableExceptionsToRetryAgainstLeaderCluster").GetChildren().ToList().Count() : 0;
```
nmallick1, line 234: Instead of counting the elements in the array and then looping through them, see if you can make use of foreach

nmallick1 (changes requested):

ShekharGupta1988 (approved):

nmallick1 (approved):

PR #721, opened 2021-07-23
Get InternalOnly prop in Detector metadata

XiaoxuHU:
This PR is for adding InternalOnly property in detector metadata for listing detectors table in new Applens.

I added a new API **/resourceUri/internal/detectors** to get detector metadata from storage table. So in UI project I can merge it with detector metadata from GetDetectors() call

image

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -13,6 +13,7 @@ using Diagnostics.Logger;
 using Diagnostics.ModelsAndUtils.Attributes;
 using Diagnostics.ModelsAndUtils.Models;
 using Diagnostics.ModelsAndUtils.Models.ResponseExtensions;
+using Diagnostics.ModelsAndUtils.Models.Storage;
 using Diagnostics.ModelsAndUtils.ScriptUtilities;
 using Diagnostics.ModelsAndUtils.Utilities;
 using Diagnostics.RuntimeHost.Models;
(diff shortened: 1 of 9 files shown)


hforeste (approved):

nmallick1:
Can the error message for InternalOnly be modified to include a suggestion to the user? Something like..
"InternalOnly is not a valid named attribute field for the Definition attribute. Did you mean to specify InternalOnly on the App filter instead?"

nmallick1 (approved):

XiaoxuHU:
> Can the error message for InternalOnly be modified to include a suggestion to the user? Something like..
> "InternalOnly is not a valid named attribute field for the Definition attribute. Did you mean to specify InternalOnly on the App filter instead?"

Yes, let me figure out how I can customize the exception message

ShekharGupta1988 (approved):

ShekharGupta1988:
@nmallick1 - Can this impact the swagger doc ?

nmallick1:
> @nmallick1 - Can this impact the swagger doc ?

Good point.... yeah, the swagger would have to be updated for this..

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -116,6 +116,40 @@ protected async Task<IActionResult> ListDetectors(TResource resource, string que
...
+        /// Get Detector Meta Data with InternalOnly flag from Stoarge
+        /// </summary>
+        /// <param name="resource"></param>
+        /// <returns></returns>
+        protected async Task<IActionResult> ListDetectorsWithExtendMetaData(TResource resource)
+        {
+            DateTimeHelper.PrepareStartEndTimeWithTimeGrain(string.Empty, string.Empty, string.Empty, out DateTime startTimeUtc, out DateTime endTimeUtc, out TimeSpan timeGrainTimeSpan, out string errorMessage);
+            var cxt = PrepareContext(resource, startTimeUtc, endTimeUtc);
+            IEnumerable<ExtendedDefinition> detectors = new List<ExtendedDefinition>();
+            if (this.tableCacheService.IsStorageAsSourceEnabled())
```
nmallick1, line 129: Will this IsStorageAsSourceEnabled() ever be false (e.g.. MoonCake)? If so, how should that be handled?

XiaoxuHU: If there is no storage account table, it will return an empty list.
Then we can in UI show as "unknown" or something else

XiaoxuHU: Or I can add the same logic as how we did in **ListDetectorsInternal()** to get internalOnly flag from SourceWatcher if storage account is disabled

nmallick1 (approved):

rekhaswaminathan (approved):

PR #723, opened 2021-07-26
Mask PII data from AppInsights data provider

puneetg1983:
Mask any PII data in the data returned by App Insights data provider

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/ScriptUtilities/Anonymizer.cs
+++ b/src/Diagnostics.ModelsAndUtils/ScriptUtilities/Anonymizer.cs
@@ -11,7 +11,7 @@ namespace Diagnostics.ModelsAndUtils.ScriptUtilities
 
         public static string AnonymizeMessage(string message)
         {
-            if (string.IsNullOrEmpty(message))
+            if (string.IsNullOrWhiteSpace(message))
             {
                 return message;
             }
--- a/src/Diagnostics.ModelsAndUtils/ScriptUtilities/Anonymizer.cs
+++ b/src/Diagnostics.ModelsAndUtils/ScriptUtilities/Anonymizer.cs
@@ -1,0 +1,55 @@
+﻿using System;
+using System.Text.RegularExpressions;
+
+namespace Diagnostics.ModelsAndUtils.ScriptUtilities
+{
+    public static class Anonymizer
+    {
+        private const string DigitRegularExpression = @"\d";
+        private const string EmailRegularExpression = @"(?<=[\w]{0})[\w-.+%]*(?=@([\w-]+)[.]{0})";
+        private const string GuidRegularExpression = @"(\{){0,1}[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}(\}){0,1}";
+
+        public static string AnonymizeMessage(string message)
+        {
+            if (string.IsNullOrEmpty(message))
+            {
+                return message;
+            }
+
+            var array = message.Split(' ');
+            for (int i = 0; i < array.Length; i++)
+            {
+                var item = array[i];
+                item = ReplaceIfGuid(item);
+                item = ReplaceIfEmail(item);
+                item = ReplaceIfDigit(item);
+                array[i] = item;
+            }
+
+            return string.Join(" ", array);
+        }
+
+        private static string ReplaceIfGuid(string input)
+        {
+            return Regex.Replace(input, GuidRegularExpression, "{{guid}}");
+        }
+
+        private static string ReplaceIfDigit(string input)
+        {
+            var output = Regex.Replace(input, DigitRegularExpression, "n");
+            return output;
+        }
+
+        private static string ReplaceIfEmail(string input)
+        {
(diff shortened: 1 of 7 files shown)


rekhaswaminathan (approved):

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
```diff
@@ -140,7 +140,7 @@ private void BeginRequestHandle(HttpContext httpContext)
                 geomasterName = httpContext.Request.Headers[GeomasterNameHeader];
             }
 
-            httpContext.Items.Add(HostConstants.DataProviderContextKey, new DataProviderContext(dataSourcesConfigurationService.Config, values.FirstOrDefault().Any() ? values.FirstOrDefault().Split(new char[] { ',' })[0] : string.Empty, cTokenSource.Token, startTimeUtc, endTimeUtc, wawsObserverTokenService, supportBayApiObserverTokenService, clientObjId, clientPrincipalName, kustoHeartBeatService, geomasterHostName, geomasterName, null, httpContext.Request.Headers));
+            httpContext.Items.Add(HostConstants.DataProviderContextKey, new DataProviderContext(dataSourcesConfigurationService.Config, values.FirstOrDefault().Any() ? values.FirstOrDefault().Split(new char[] { ',' })[0] : string.Empty, cTokenSource.Token, startTimeUtc, endTimeUtc, wawsObserverTokenService, supportBayApiObserverTokenService, clientObjId, clientPrincipalName, kustoHeartBeatService, geomasterHostName, geomasterName, null, httpContext.Request.Headers, isInternalClient: isInternalClient));
         }
```
rekhaswaminathan, line 144: side note - This list is getting huge :D we should look to pass these parameters as class object (maybe in future)

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Anonymizer.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
```
cindylovescoding, line 1: Puneet do you think if customers accidentally put secrets here it will be handled with these 3 methods?

cindylovescoding (approved):

src/Diagnostics.RuntimeHost/Middleware/DiagnosticsRequestMiddleware.cs
puneetg1983: Yea, I felt the same. Leaving it as is for now but we can do it if this keep growing!

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Anonymizer.cs
puneetg1983: Secrets explicitly won't be handled. This is just a wait to avoid showing PII (Email, phone numbers, digits) etc in exception messages in detectors shown in Applens. (Someone reached out mentioning that some customer's email addresses were being revealed here). 

This is just one small way of masking that information to Applens users.

puneetg1983: We can come up with ways to hide secrets too in future in this class but the existing logic is not complete yet.

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Anonymizer.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Text.RegularExpressions;
+
+namespace Diagnostics.ModelsAndUtils.ScriptUtilities
```
hforeste, line 4: Small nitpik comment to put this instead in the Utilities folder versus ScriptUtilities

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Anonymizer.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    public static class Anonymizer
+    {
+        private const string DigitRegularExpression = @"\d";
+        private const string EmailRegularExpression = @"(?<=[\w]{0})[\w-.+%]*(?=@([\w-]+)[.]{0})";
+        private const string GuidRegularExpression = @"(\{){0,1}[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}(\}){0,1}";
+
+        public static string AnonymizeMessage(string message)
+        {
+            if (string.IsNullOrEmpty(message))
```
hforeste, line 14: Use `string.IsNullOrWhitespace`

hforeste (approved):
Looks good to me, left a few comments.

ShekharGupta1988 (approved):

src/Diagnostics.ModelsAndUtils/ScriptUtilities/Anonymizer.cs
puneetg1983: Hey Hawk - I reason I had it under ScriptUtilities is in case some detector wants to use this class to mask PII data from some other strings so I think it makes sense here. Let me know if you think otherwise.

PR #727, opened 2021-08-24
Darreldonald devops

DarrelDonald:
Added a client and controller to manage DevOps repositories for the detector graduation project
Has functions to retrieve code, view branches, make pull requests, and push changes


src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -0,0 +1,239 @@
...
+        private static string _repoID;
+        private static string _project;
+        private static IConfiguration _config;
+        private static VssCredentials credentials;
+        private static VssConnection connection;
+        private GitHttpClient gitClient;
+
+        public DevOpsClient(IConfiguration config)
+        {
+            _config = config;
```
rekhaswaminathan, line 49: Is `_config` used anywhere else? I see that you are passing constructor parameter `config` to `LoadConfigurations` method

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -0,0 +1,239 @@
...
+            LoadConfigurations(config);
+            credentials = (VssCredentials)(FederatedCredential)new VssBasicCredential("pat", _accessToken);
+            connection = new VssConnection(new Uri($"https://dev.azure.com/{_organization}"), credentials);
```
rekhaswaminathan, line 52: `dev.azure.com` may not be the prefix every time. In some cases it changes based on a team's set up. Eg; it can be msazure.visualstudio.com/<org>/<repo>. I recommend storing the entire domain as config and pull from that. In the future we can save each partner's URL as a config property.

rekhaswaminathan (changes requested):

src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        public DevOpsController(IConfiguration configuration, IDevOpsClient devOpsClient)
+        {
+            _devOpsClient = devOpsClient;
+        }
+
+        [HttpGet(UriElements.DevOpsMakePR)]
+        public async Task<IActionResult> makePullRequestAsync(string sourceBranch, string targetBranch, string title)
+        {
+            DevOpsResponse response = await _devOpsClient.makePullRequestAsync(sourceBranch, targetBranch, title);
```
rekhaswaminathan, line 33: How does AppLens client know which target branch to make PR against?

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
DarrelDonald: It does not. I meant to remove it.

src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
DarrelDonald: It gets it from user input

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
DarrelDonald: I tried accessing a repo on `msazure.visualstudio.com` with the API using `dev.azure.com` and was successful. I believe this format will work even if the domain is different.

rekhaswaminathan: Thats great. Then we can stick to this

src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
```diff
@@ -0,0 +1,58 @@
...
+{
+    [Authorize]
+    [Produces("application/json")]
+    [Route(UriElements.DevOps)]
+    public class DevOpsController : Controller
+    {
+        IDevOpsClient _devOpsClient;
+
+        public DevOpsController(IConfiguration configuration, IDevOpsClient devOpsClient)
+        {
```
rekhaswaminathan, line 26: Are you using the `configuration` parameter anywhere?

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -0,0 +1,237 @@
...
+        string Project { get; }
+    }
+
+    public class DevOpsResponse
+    {
+        public object result;
+        public HttpStatusCode responseCode;
+    }
+
+    class DevOpsClient : IDevOpsClient
```
rekhaswaminathan, line 36: I see that `IRepoClient` is the one with the methods defined. Shouldn't that be the interface here?

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -0,0 +1,237 @@
...
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Threading.Tasks;
+using System.Web;
+
+namespace Diagnostics.RuntimeHost.Services.DevOpsClient
+{
+    public interface IDevOpsClient : IDisposable, IRepoClient
+    {
```
rekhaswaminathan, line 23: What is the purpose of this interface?

src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -160,6 +161,7 @@ public void ConfigureServices(IServiceCollection services)
             {
                 services.AddSingleton<IGithubClient, GithubClient>();
             }
+            services.AddSingleton<IDevOpsClient, DevOpsClient>();
```
rekhaswaminathan, line 164: Shouldn't it be `IRepoClient` here instead of `IDevOpsClient`? 
`IDevOpsClient` seems to have only some properties defined but no methods.

rekhaswaminathan (changes requested):

rekhaswaminathan:
For all the APIs in the new Controller, consider passing the resource URI or at least the Resource Provider & Resource Type as post body object or part of query param because in the future not every resource will end up having same Devops Repo. We have to switch between the Repos based on the service caller or the resource type.

src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        IRepoClient _devOpsClient;
+
+        public DevOpsController(IRepoClient devOpsClient)
+        {
+            _devOpsClient = devOpsClient;
+        }
+
+        [HttpPost(UriElements.DevOpsMakePR)]
+        public async Task<IActionResult> MakePullRequestAsync(string resourceURI, [FromBody]JToken jsonBody)
```
ShekharGupta1988, line 32: can resource Uri also be moved to json body ?

src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            string sourceBranch = jsonBody[$"sourceBranch"].ToString();
+            string targetBranch = jsonBody[$"targetBranch"].ToString();
+            string title = jsonBody[$"title"].ToString();
```
ShekharGupta1988, line 36: This parameter validation in json body is a repeated pattern I see in multiple controllers.
Can you create a utility method to do this.
Something like :

bool ValidateRequestBody(JToken jsonBody, string[] fieldNames, out string outputMessage = null);

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -0,0 +1,221 @@
...
+                List<GitBranchStats> branches = await gitClient.GetBranchesAsync(_project, _repoID);
+                result = branches.Select(x => x.Name).ToList();
+            }
+            catch (Exception ex)
+            {
+                result = ex.ToString();
+            }
+            finally
+            {
+                response.responseCode = gitClient.LastResponseContext.HttpStatusCode;
```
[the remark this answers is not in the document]
rekhaswaminathan: Hey @DarrelDonald , any reason for sending back gitClient HttpStatusCode back to the client? The client would ideally need to parse out only the result object, correct? Is there a need for Runtimehost to simply echo back the `gitClient.LastResponseContext.HttpStatusCode`?

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -0,0 +1,221 @@
...
+                result = await gitClient.CreatePullRequestAsync(pr, _project, _repoID);
+            }
+            catch (Exception ex)
+            {
+                DiagnosticsETWProvider.Instance.LogDevOpsApiException(
+                    resourceUri,
+                    ex.Message,
+                    ex.StackTrace
+                    );
+                throw new VssServiceException();
```
rekhaswaminathan, line 159: Do we need to throw a `new VssServiceException()` here? Can we just rethrow the original exception? This may end up changing the stack trace when its captured by RuntimeHost's default unhandled exception event logging.

src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs
```diff
@@ -282,6 +282,17 @@ public void LogFullAscInsight(string RequestId, string Source, string Message, s
                 WebsiteHostName);
         }
 
+        public void LogDevOpsApiException(string ResourceUri, string Message, string Details, string DiagEnvironment = null, string DiagWebsiteHostName = null)
```
rekhaswaminathan, line 285: I believe this would require the Event attribute & its properties to be given. You can refer to other logging methods to see how its constructed.

rekhaswaminathan (approved):

rekhaswaminathan:
Hey @DarrelDonald, changes look good! I'd get @ShekharGupta1988's approval as well before merging

ShekharGupta1988 (approved):

ShekharGupta1988:
Looks good to me as well. Just shared some comments offline with Darrel regarding the ETW event parameters. Rest all is good :)

PR #729, opened 2021-08-30
Enable query shadowing for kusto provider

yifguoMSFT:
Enable query shadowing for kusto provider to help testing the new clusters under the same or nearly same traffic as our current follower cluster has so that we can compare the performance in a more scientific way

Minor changes include
- add start and end time in kusto query client Id
- add user agent header in observer data provider http client

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviderConfigurations/KustoDataProviderConfiguration.cs
+++ b/src/Diagnostics.DataProviders/DataProviderConfigurations/KustoDataProviderConfiguration.cs
@@ -186,6 +186,20 @@ namespace Diagnostics.DataProviders
         [ConfigurationName("Retry:ExceptionsToRetryFor")]
         public string ExceptionsToRetryFor { get; set; }
 
+        /// <summary>
+        /// List of , separated clusters which requests will be shadowed from
+        /// </summary>
+        [ConfigurationName("QueryShadowingClusterMaping:ClusterNameGroupings")]
+        public string QueryShadowingClusterNameGroupings { get; set; }
+
+        /// <summary>
+        /// List of , separated cluster lists seprated by : which requests will be shadowed to
+        /// </summary>
+        [ConfigurationName("QueryShadowingClusterMaping:TestingClusterNamesGroupings")]
+        public string QueryShadowingTestClustersGroupings { get; set; }
+
+        public ConcurrentDictionary<string, string[]> QueryShadowingClusterMapping;
+
         public List<ITuple> OverridableExceptionsToRetryAgainstLeaderCluster { get; set; }
 
         public IConfiguration config { private get; set; }
@@ -244,6 +258,18 @@ namespace Diagnostics.DataProviders
                     {
                         OverridableExceptionsToRetryAgainstLeaderCluster.Add((ExceptionString, MaxFailureResponseTimeInSeconds));
                     }
+                }
+            }
+
+            if (!string.IsNullOrWhiteSpace(QueryShadowingClusterNameGroupings) && !string.IsNullOrWhiteSpace(QueryShadowingTestClustersGroupings))
+            {
+                var clusterNamesSplited = QueryShadowingClusterNameGroupings.Split(',');
+                var testClusterGroupSplit = QueryShadowingTestClustersGroupings.Split(',');
+                if (clusterNamesSplited.Length == testClusterGroupSplit.Length)
+                {
+                    QueryShadowingClusterMapping = new ConcurrentDictionary<string, string[]>(
+                        Enumerable.Zip(clusterNamesSplited, testClusterGroupSplit)
+                            .Select(p => KeyValuePair.Create(p.First, p.Second.Split(':'))));
                 }
             }
         }
--- a/src/Diagnostics.DataProviders/DataProviderConfigurations/KustoDataProviderConfiguration.cs
+++ b/src/Diagnostics.DataProviders/DataProviderConfigurations/KustoDataProviderConfiguration.cs
@@ -188,12 +188,14 @@ namespace Diagnostics.DataProviders
 
         /// <summary>
         /// List of , separated clusters which requests will be shadowed from
+        /// e.g. "wawswusfollower,wawseusfollower"
         /// </summary>
         [ConfigurationName("QueryShadowingClusterMaping:ClusterNameGroupings")]
         public string QueryShadowingClusterNameGroupings { get; set; }
 
         /// <summary>
         /// List of , separated cluster lists seprated by : which requests will be shadowed to
+        /// e.g. "wustestcluster1:wustestcluster2,eustestcluster1"
         /// </summary>
         [ConfigurationName("QueryShadowingClusterMaping:TestingClusterNamesGroupings")]
         public string QueryShadowingTestClustersGroupings { get; set; }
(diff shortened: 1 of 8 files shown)


src/Diagnostics.DataProviders/KustoSDKClient.cs
```diff
@@ -95,7 +95,16 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
...
                 timeTakenStopWatch.Start();
                 var kustoClient = Client(cluster, database);
-                var result = await kustoClient.ExecuteQueryAsync(database, query, clientRequestProperties);
+                var kustoTask = kustoClient.ExecuteQueryAsync(database, query, clientRequestProperties);
+                if (_config.QueryShadowingClusterMapping != null && _config.QueryShadowingClusterMapping.TryGetValue(cluster, out var shadowClusters))
+                {
+                    foreach (string shadowCluster in shadowClusters)
+                    {
+                        var shadowKustoClient = Client(shadowCluster, database);
+                        shadowKustoClient.ExecuteQueryAsync(database, query, clientRequestProperties);
```
rekhaswaminathan, line 104: What would happen to the original query if the query by `shadowKustoClient` fails/times out for some reason?

src/Diagnostics.DataProviders/DataProviders.cs
```diff
@@ -29,7 +29,9 @@ public DataProviders(DataProviderContext context)
             Kusto = GetOrAddDataProvider(new KustoLogDecorator(context, new KustoDataProvider(_cache,
                      context.Configuration.KustoConfiguration,
                      context.RequestId,
-                     context.KustoHeartBeatService)));
+                     context.KustoHeartBeatService,
+                     context.QueryStartTime,
```
rekhaswaminathan, line 33: Where is the value for `context.QueryStartTime` & `context.QueryEndTime` getting set?

src/Diagnostics.DataProviders/KustoSDKClient.cs
yifguoMSFT: nothing, the task of the async call is not awaited, thus the exception won't be handled. We can check the slow or failed queries in .show query, so I think nothing is need to be logged here at runtimehost side for testing purpose. Let me know if you suggest any changes to this behavior.

hforeste:
Hey @yifguoMSFT , can you give brief description of why the query shadowing work?

src/Diagnostics.DataProviders/DataProviders.cs
yifguoMSFT: Looks like they were set in the `BeginRequestHandle` function of `DiagnosticsRequestMiddleware.cs`. I've verified the timestamps were correct by step debugging and checking the query in kusto
image

yifguoMSFT:
> Hey @yifguoMSFT , can you give brief description of why the query shadowing work?

Sure, added in the PR description

src/Diagnostics.DataProviders/KustoSDKClient.cs
rekhaswaminathan: As long as the original query isn't affected, this approach sounds good to me.

rekhaswaminathan (approved):

src/Diagnostics.DataProviders/DataProviderConfigurations/KustoDataProviderConfiguration.cs
```diff
@@ -186,6 +186,22 @@ public string KustoApiEndpoint
...
+        /// </summary>
+        [ConfigurationName("QueryShadowingClusterMaping:ClusterNameGroupings")]
+        public string QueryShadowingClusterNameGroupings { get; set; }
+
+        /// <summary>
+        /// List of , separated cluster lists seprated by : which requests will be shadowed to
+        /// e.g. "wustestcluster1:wustestcluster2,eustestcluster1"
+        /// </summary>
+        [ConfigurationName("QueryShadowingClusterMaping:TestingClusterNamesGroupings")]
+        public string QueryShadowingTestClustersGroupings { get; set; }
```
nmallick1, line 201: Would it work if only QueryShadowingTestClustersGroupings setting existed and QueryShadowingClusterNameGroupings did not?

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
```diff
@@ -36,6 +36,7 @@ public abstract class SupportObserverDataProviderBase : DiagnosticDataProvider,
             // --------------------------------------------------------------------------
             var client = new HttpClient(handler);
             client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+            client.DefaultRequestHeaders.Add(HeaderConstants.UserAgentHeaderName, "appservice-diagnostics");
```
nmallick1, line 39: Do you want the user-agent to somehow indicate whether the call was from external DiagnoseAndSolve or for a detector executed via AppLens? Do you think something like this will help you debug traffic/load scenarios in future?

nmallick1 (approved):

src/Diagnostics.DataProviders/DataProviderConfigurations/KustoDataProviderConfiguration.cs
yifguoMSFT: No, these strings are used in `PostInitialize` below, and only both of them are not null or empty, then the `QueryShadowingClusterMapping` dictionary will be initialized

src/Diagnostics.DataProviders/DataProviders/Base/SupportObserverDataProviderBase.cs
yifguoMSFT: I think it's OK with this header, as the traffic from Applens and external DianoseAndSolve are handled by different runtimehost endpoints (euaps and non euaps) which connects to different Observer endpoints as well

src/Diagnostics.DataProviders/KustoSDKClient.cs
yifguoMSFT: Found that a malformed cluster name can cause the exception when calling the `Client`. Added a try catch to swallow the exception.

hforeste (approved):

src/Diagnostics.DataProviders/KustoSDKClient.cs
```diff
@@ -95,7 +95,23 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
...
+                        try
+                        {
+                            var shadowKustoClient = Client(shadowCluster, database);
+                            shadowKustoClient.ExecuteQueryAsync(database, query, clientRequestProperties);
+                        }
+                        catch (Exception)
+                        {
+                            // swallow this exception
+                        }
+                    }
```
hforeste, line 112: Something I realize is if a sourceCluster can have say two testClusters then we can potentially use 3x the amount of threads. Or if there is 1 testCluster then we are potentially doubling the amount of threads though 1 task does not always equate to 1 thread.

I am curious on the affects of increased threads on the service. Maybe keep a close watch  of the thread count when this is deployed.

https://applens.azurewebsites.net/subscriptions/0e7b3183-4428-4c1f-93cd-cb2a26616a8a/resourceGroups/RuntimeHost-Prod-RG/providers/Microsoft.Web/sites/diag-runtimehost-prod-bay-001/detectors/threadcount?startTime=2021-08-30T16:59&endTime=2021-08-31T16:47

PR #730, opened 2021-09-02
Rekha/detector deploy

rekhaswaminathan:
Added new Deployment API which clients (Azure Devops Task) can call with required parameters such as commit id or commit range to deploy the changes from the repo. 
Runtimehost will fetch the files modified, compile the detectors to generate dll and save the bits to storage.


src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -112,6 +116,13 @@ public async Task<object> GetFileContentAsync(string filePathInRepo, string reso
 
             try
             {
+                GitRepository repositoryAsync = await gitClient.GetRepositoryAsync(_project, _repoID, (object)null, new CancellationToken());
+                // If branch is not given, use the default of the repo.
+                if (string.IsNullOrWhiteSpace(branch))
+                {
+                    branch = repositoryAsync.DefaultBranch.Replace("refs/heads/", "");
+                }
```
DarrelDonald, line 124: Is this part needed? It uses the default branch when version isn't specified by default.

rekhaswaminathan: Good catch. I missed that. I will change it

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -221,6 +232,160 @@ public async Task<object> PushChangesAsync(string branch, string file, string re
...
+                {
+                    // hack right now, ideally get from config
+                    var detectorId = String.Join(";", Regex.Matches(change.Item.Path, @"\/(.+?)\/")
+                                        .Cast<Match>()
+                                        .Select(m => m.Groups[1].Value));
+
+
+                    var detectorScriptContent = await GetFileContentInCommit(repositoryAsync.Id, change.Item.Path, gitversion);
+                    var packageContent = await GetFileContentInCommit(repositoryAsync.Id, $"/{detectorId}/package.json", gitversion);
+                    var metadataContent = await GetFileContentInCommit(repositoryAsync.Id, $"/{detectorId}/metadata.json", gitversion);
```
hforeste, line 264: These can be ran concurrently

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -64,9 +65,9 @@ public StorageService(IConfiguration configuration, IHostingEnvironment hostingE
                 }
                 var storageAccount = new CloudStorageAccount(new StorageCredentials(accountname, key), accountname, dnsSuffix, true);
                 tableClient = storageAccount.CreateCloudTableClient();
-                containerClient = storageAccount.CreateCloudBlobClient().GetContainerReference(container);
+                containerClient = storageAccount.CreateCloudBlobClient();
```
hforeste, line 68: nitpik; should we from `containerclient` to `cloudblobclient`?

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -126,11 +127,12 @@ public async Task<List<DiagEntity>> GetEntitiesByPartitionkey(string partitionKe
                     } while (tableContinuationToken != null);
                     timeTakenStopWatch.Stop();
                     DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(StorageService), $"GetEntities by Partition key {partitionKey} took {timeTakenStopWatch.ElapsedMilliseconds}, Total rows = {detectorsResult.Count}, ClientRequestId = {clientRequestId} ");
-                    return detectorsResult.Where(result => !result.IsDisabled).ToList();
+                     return startTime == DateTime.MinValue ? detectorsResult.Where(result => !result.IsDisabled).ToList() :
+                        detectorsResult.ToList();
```
hforeste, line 131: Just making sure I understand this correctly, you want to return detectors regardless if they were disabled or not that were created/updated in the last 5 mins?

https://github.com/Azure/Azure-AppServices-Diagnostics/blob/df3b71f148ae7e0021a2ac6b1f3e44c29754bfdc/src/Diagnostics.RuntimeHost/Services/CacheService/DiagEntityTableCacheService.cs#L45

https://github.com/Azure/Azure-AppServices-Diagnostics/blob/df3b71f148ae7e0021a2ac6b1f3e44c29754bfdc/src/Diagnostics.RuntimeHost/Services/SourceWatcher/Watchers/StorageWatcher.cs#L191-L195

src/Diagnostics.RuntimeHost/Controllers/DeploymentController.cs
```diff
@@ -0,0 +1,162 @@
...
+        private IRepoClient devopsClient;
+        protected ICompilerHostClient _compilerHostClient;
+        protected IStorageService storageService;
+        private IInvokerCacheService detectorCache;
+
+        public DeploymentController(IServiceProvider services, IConfiguration configuration, IInvokerCacheService invokerCache)
+        {
+            this.storageService = (IStorageService)services.GetService(typeof(IStorageService));
+            this.devopsClient = (IRepoClient)services.GetService(typeof(IRepoClient));
+            this._compilerHostClient = (ICompilerHostClient)services.GetService(typeof(ICompilerHostClient));
```
hforeste, line 40: Use constructor injection

src/Diagnostics.RuntimeHost/Controllers/DeploymentController.cs
```diff
@@ -0,0 +1,162 @@
...
+                     
+            DeploymentResponse response = new DeploymentResponse();
+            response.DeploymentGuid = Guid.NewGuid().ToString();
+            response.DeployedDetectors = new List<string>();
+            response.FailedDetectors = new Dictionary<string, string>();
+
+            var commitId = deploymentParameters.CommitId;
+            var timeTakenStopWatch = new Stopwatch();
+            timeTakenStopWatch.Start();
+            DiagnosticsETWProvider.Instance.LogRuntimeHostMessage($"Starting deployment operation for {response.DeploymentGuid}");
```
hforeste, line 62: @rekhaswaminathan maybe we should create new events for this since this is a new and pretty big feature.

src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
rekhaswaminathan: Yes that's correct.  This is because people might delete detectors through PR and when we refresh our in-memory cache, we have to get all detectors regardless they were disabled or not within last 5 mins and remove the ones that were deleted.
Currently there is a limitation that deleted detectors won't go away in the UI until the process is restarted.

src/Diagnostics.RuntimeHost/Controllers/DeploymentController.cs
```diff
@@ -0,0 +1,162 @@
...
+            this.devopsClient = (IRepoClient)services.GetService(typeof(IRepoClient));
+            this._compilerHostClient = (ICompilerHostClient)services.GetService(typeof(ICompilerHostClient));           
+            this.detectorCache = invokerCache;
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> Post([FromBody] DeploymentParameters deploymentParameters)
+        {
+            // If all required parameters are empty, reject the request.
+            if(string.IsNullOrWhiteSpace(deploymentParameters.CommitId) && string.IsNullOrWhiteSpace(deploymentParameters.FromCommitId)
```
ShekharGupta1988, line 48: I am assuming there are other invalid combinations as well, like:
- only FromCommitId present
- only StartDate Present or only EndDate Present
- StartDate > EndDate etc
 
 I would suggest add a class Method in the DeploymentParameters Class only called Validate or something and have this logic there.

src/Diagnostics.RuntimeHost/Controllers/DeploymentController.cs
ShekharGupta1988: ya, I agree. Good to create a bunch of events specific to deployment. Easy to manage and alert on

src/Diagnostics.RuntimeHost/Controllers/DeploymentController.cs
```diff
@@ -0,0 +1,162 @@
...
+                        EntityType entityType = EntityType.Detector;
+                        EntityMetadata metaData = new EntityMetadata(file.Content, entityType, null);
+                        var newInvoker = new EntityInvoker(metaData);
+                        newInvoker.InitializeEntryPoint(tempAsm);
+                        detectorCache.AddOrUpdate(detectorId, newInvoker);
+                    }
+                   
+                } else
+                {
+                    DiagnosticsETWProvider.Instance.LogRuntimeHostMessage($"Deployment id {response.DeploymentGuid}, compilation failed for {detectorId}");
```
ShekharGupta1988, line 151: Once you add a new event for this, can we also add the reason for failure (i.e. compilation output)
It might help us in future to debug in case  partners wonder why their detector didnt deployed successfully.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -273,14 +288,29 @@ protected async Task<IActionResult> ExecuteQuery<TPostBodyResource>(TResource re
...
             {
-                if (!jsonBody.References.ContainsKey(p.Key))
+                List<string> gistReferences = DetectorParser.GetLoadDirectiveNames(jsonBody.Script);
+                foreach (string gist in gistReferences)
                 {
-                    // Add latest version to references
-                    jsonBody.References.Add(p);
+                    if(!jsonBody.References.ContainsKey(gist))
+                    {
+                        object gistContent = await devopsClient.GetFileContentAsync($"{gist}/{gist}.csx", resource.ResourceUri, HttpContext.Request.Headers[HeaderConstants.RequestIdHeaderName]);
```
ShekharGupta1988, line 298: Quick question here - For every single ExecuteQuery(i.e. Compilation request), we are getting each gist file code from devops repo, do you forsee us hitting throttling limits on devops pretty frequently? 
Maybe we can consider caching the gist content in memory, however the question would be how to invalidate the cache. Something to think about.

src/Diagnostics.RuntimeHost/Controllers/DeploymentController.cs
```diff
@@ -0,0 +1,162 @@
...
+            {
+                InvocationOutput = new DiagnosticApiResponse()
+            };
+
+            IDictionary<string, string> references = new Dictionary<string, string>();
+
+
+            DiagnosticsETWProvider.Instance.LogRuntimeHostMessage($"Deployment id {response.DeploymentGuid}: {filesToCompile.Count} to compile");
+            //  Compile files
+            foreach( var file in filesToCompile)
```
ShekharGupta1988, line 78: Maybe something you can consider in future to speed up this code (The slowness problem or maybe even throttling will surface for big detector repositories such as ours or AKS and a bunch  of edits by the team): 

- you can split the workload into 2 job tasks: The first task is to handle deleted files and the second task to handle compilation and publish. 
- For deleted files task, instead of updating storage one by one, batch them up and send one request to update.
- For compilation and publish, I see you are compiling each of the files sequentially and then publishing it. This code piece will be hit more regularly.  As a next iteration, you might want to add some degree of parallelism here as well.

What would be nice Rekha here is to keep a close watch (maybe add an email alert as well) on the latency of this API and send an email if latency exceeds a certain dangerous value (like 20 sec, reason - most http clients have 30 sec timeout). If we are seeing too many emails for this, it means we would need to optimize the speed of this API soonish.

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -112,6 +116,13 @@ public async Task<object> GetFileContentAsync(string filePathInRepo, string reso
 
             try
             {
+                GitRepository repositoryAsync = await gitClient.GetRepositoryAsync(_project, _repoID, (object)null, new CancellationToken());
```
ShekharGupta1988, line 119: I think we should use the API cancellation token which has the 60 sec timeout instead of creating new here.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
rekhaswaminathan: Yes Shekhar,  I also thought of implementing a cache at first. I also felt there would be a situation where gist code in the cache might become out of sync with latest code in repo (In case it gets updated via separate PR). For the first iteration, I am taking this approach and will be improving based on usage and metrics (Number of times compilation is called)

src/Diagnostics.RuntimeHost/Controllers/DeploymentController.cs
rekhaswaminathan: For the first two in the list, I have handled them inside inner method that calls the devops API. I'll add validation for last one as well.

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
rekhaswaminathan: This API call was actually not needed (ref: Darrel's comment on the same file), so I've removed it. For other such calls, I will add the time out.

tests/Diagnostics.Tests/TestData/AutoHeal.csx
```diff
@@ -0,0 +1,442 @@
+#load "DetectorUtils"
```
ShekharGupta1988, line 1: QQ - Can we write a test detector code here using gist and not use our actual detector code, to not accidentally leak any platform info (like table names etc)

rekhaswaminathan: Sure,  I can write a sample one.

hforeste (approved):

ShekharGupta1988 (approved):

PR #732, opened 2021-09-07
Observer integration for Worker apps

SharmaAjay19:


michimune:
Just want to make sure. Is the GeoMaster data provider object instantiated with the correct region? I could not find a code change updating it.

SharmaAjay19:
> Just want to make sure. Is the GeoMaster data provider object instantiated with the correct region? I could not find a code change updating it.

Geomaster details are read from request headers in the diagnostic request middleware and that is how the data provider context is initialized.
image

So the correct geomaster details will be passed by the client, e.g. AppLens or Diagnose & Solve etc.

hforeste (approved):

src/Diagnostics.RuntimeHost/Models/PostBodyModel.cs
```diff
@@ -73,4 +73,14 @@ public class BodyValidationResult
...
     }
+
+    public class DiagnosticWorkerAppData
+    {
+        public string WorkerAppName { get; set; }
+        public string Tags { get; set; }
+        public string ResourceGroupName { get; set; }
+        public string SubscriptionName { get; set; }
+        public string KubeEnvironmentName { get; set; }
+        public string GeoMasterInstance { get; set; }
```
yifguoMSFT, line 84: Just a reminder that we just updated the response structure returned by Observer to
image

yifguoMSFT (approved):

src/Diagnostics.RuntimeHost/Models/PostBodyModel.cs
SharmaAjay19: Yes, I will be incorporating these changes soon.

ShekharGupta1988 (approved):

PR #735, opened 2021-09-13
Fixed bug in listing blobs

rekhaswaminathan:
Fixed bug in Listing blobs which is part of `/dependencyCheck`.


src/Diagnostics.RuntimeHost/Services/StorageService/StorageService.cs
```diff
@@ -278,7 +278,8 @@ public async Task<int> ListBlobsInContainer()
                 options.MaximumExecutionTime = TimeSpan.FromSeconds(60);
                 OperationContext oc = new OperationContext();
                 oc.ClientRequestID = clientRequestId;
-                var blobsResult = await cloudBlobClient.ListBlobsSegmentedAsync(null, true, BlobListingDetails.All, 1000, null, options, oc);
```
hforeste, line 281: Was this not listing blobs before?

rekhaswaminathan: It was. I refactored the `clobBlobClient` in my previous PR and forgot to add the statement to get container reference so it broke.

hforeste (approved):

PR #740, opened 2021-10-04
Curibe/resiliency score

cristhianu:
Fixing _name and _customerName getting assigned string.Empty at the beginning of the class instead of in the constructor.


cristhianu:
Resolving conflicts

ShekharGupta1988 (approved):

PR #744, opened 2021-10-15
Accept cert based authentication in compiler host and Search Service

SharmaAjay19:
Security.CertificateAuth classes are taken from our runtimehost classes. Didn't add a direct import to avoid creating dependency on runtimehost which uses different versions of AspNetCore.Authentication module.


src/Diagnostics.CompilerHost/appsettings.json
```diff
@@ -18,5 +18,11 @@
     "AADAuthority": "",
     "ClientId": "", // Azure AAd App id of CompilerHost
     "AllowedAppIds": "" // Azure AAD App id of RuntimeHost
+  },
+  "SecuritySettings": {
+    "AllowedCertIssuers": "",
+    "AllowedCertSubjectNames": "",
+    "ShowIdentityModelErrors": false,
+    "AdditionalAllowedCertSubjectNames": ""
```
rekhaswaminathan, line 26: I don't think we'd need `AdditionalAllowedCertSubjectNames` at this point. Hawk created that for Runtime host for some other reason. Given that CompilerHost will be majorily called by Runtimehost, we can remove this.

rekhaswaminathan:
Where is the code for Runtimehost sending the cert? is that a separate PR?

rekhaswaminathan (approved):

hforeste (approved):

PR #763, opened 2021-11-02
push changes in single commit

DarrelDonald:
when a new detector is created or modified, all the changes will be put into one commit and pushed at the same time

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
@@ -58,8 +58,8 @@ namespace Diagnostics.RuntimeHost.Controllers
             string[] fieldNames =
             {
                 "branch",
-                "file",
-                "repoPath",
+                "files",
+                "repoPaths",
                 "comment",
                 "changeType",
                 "resourceUri"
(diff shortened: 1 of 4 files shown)


src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
```diff
@@ -69,13 +69,13 @@ public async Task<IActionResult> PushChangesAsync([FromBody]JToken jsonBody)
                 return BadRequest(validationMessage);
 
             string branch = jsonBody[$"branch"].ToString();
-            string file = jsonBody[$"file"].ToString();
-            string repoPath = jsonBody[$"repoPath"].ToString();
+            List<string> files = jsonBody[$"files"].Select(file => file.ToString()).ToList();
+            List<string> repoPaths = jsonBody[$"repoPaths"].Select(path => path.ToString()).ToList();
```
rekhaswaminathan, line 73: Why do we need a list for repoPaths? List of files will go into the same repoPath, correct?

DarrelDonald: the path includes the filename. so there will be "/detectorid/detectorid.csx" and "/detectorid/metadata.json"

rekhaswaminathan: Got it. I thought its the parent path. Makes sense.

rekhaswaminathan (approved):

ShekharGupta1988 (approved):

PR #765, opened 2021-11-05
Allow specific PInvoke calls for Certs and Domains detectors

nmallick1:
Update API Block config to allow specific PInvoke calls for Certs and Domains detectors.
Also consider EntryPoint in DllImport attribute while evaluating PInvoke calls.


yifguoMSFT (approved):

src/Diagnostics.Scripts/ApiBlockConfig.json
```diff
@@ -1,14 +1,24 @@
...
           "MessageToShowWhenBlocked": "PInvoke calls are prohibited."
         }
       ]
+    },
+    {
+      "BlockMatchingRegExPattern": "^(dnsapi.dll|dnsapi)\\b",
+      "MessageToShowWhenBlocked": "",
+      "MethodsToBlock": [
+        {
+          "BlockMatchingRegExPattern": "^(?!(DnsQuery_W|DnsRecordListFree)\\b).*",
```
hforeste, line 18: if I understand this correctly, this means block all methods that does not have DnsQuery_W or DnsRecordListFree??

nmallick1: Yes, that is correct. Block call calls other than DnsQuery_W and DnsRecordListFree from dnsapi.dll

src/Diagnostics.Scripts/CompilationService/CustomCodeAnalyzers/BlockCallsAnalyzer.cs
```diff
@@ -106,7 +106,9 @@ private void HandleInvocation(OperationAnalysisContext operationContext)
                         _blockConfig.GetMatchingPInvokeConfig(targetPInvokeDll)?.ForEach(piConfig => {
                             if (piConfig.MethodsToBlock?.Count > 0)
                             {
-                                BlockConfig.GetMatchingBlockMessageList(piConfig.MethodsToBlock, invocation.TargetMethod.Name)?.ForEach(strMessage =>
+                                string pInvokeFunctionName = !string.IsNullOrWhiteSpace(invocation.TargetMethod.GetDllImportData().EntryPointName)? invocation.TargetMethod.GetDllImportData().EntryPointName : invocation.TargetMethod.Name;
```
michimune, line 109: nit: you could write line like this: string pInvokeFunctionName = invocation.TargetMethod.GetDllImportData().EntryPointName ?? invocation.TargetMethod.Name;

michimune (approved):

src/Diagnostics.Scripts/CompilationService/CustomCodeAnalyzers/BlockCallsAnalyzer.cs
nmallick1: @michimune won't ?? check only for NULL and not for whitespace/empty string?

michimune: Sure, you are correct. I am going to dismiss my comment.

hforeste (approved):

PR #768, opened 2021-11-05
updates for demo

DarrelDonald:
adding delete detector
fixing bad requests return 200
added handling for duplicate detector ids

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
@@ -90,15 +90,7 @@ namespace Diagnostics.RuntimeHost.Controllers
         public async Task<IActionResult> GetFileContentAsync(string filePathInRepo, string resourceUri, string branch)
         {
             object response = await _devOpsClient.GetFileContentAsync(filePathInRepo, resourceUri, this.HttpContext.Request.Headers[RequestIdHeaderName], branch);
-
-            if (response.GetType() != typeof(BadRequestObjectResult))
-            {
-                return Ok(response);
-            }
-            else
-            {
-                return (IActionResult)response;
-            }
+            return Ok(response);
         }
 
         [HttpGet(UriElements.DevOpsGetBranches)]
--- a/src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
@@ -76,14 +76,29 @@ namespace Diagnostics.RuntimeHost.Controllers
             string resourceUri = jsonBody[$"resourceUri"].ToString();
 
             object response = await _devOpsClient.PushChangesAsync(branch, files, repoPaths, comment, changeType, resourceUri, this.HttpContext.Request.Headers[RequestIdHeaderName]);
-            return Ok(response);
+            if (response.GetType() != typeof(BadRequestObjectResult))
+            {
+                return Ok(response);
+            }
+            else
+            {
+                return (IActionResult)response;
+            }
         }
 
         [HttpGet(UriElements.DevOpsGetCode)]
         public async Task<IActionResult> GetFileContentAsync(string filePathInRepo, string resourceUri, string branch)
         {
             object response = await _devOpsClient.GetFileContentAsync(filePathInRepo, resourceUri, this.HttpContext.Request.Headers[RequestIdHeaderName], branch);
-            return Ok(response);
+
+            if (response.GetType() != typeof(BadRequestObjectResult))
+            {
+                return Ok(response);
+            }
+            else
+            {
+                return (IActionResult)response;
+            }
         }
 
         [HttpGet(UriElements.DevOpsGetBranches)]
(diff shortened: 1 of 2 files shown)


src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
```diff
@@ -76,14 +76,29 @@ public async Task<IActionResult> PushChangesAsync([FromBody]JToken jsonBody)
             string resourceUri = jsonBody[$"resourceUri"].ToString();
 
             object response = await _devOpsClient.PushChangesAsync(branch, files, repoPaths, comment, changeType, resourceUri, this.HttpContext.Request.Headers[RequestIdHeaderName]);
-            return Ok(response);
+            if (response.GetType() != typeof(BadRequestObjectResult))
```
rekhaswaminathan, line 79: Can you always return `(IActionResult)response` ? Will that work in success scenarios as well?

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -173,27 +175,43 @@ public async Task<object> MakePullRequestAsync(string sourceBranch, string targe
 
             try
             {
-                dictionary[requestId + "--result"] = await gitClient.CreatePullRequestAsync(dictionary[requestId + "--pr"], _project, _repoID);
-            }
-            catch (Exception ex)
-            {
-                if (ex.Message.Contains("An active pull request for the source and target branch already exists"))
+                dictionary[requestId + "--prList"] = await GetPRListAsync(dictionary[requestId + "--source"], dictionary[requestId + "--target"], requestId);
```
rekhaswaminathan, line 178: Quick question regarding this - This dictionary is maintained in-memory per process right? What if the request lands to create a PR lands in a different instance? How do we handle that scenario?

DarrelDonald: do you mean if the request for PR and the request for Push land in different instances? each request gets passed all the data it needs from the ui. they dont communicate with each other

rekhaswaminathan:
Adding some comments from our demo today -
we should not be throwing a BadRequest to the client when active PR exists for the branch, because the client itself is not passing any incorrect data. We should check if active PR exists for that branch and just push the changes to the branch & not create a PR. Devops automatically updates the PR with the latest push ( as we saw in demo today)

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -173,27 +175,43 @@ public async Task<object> MakePullRequestAsync(string sourceBranch, string targe
...
-                    requestId,
-                    resourceUri,
-                    ex.Message,
-                    ex.GetType().ToString(),
-                    ex.StackTrace
-                    );
-                    throw;
+                else
+                {
+                    dictionary[requestId + "--result"] = dictionary[requestId + "--prList"][0];
```
DarrelDonald, line 185: @rekhaswaminathan the change for handling duplicate PRs instead of throwing a BadRequest is here. If its already there i retrieve the existing pr and return that

src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
DarrelDonald: I just tried this out. Trying to cast the success response to IActionResult throws an InvalidCastException. It does still successfully push but UI will get a 500 instead of the success response

rekhaswaminathan: With new change of handling existing active PR, we don't need to check for `BadRequestObjectResult`, correct?

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
rekhaswaminathan: So this is where I initially asked about my question regarding request going to different instance.
Lets consider the following flow -
Author writes a new detector, Make Push + Create PR -> Lands in instance A & after PR creating, in-memory cache dictionary object now has the PR.
Now the author goes to make an edit to an existing branch which would need to update the active PR. This requests lands in instance B. Since you're checking against dictionary cache, how would this work?

rekhaswaminathan: Also, I did not understand how appending requestid to the key works. The requestid keeps changing every time. Can you explain the purpose a bit more.

src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
DarrelDonald: This is in the push. The PR doesnt check for bad request. Push creates a BadRequest when the ID is already in use because duplicate IDs try to create a file that already exists in the repo and that causes failure. the other bad request check i added accidentally. im about to remove that one

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
DarrelDonald: all of the data retrieved from the dictionary is created in the same request it is being accessed by. the dictionary is only there so that multiple requests on the same instance wont change the data that other requests are still using. the values from the request on instance A wont be needed for the request on instance B

rekhaswaminathan (approved):

PR #774, opened 2021-11-16
minor change for allowing heartbeat to be mirrored to testing kusto cluster

yifguoMSFT:


src/Diagnostics.DataProviders/KustoSDKClient.cs
```diff
@@ -96,7 +96,7 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
                 timeTakenStopWatch.Start();
                 var kustoClient = Client(cluster, database);
                 var kustoTask = kustoClient.ExecuteQueryAsync(database, query, clientRequestProperties);
-                if (_config.QueryShadowingClusterMapping != null && _config.QueryShadowingClusterMapping.TryGetValue(cluster, out var shadowClusters) && startTime > DateTime.Parse("2021-11-12 23:59:59"))
+                if (_config.QueryShadowingClusterMapping != null && _config.QueryShadowingClusterMapping.TryGetValue(cluster, out var shadowClusters) && (startTime == null || startTime > DateTime.Parse("2021-11-12 23:59:59")))
```
nmallick1, line 99: If the start time is Null, can we check if the query is actually a heartbeat query before mirroring it?

nmallick1 (approved):

PR #778, opened 2021-11-23
Switch to using System.Text.Json instead of Newtonsoft to improve performance.

puneetg1983:
With this PR, we are changing the default formatted for the RuntimeHost project from NewtonSoft to System.Text.Json (STJ). Few changes were done in the code because System.Text.Json had some limitations or is a bit stricter than NewtonSoft. They are:-

1. STJ had a bug due to which **KeyValuePairs are serialized in to camel case properly**. This was https://github.com/dotnet/runtime/issues/44385 and we had to update STJ to 6.0 to fix it.
2. STJ requires that for serialization properties should be defined. It does not work with public fields only. That's why you see setter and getter added for some fields.
3. The one issue that took the most of my time was classes **inheriting from Attribute** fail with a really bad error and this is a bug which is still open https://github.com/dotnet/runtime/issues/58947
4. Multi dimensional arrays are **not** supported so we had to use the DataTableResponseObject to use jagged arrays.

**This is still a risky PR** because I cannot test all API combinations locally. I plan to test this out in staging environment once the PR is merged and if things don't look fine, I will revert the PR.


src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -145,20 +146,20 @@ public void ConfigureServices(IServiceCollection services)
...
+                    options.JsonSerializerOptions.WriteIndented = true;
                 });
             }
             else
             {
                 services.AddControllers().AddJsonOptions(options =>
                 {
                     options.JsonSerializerOptions.WriteIndented = true;
-                }).AddNewtonsoftJson(options =>
+                }).AddJsonOptions(options =>
```
rekhaswaminathan, line 159: CompilerHost Startup.cs also has `.AddNewtonsoftJson()` method. Should we change that as well? https://github.com/Azure/Azure-AppServices-Diagnostics/blob/b31dd4f95cacf16e81035c2a168136ff386d7acd/src/Diagnostics.CompilerHost/Startup.cs#L155

src/Diagnostics.RuntimeHost/Services/CacheService/DiagEntityTableCacheService.cs
```diff
@@ -42,8 +42,8 @@ private async void StartPolling()
                 DiagnosticsETWProvider.Instance.LogAzureStorageMessage(nameof(DiagEntityTableCacheService), "Start polling Azure Storage for refreshing cache");
                 try
                 {
-                    var detectorTask = storageService.GetEntitiesByPartitionkey("Detector", startUp ? DateTime.MinValue : DateTime.UtcNow.AddMinutes(-5));
-                    var gistTask = storageService.GetEntitiesByPartitionkey("Gist", startUp ? DateTime.MinValue : DateTime.UtcNow.AddMinutes(-5));
+                    var detectorTask = storageService.GetEntitiesByPartitionkey("Detector", startUp ? DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Utc) : DateTime.UtcNow.AddMinutes(-5));
+                    var gistTask = storageService.GetEntitiesByPartitionkey("Gist", startUp ? DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Utc) : DateTime.UtcNow.AddMinutes(-5));
```
rekhaswaminathan, line 46: Thanks for taking care of this 👍

nmallick1:
@puneetg1983 , does the Response JSON change in anyway so as to require a change in swagger?

src/Diagnostics.RuntimeHost/Startup.cs
puneetg1983: That is not required. We are just changing the serializer in the runtime host project.

puneetg1983:
> @puneetg1983 , does the Response JSON change in anyway so as to require a change in swagger?

I tested only the listdetectors and invoke detector call and ensured that they are exactly the same. But we have lot of other WEBAPI too and I dont know the best way to test all of them :(

hforeste:
> > @puneetg1983 , does the Response JSON change in anyway so as to require a change in swagger?
> 
> I tested only the listdetectors and invoke detector call and ensured that they are exactly the same. But we have lot of other WEBAPI too and I dont know the best way to test all of them :(

Hey @puneetg1983 , do we believe anything else to be different if ListDetectors and ExecuteDetector calls are fine?

puneetg1983:
> > > @puneetg1983 , does the Response JSON change in anyway so as to require a change in swagger?
> > 
> > 
> > I tested only the listdetectors and invoke detector call and ensured that they are exactly the same. But we have lot of other WEBAPI too and I dont know the best way to test all of them :(
> 
> Hey @puneetg1983 , do we believe anything else to be different if ListDetectors and ExecuteDetector calls are fine?

Hey @hforeste , I do see a bunch of other controllers and we should validate once that they are working as expected. For e.g. I saw yesterday that the Observer controller is failing and thats because it was returning a JArray and that is expected to fail. I quickly glanced at other controllers and they dont seem to be using classes from NewtwonSoft but its better if we can check them all one.

hforeste:
Hey @puneetg1983 , added PR to remove the obsolete ObserverController. https://github.com/Azure/Azure-AppServices-Diagnostics/pull/780/files

src/Diagnostics.ModelsAndUtils/Attributes/Definition.cs
```diff
@@ -56,17 +57,38 @@ public Definition()
...
 
-        [JsonIgnore]
+        /// <summary>
+        /// Property created only for Json Serialization as Attributes
+        /// are not serialized today properly by System.Text.Json
+        /// https://github.com/dotnet/runtime/issues/58947
+        /// </summary>
+        [DataMember]
+        [JsonPropertyName("supportTopicList")]
+        public IEnumerable<SupportTopicJson> SupportTopicListJson
```
hforeste, line 73: Small nitpik, but maybe rename this as `SupportTopicListSTJ` or `SupportTopicListSTJCompat` since the return type is technically not json type.

src/Diagnostics.ModelsAndUtils/Models/DataTableResponseObject.cs
```diff
@@ -125,14 +125,11 @@ public static DataTableResponseObject ToDataTableResponseObject(this DataTable t
                 columns.Add(new DataTableResponseColumn() { ColumnName = col.ColumnName, DataType = col.DataType.ToString().Replace("System.", "") });
             }
 
-            var rows = new dynamic[table.Rows.Count, table.Columns.Count];
+            var rows = new dynamic[table.Rows.Count][];
```
hforeste, line 128: new dynamic[table.Rows.Count] [table.Columns.Count]??

src/Diagnostics.ModelsAndUtils/Models/ResponseExtensions/Downtime.cs
```diff
@@ -10,12 +10,12 @@ public class DownTime
         /// <summary>
         /// Represents the start time for the downtime period
         /// </summary>
-        public DateTime StartTime { get; set; } = DateTime.MinValue;
+        public DateTime StartTime { get; set; } = DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Utc);
```
hforeste, line 13: Is there a reason we can't simply use `DateTime.MinValue`?

src/Diagnostics.ModelsAndUtils/Models/DataTableResponseObject.cs
puneetg1983: You cannot do this for C# Jagged arrays. The compiler doesn't allow this.

https://stackoverflow.com/a/4384236/1900166

src/Diagnostics.ModelsAndUtils/Attributes/Definition.cs
puneetg1983: done

hforeste (approved):

PR #782, opened 2021-12-01
Fix geomaster endpoint for container apps

michimune:


src/Diagnostics.DataProviders/GeoMasterCertClient.cs
```diff
@@ -30,9 +31,14 @@ public GeoMasterCertClient(GeoMasterDataProviderConfiguration configuration, str
             var geoEndpoint = new UriBuilder(geoMasterHostName)
             {
                 Scheme = "https",
-                Port = GeoMasterCsmApiPort
             };
 
+            // Use the default CSM port if a port number isn't specified at the end
+            if (!Regex.IsMatch(geoMasterHostName, @":\d+/?$"))
+            {
+                geoEndpoint.Port = GeoMasterCsmApiPort;
```
hforeste, line 39: Is there a different port number for containerapps?

michimune: Yes, that's the fix we need.

hforeste (approved):

yifguoMSFT (approved):

PR #786, opened 2021-12-04
Fix broken compile response

michimune:
This PR fixes the empty response from runtimehost when compiling a detector.

This PR also changes the following things:
- Add ConfigureAwait(false) for some async calls
- Disable Application Insights for local development mode
- Fix unhandled exception when launching runtime host if CloudDomain is set to UnitTest

### DIFF
--- a/src/Diagnostics.CompilerHost/Startup.cs
+++ b/src/Diagnostics.CompilerHost/Startup.cs
@@ -150,8 +150,12 @@ namespace Diagnostics.CompilerHost
                                         AddAuthenticationSchemes(CertificateAuthDefaults.AuthenticationScheme, "AzureAd")
                                         .Build();
             });
-            // Enable app insights telemetry
-            services.AddApplicationInsightsTelemetry();
+
+            if (!Environment.IsDevelopment())
+            {
+                // Enable app insights telemetry
+                services.AddApplicationInsightsTelemetry();
+            }
             services.AddControllers().AddNewtonsoftJson();
             CustomStartup();
 
--- a/src/Diagnostics.DataProviders/TokenService/LogAnalyticsTokenServiceBase.cs
+++ b/src/Diagnostics.DataProviders/TokenService/LogAnalyticsTokenServiceBase.cs
@@ -84,7 +84,7 @@ namespace Diagnostics.DataProviders.TokenService
                 ValidateAuthority = true
             };
 
-            creds = await ApplicationTokenProvider.LoginSilentAsync(domain, clientId, clientSecret, adSettings);
+            creds = await ApplicationTokenProvider.LoginSilentAsync(domain, clientId, clientSecret, adSettings).ConfigureAwait(false);
         }
     }
 }
--- a/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
@@ -348,7 +348,7 @@ namespace Diagnostics.RuntimeHost.Controllers
             bool isCompilationNeeded = !ScriptCompilation.IsSameScript(jsonBody.Script, scriptETag) || !_assemblyCacheService.IsAssemblyLoaded(assemblyFullName);
             if (isCompilationNeeded)
             {
-                queryRes.CompilationOutput = await _compilerHostClient.GetCompilationResponse(jsonBody.Script, jsonBody.EntityType, jsonBody.References, runtimeContext.OperationContext.RequestId);
+                queryRes.CompilationOutput = await _compilerHostClient.GetCompilationResponse(jsonBody.Script, jsonBody.EntityType, jsonBody.References, runtimeContext.OperationContext.RequestId).ConfigureAwait(false);
             }
             else
             {
--- a/src/Diagnostics.RuntimeHost/Extensions/StorageServiceExtensions.cs
+++ b/src/Diagnostics.RuntimeHost/Extensions/StorageServiceExtensions.cs
@@ -27,10 +27,12 @@ namespace Microsoft.Extensions.DependencyInjection
             }
             if (configuration.IsAzureChinaCloud() || configuration.IsAzureUSGovernment())
             {
+                // Sovereign clouds cannot access our StorageService
                 services.AddSingleton<IStorageService, NullableStorageService>();
             }
             else
             {
+                // IStorageService dependency needs to be initialized for any other cloud including UnitTest
                 services.AddSingleton<IStorageService, NullableStorageService>();
             }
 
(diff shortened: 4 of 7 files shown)


src/Diagnostics.CompilerHost/Startup.cs
```diff
@@ -150,8 +150,12 @@ public void ConfigureServices(IServiceCollection services)
                                         AddAuthenticationSchemes(CertificateAuthDefaults.AuthenticationScheme, "AzureAd")
                                         .Build();
             });
-            // Enable app insights telemetry
-            services.AddApplicationInsightsTelemetry();
+
+            if (!Environment.IsDevelopment())
+            {
```
puneetg1983, line 155: While this looks good to me, I am not too sure if someone is relying on App Insights Telemetry from the dev environement. @rekhaswaminathan do you know if AI is used for DEV environment at all?

src/Diagnostics.RuntimeHost/Extensions/StorageServiceExtensions.cs
```diff
@@ -29,6 +29,10 @@ public static IServiceCollection AddDiagEntitiesStorageService(this IServiceColl
             {
                 services.AddSingleton<IStorageService, NullableStorageService>();
             }
+            else
+            {
+                services.AddSingleton<IStorageService, NullableStorageService>();
+            }
```
puneetg1983, line 35: You can just live with the else and remove the extra if statement 

`if (configuration.IsAzureChinaCloud() || configuration.IsAzureUSGovernment())`

src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -150,18 +150,19 @@ public void ConfigureServices(IServiceCollection services)
                 }).AddJsonOptions(options =>
                 {
                     options.JsonSerializerOptions.WriteIndented = true;
+                    options.JsonSerializerOptions.IncludeFields = true;
```
puneetg1983, line 153: >IncludeFields [](http://example.com/codeflow?start=50&length=13)

very cool change. I was literally changing the fields to properties in other places in the code :D

puneetg1983 (approved):
Thanks @michimune for fixing this!

src/Diagnostics.RuntimeHost/Startup.cs
hforeste: How will this show now in applens output?

src/Diagnostics.RuntimeHost/Extensions/StorageServiceExtensions.cs
hforeste: Hey @michimune , when is it will we set `CloudDomain` as `UnitTest`? Are we doing this somewhere in an actual unit test?

src/Diagnostics.CompilerHost/Startup.cs
rekhaswaminathan: We don't rely on App Insights telemetry for dev environment

src/Diagnostics.DataProviders/TokenService/LogAnalyticsTokenServiceBase.cs
```diff
@@ -84,7 +84,7 @@ private async Task Authenticate()
                 ValidateAuthority = true
             };
 
-            creds = await ApplicationTokenProvider.LoginSilentAsync(domain, clientId, clientSecret, adSettings);
+            creds = await ApplicationTokenProvider.LoginSilentAsync(domain, clientId, clientSecret, adSettings).ConfigureAwait(false);
```
rekhaswaminathan, line 87: Do we need `ConfigureAwait(false)` here?

src/Diagnostics.CompilerHost/Startup.cs
michimune: Thank you for confirmation. I am resolving this comment.

src/Diagnostics.RuntimeHost/Startup.cs
michimune: Without this change, the response from compiler host is empty. Pressing Run never succeeds.

src/Diagnostics.RuntimeHost/Controllers/DiagnosticControllerBase.cs
```diff
@@ -348,7 +348,7 @@ protected async Task<IActionResult> ExecuteQuery<TPostBodyResource>(TResource re
             bool isCompilationNeeded = !ScriptCompilation.IsSameScript(jsonBody.Script, scriptETag) || !_assemblyCacheService.IsAssemblyLoaded(assemblyFullName);
             if (isCompilationNeeded)
             {
-                queryRes.CompilationOutput = await _compilerHostClient.GetCompilationResponse(jsonBody.Script, jsonBody.EntityType, jsonBody.References, runtimeContext.OperationContext.RequestId);
+                queryRes.CompilationOutput = await _compilerHostClient.GetCompilationResponse(jsonBody.Script, jsonBody.EntityType, jsonBody.References, runtimeContext.OperationContext.RequestId).ConfigureAwait(false);
```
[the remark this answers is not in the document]
michimune: It's always good to have ConfigureAwait(false) for async calls in background tasks to avoid deadlocks. VS suggested it as well.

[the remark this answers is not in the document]
michimune: From debugging perspective, without ConfigureAwait(false) here, when you press F10 at this line, VS doesn't stop at the next line.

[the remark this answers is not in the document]
hforeste: So this wouldn't be a background task for this API. Its results are immediately needed.

src/Diagnostics.RuntimeHost/Startup.cs
hforeste: Oh so if the compilation output is empty then its considered a failed compilation? Sorry I can't remember

src/Diagnostics.DataProviders/TokenService/LogAnalyticsTokenServiceBase.cs
hforeste: Michi so this part of this code is not run as background task but the StartTokenRefresh method which is being called somewhere else in runtimehost code might be better suited to have the `ConfigureAwait(false)` as that is run as a background task.

hforeste: https://github.com/Azure/Azure-AppServices-Diagnostics/pull/786/files#diff-158490db55e5872149db5f22ad70d157ac7306decd376acb31ea58a33061daffR39

src/Diagnostics.RuntimeHost/Extensions/StorageServiceExtensions.cs
michimune: I checked again and found it in AzureStorageTests.cs. Although the test code is currently commented out, I think it still makes sense to keep the UnitTest logic.
@puneetg1983 the reason I just added the else statement was to keep the intension of the original code. But I will add a comment to this logic so it makes more sense.

michimune: Comments are added.

PR #788, opened 2021-12-08
Support for AAD subject name issue authentication 

nmallick1:
Add support to acquire token via certificate for AAD subject name + issuer authentication (SN+I) with token cache and token auto refresh.
Switch Kusto and ASC data providers to SN+I authentication
Add generic cert loader service. This supplies certs that are passed to AAD.
Auto configuration of dev environment with certificates for local dev to remain uninterrupted.
Code cleanup.
Addition of a new Http data provider, helps test the token service while switching additional data providers to use SN+I authentication. This also enables scenarios where partners wanted to call into their observer counterpart from detectors.


src/Diagnostics.DataProviders/DataProviders/HttpDataProvider.cs
```diff
@@ -0,0 +1,275 @@
...
+            {
+                ProviderName = "Http"
+            };
+        }
+
+        #region GET Requests
+
+        #region Anonymous GET request
+        /// <inheritdoc/>
+        public async Task<T> SendGetRequestAnonyousAsync<T>(Uri requestUri, Dictionary<string, string> additionalHeaders = null, string operationName = "", CancellationToken cancellationToken = default(CancellationToken))
```
ShekharGupta1988, line 51: nit : you meant 'Anonymous' ?

src/Diagnostics.RuntimeHost/Program.cs
```diff
@@ -64,7 +63,8 @@ public static IHostBuilder CreateHostBuilder(string[] args)
                             .AddAzureKeyVault(
                                 keyVaultUri,
                                 keyVaultClient,
-                                new DefaultKeyVaultSecretManager());
+                                new DefaultKeyVaultSecretManager())
+                            .AddAndInstallAzureKeyVaultCertificates(keyVaultUri, keyVaultClient);
```
rekhaswaminathan, line 67: Just so I understand, in production, the certs will be synced with Prod keyvault, correct?

nmallick1: Nope, the certs would have to be installed on the web app (App Service). Antares takes care of setting up the cert store for us and removes the KV dependency. Before the website starts, cert store is already set up and the app should have all that it needs to contact AAD. 

This piece is us mimicking Antares capability of syncing certificates for our local dev to work. 

I have already manually installed the certs required for Kusto, ASC and Http data provider on each of our runtimehost web apps.

hforeste (approved):

PR #790, opened 2021-12-08
Restrict the use of SuppressMessage attribute and #pragma tag

nmallick1:
Restrict the use of SuppressMessage attribute and #pragma tag. 
Also mask the actual rule id to leave a false trail.

### DIFF
--- a/src/Diagnostics.Scripts/EntityInvoker.cs
+++ b/src/Diagnostics.Scripts/EntityInvoker.cs
@@ -20,6 +20,7 @@ using Diagnostics.Scripts.Models;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Scripting;
 using Diagnostics.Scripts.CompilationService.Utilities.Extensions;
+using System.Text.RegularExpressions;
 
 namespace Diagnostics.Scripts
 {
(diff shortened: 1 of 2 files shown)


hforeste (approved):

src/Diagnostics.Scripts/EntityInvoker.cs
```diff
@@ -395,25 +396,32 @@ private void Validate()
...
+                if (this._entityMetaData.ScriptText.Contains("SuppressMessage(", StringComparison.OrdinalIgnoreCase))
+                {
+                    Regex suppressMessageRegex = new Regex(@"\[(.*)SuppressMessage\((.*)\)\]", RegexOptions.IgnoreCase);
+                    if (suppressMessageRegex.Match(this._entityMetaData.ScriptText).Success)
+                    {
+                        throw new ScriptCompilationException("ERROR : Use of System.Diagnostics.CodeAnalysis.SuppressMessage attribute is prohibited.");
+                    }
+                }
+
+                if (this._entityMetaData.ScriptText.Contains("pragma", StringComparison.OrdinalIgnoreCase))
```
hforeste, line 408: @nmallick1 remind me again what one can do with pragma?

ShekharGupta1988 (approved):

src/Diagnostics.Scripts/EntityInvoker.cs
nmallick1: @hforeste You can configure pragma tag to suppress code analysis generated errors and warnings. A malicious user can use pragma to suppress the error we generate to block restricted API's. Once the error is suppressed, they are free to use any method calls. Essentially use pragma to mimic the behavior of he SuppressMessage attribute.

PR #797, opened 2021-12-10
Resource Provider Config for Devops

rekhaswaminathan:


src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
```diff
@@ -99,5 +94,17 @@ public async Task<IActionResult> GetBranchesAsync(string resourceUri)
             object response = await _devOpsClient.GetBranchesAsync(resourceUri, this.HttpContext.Request.Headers[RequestIdHeaderName]);
             return Ok(response);
         }
+
+        [HttpGet(UriElements.DevOpsConfig)]
+        public async Task<IActionResult> GetResourceProviderConfig(string resourceProviderType)
```
ShekharGupta1988, line 99: Thinking out loud here from security standpoint:
Do we need to pass the entire config as an API response to client (in this case applens).
This config might have default approvers aliases or groups in future.  Maybe we just need the default branch for the UI purpose.

From Applens standpoint, even though user needs to be authenticated but once you are authenticated(even bad actor), today you can execute any runtime host controller API via Applens Backend (and Durian is not applicable here since this is not case specific data)

src/Diagnostics.RuntimeHost/Services/DevOpsClient/DevOpsClient.cs
```diff
@@ -52,6 +52,49 @@ private void LoadConfigurations(IConfiguration config)
...
+
+        private async Task LoadResourceProviderConfigStorage()
+        {
+            byte[] repoConfigBuffer = await storageService.GetResourceProviderConfig();
+            string repoConfig = Encoding.UTF8.GetString(repoConfigBuffer, 0, repoConfigBuffer.Length);
+            var resourceProviderRepoConfigs = JsonConvert.DeserializeObject<List<ResourceProviderRepoConfig>>(repoConfig);
+            foreach (var resourceProviderRepoConfig in resourceProviderRepoConfigs)
+            {
+                string resourceProvider = resourceProviderRepoConfig.ResourceProvider.ToLower();
+                if (!string.IsNullOrEmpty(resourceProvider) && !resourceProviderMapping.ContainsKey(resourceProvider))
```
ShekharGupta1988, line 66: nit: Use IsNullOrWhiteSpace

src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
rekhaswaminathan: Agreed. This was mainly for checking if Resource Provider type has been enabled for graduation or not. I can make changes to just return bool `true` or `false`

DarrelDonald (approved):

ShekharGupta1988 (approved):

PR #798, opened 2021-12-10
KustSDKClient uses test leader cluster

yifguoMSFT:
If test cluster exists, KustSDKClient will use test leader cluster first and falls back to normal routine on any failure

Besides, add a `GetAggHiPerfClusterNameByStampAsync` function to KustoDataProvider

### DIFF
--- a/src/Diagnostics.DataProviders/KustoSDKClient.cs
+++ b/src/Diagnostics.DataProviders/KustoSDKClient.cs
@@ -95,10 +95,11 @@ namespace Diagnostics.DataProviders
             {
                 timeTakenStopWatch.Start();
                 var kustoClient = Client(cluster, database);
-                var kustoTask = kustoClient.ExecuteQueryAsync(database, query, clientRequestProperties);
+                Task<IDataReader> kustoTask = null;
+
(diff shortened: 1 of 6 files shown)


src/Diagnostics.DataProviders/KustoSDKClient.cs
```diff
@@ -111,15 +112,40 @@ public async Task<DataTable> ExecuteQueryAsync(string query, string cluster, str
                                     shadowClientRequestProperties.SetOption(ClientRequestProperties.OptionQueryConsistency, ClientRequestProperties.OptionQueryConsistency_Strong);
                                 }
                                 var shadowKustoClient = Client(shadowCluster, database);
-                                shadowKustoClient.ExecuteQueryAsync(database, query, shadowClientRequestProperties);
+                                kustoTask = shadowKustoClient.ExecuteQueryAsync(database, query, shadowClientRequestProperties)
```
nmallick1, line 115: @yifguoMSFT , will this create a log in the DiagnosticRole table when the query is fired against the new leader cluster?

yifguoMSFT: It should, as I changed the cluster variable in line 125 if the task succeeds and will be logged in line 171. But I am not very sure what will happen after `LogKustoQuery` is called at line 171, just assumed it will create a log

nmallick1 (approved):

PR #799, opened 2021-12-10
Fix runtime exception incorrectly shown in the runtime logs

michimune:
This PR fixes an issue where runtime exception isn't correctly shown in the runtime output text when compiling a detector.

### DIFF
--- a/src/Diagnostics.RuntimeHost/SerializationConvertors/AsRuntimeTypeConverter.cs
+++ b/src/Diagnostics.RuntimeHost/SerializationConvertors/AsRuntimeTypeConverter.cs
@@ -19,7 +19,7 @@ namespace Diagnostics.RuntimeHost
                 WriteIndented = true
             };
 
-            options.Converters.Add(new AsPrintableTypeConverter<Exception>());
+            options.Converters.Add(new ExceptionConverter());
 
             return options;
         });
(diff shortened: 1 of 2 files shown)


src/Diagnostics.RuntimeHost/SerializationConvertors/AsRuntimeTypeConverter.cs
```diff
@@ -42,22 +42,23 @@ public override void Write(Utf8JsonWriter writer, T value, JsonSerializerOptions
...
-        public override void Write(Utf8JsonWriter writer, T value, JsonSerializerOptions options)
+        public override void Write(Utf8JsonWriter writer, Exception value, JsonSerializerOptions options)
         {
-            JsonSerializer.Serialize(writer, value.ToString(), _serializeWithoutFieldOptions);
+            writer.WriteStartObject();
+
+            writer.WriteString("ClassName", value?.GetType()?.FullName);
+            writer.WriteString("Message", value?.Message);
+            writer.WriteString("Source", value?.Source);
+            writer.WriteString("StackTraceString", value?.StackTrace);
```
puneetg1983, line 59: Do you also want to log Inner exception of they are not null? Maybe check to 2-3 levels instead of logging all inner exceptions.

michimune: Good point. The portal code isn't actually using the InnerException property, but it's better to have it. It's a little tricky to check the level, so I just added the whole.

puneetg1983: nice addition.

puneetg1983 (approved):

PR #802, opened 2021-12-15
keep validation function

DarrelDonald:


rekhaswaminathan (approved):

src/Diagnostics.RuntimeHost/Utilities/RequestBodyValidator.cs
```diff
@@ -20,5 +23,25 @@ public static bool ValidateRequestBody(JToken jsonBody, string[] fieldNames, out
...
+            {
+                outputMessage = "Missing request body";
+                return false;
+            }
+
+            foreach (var property in body.GetType().GetProperties())
+            {
+                if (property.GetValue(body) == null)
+                {
+                    outputMessage = $"Missing {outputMessage += property.Name} from request body. ";
```
michimune, line 40: This will generate a malformed message when multiple properties are missing. Don't do outputMessage += property.Name while assigning to outputMessage.
I would suggest to use a List object to store missing property names during the foreach loop, and create the outputMessage after the loop, like:

outputMessage = $"Missing {String.Join(", ", missingProperties)} from request body.";

src/Diagnostics.RuntimeHost/Utilities/RequestBodyValidator.cs
```diff
@@ -20,5 +23,32 @@ public static bool ValidateRequestBody(JToken jsonBody, string[] fieldNames, out
...
+                    missingProperties.Add(property.Name);
+                }
+            }
+
+
+            var message = "The following fields are missing from the request body: ";
+            foreach (var x in missingProperties) message += x + ", ";
+            outputMessage = message.Remove(message.Length - 2);
+
+            return !(missingProperties.Count > 0);
```
michimune, line 51: Two points:
1. You can check if Count > 0 and return true before crafting outputMessage. This way you can save cost when validated successfully.
2. For line 47 - 49, you could write this way: outputMessage = message + String.Join(", ", missingProperties);
String.Join() is quite useful because it inserts delimiters between elements. You don't need Remove(message.Length - 2).

src/Diagnostics.RuntimeHost/Utilities/RequestBodyValidator.cs
```diff
@@ -20,5 +23,34 @@ public static bool ValidateRequestBody(JToken jsonBody, string[] fieldNames, out
...
+
+            foreach (var property in body.GetType().GetProperties())
+            {
+                if (property.GetValue(body) == null)
+                {
+                    missingProperties.Add(property.Name);
+                }
+            }
+
+            if (!(missingProperties.Count() > 0))
```
michimune, line 46: nit: it's ok with: if (missingProperties.Count() == 0) because Count() never returns a negative value.

src/Diagnostics.RuntimeHost/Utilities/RequestBodyValidator.cs
```diff
@@ -20,5 +23,34 @@ public static bool ValidateRequestBody(JToken jsonBody, string[] fieldNames, out
...
+                return true;
+
+
+            var message = "The following fields are missing from the request body: ";
+            outputMessage = message + String.Join(", ", missingProperties);
+
+            return !(missingProperties.Count > 0);
```
michimune, line 53: Since Count() is guaranteed to be > 0 here, you could simply return false;

michimune (approved):

PR #807, opened 2022-01-02
Language server code and deployment

SharmaAjay19:
Language server will be hosted on App Service Linux Container. This PR is the code for that server, the dockerfile that builds the image for container, and the devops pipeline that will pick latest dlls after our Runtimehost backend build succeeds to pass onto the language server.

### DIFF
--- a/src/LanguageServer/azure-pipelines.yml
+++ b/src/LanguageServer/azure-pipelines.yml
@@ -28,11 +28,11 @@ jobs:
   - task: ExtractFiles@1
     inputs:
       archiveFilePatterns: '$(Agent.BuildDirectory)/backend-drop1/Diagnostics.RuntimeHost.zip'
-      destinationFolder: '$(Agent.BuildDirectory)/applensdlls'
+      destinationFolder: '$(Agent.BuildDirectory)/runtimehostdlls'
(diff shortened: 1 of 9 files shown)


src/LanguageServer/azure-pipelines.yml
```diff
@@ -0,0 +1,53 @@
...
+      pipeline: '12'
+      buildVersionToDownload: 'latest'
+      downloadType: 'single'
+      artifactName: 'backend-drop1'
+      downloadPath: '$(Agent.BuildDirectory)'
+
+  - task: ExtractFiles@1
+    inputs:
+      archiveFilePatterns: '$(Agent.BuildDirectory)/backend-drop1/Diagnostics.RuntimeHost.zip'
+      destinationFolder: '$(Agent.BuildDirectory)/applensdlls'
```
hforeste, line 31: nitpik; instead of applensdlls should we rename this to go to runtimehostdlls?

SharmaAjay19: sure we can. though it's a temporary folder and isn't really used anywhere.

hforeste (approved):

PR #809, opened 2022-01-05
add ExecuteQueryOnHiPerfClusterWithBackup to KustoDataProvider

yifguoMSFT:


src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -288,5 +288,18 @@ public async Task<string> GetAggHiPerfClusterNameByStampAsync(string stampName)
                 return null;
             }
         }
+
+        public async Task<DataTable> ExecuteQueryOnHiPerfClusterWithBackup(string aggQuery, string backupQuery, string stampName, string requestId = null, string operationName = null) 
+        {
+            string cluster = await GetAggHiPerfClusterNameByStampAsync(stampName);
+            if (cluster == null)
```
nmallick1, line 295: nit: Can we do !string.IsNullOrWhiteSpace(cluster)?

```suggestion
            if (!string.IsNullOrWhiteSpace(cluster))

```

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -288,5 +288,18 @@ public async Task<string> GetAggHiPerfClusterNameByStampAsync(string stampName)
                 return null;
             }
         }
+
+        public async Task<DataTable> ExecuteQueryOnHiPerfClusterWithBackup(string aggQuery, string backupQuery, string stampName, string requestId = null, string operationName = null)
```
nmallick1, line 292:
```suggestion
        public async Task<DataTable> ExecuteQueryOnHiPerfClusterWithBackup(string aggQuery, string fallbackQuery, string stampName, string requestId = null, string operationName = null) 
```

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
yifguoMSFT: sure, fixe4d

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
yifguoMSFT: updated

nmallick1 (approved):

PR #813, opened 2022-01-13
Enable MDM for container apps

michimune:
This PR allows detectors to read container apps metrics from MDM.

### DIFF
--- a/src/Diagnostics.DataProviders/ConfigurationFactory.cs
+++ b/src/Diagnostics.DataProviders/ConfigurationFactory.cs
@@ -90,6 +90,21 @@ namespace Diagnostics.DataProviders
                         return string.Empty;
                 }
             }
+            else if (prefix == "ContainerAppsMdm")
+            {
+                switch (name)
+                {
+                    case "MdmShoeboxEndpoint":
+                        return "https://az-int.metrics.nsatc.net";
+                    case "MdmRegistrationCertThumbprint":
+                        // Replace the thumbprint with the certificate installed in your machine.
+                        return "";
+                    case "MdmShoeboxAccount":
+                        return "Mock";
+                    default:
+                        return string.Empty;
+                }
+            }
             else if (prefix == "GeoMaster")
             {
                 switch (name)
--- a/src/Diagnostics.RuntimeHost/Properties/launchSettings.json
+++ b/src/Diagnostics.RuntimeHost/Properties/launchSettings.json
@@ -3,7 +3,8 @@
     "Diagnostics.RuntimeHost": {
       "commandName": "Project",
       "environmentVariables": {
-        "ASPNETCORE_ENVIRONMENT": "Development"
+        "ASPNETCORE_ENVIRONMENT": "Development",
+        "AzureServicesAuthConnectionString": "RunAs=Developer; DeveloperTool=VisualStudio"
       },
       "applicationUrl": "http://localhost:1743"
     }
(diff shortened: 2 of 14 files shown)


src/Diagnostics.RuntimeHost/appsettings.json
```diff
@@ -104,6 +104,12 @@
     "Enabled": true,
     "MdmCertSubjectName": ""
   },
+  "ContainerAppsMdm": {
```
rekhaswaminathan, line 107: Michi, does this work involve setting up a new cert in KV to talk to MDM Container Apps endpoint?

src/Diagnostics.DataProviders/ConfigurationFactory.cs
```diff
@@ -90,6 +90,21 @@ protected override string GetValue(string prefix, string name)
                         return string.Empty;
                 }
             }
+            else if (prefix == "ContainerAppsMdm")
+            {
+                switch (name)
+                {
+                    case "MdmShoeboxEndpoint":
+                        return "https://az-int.metrics.nsatc.net";
```
rekhaswaminathan, line 98: Looks like you're returning a hard-coded endpoint. Will this work across N-clouds?

src/Diagnostics.RuntimeHost/Properties/launchSettings.json
```diff
@@ -3,7 +3,8 @@
     "Diagnostics.RuntimeHost": {
       "commandName": "Project",
       "environmentVariables": {
-        "ASPNETCORE_ENVIRONMENT": "Development"
+        "ASPNETCORE_ENVIRONMENT": "Development",
+        "AzureServicesAuthConnectionString": "RunAs=Developer; DeveloperTool=VisualStudio"
```
rekhaswaminathan, line 7: What does this change do?

src/Diagnostics.RuntimeHost/appsettings.json
michimune: Yes, it does. And I already finished installing the new cert to all the runtime host apps.

src/Diagnostics.DataProviders/ConfigurationFactory.cs
michimune: It's only for MockDataProviderConfigurationFactory. Actual endpoints are defined in a KV secret.
Oh, but I forgot settings for N-clouds. I will check.

src/Diagnostics.RuntimeHost/Properties/launchSettings.json
michimune: It's an experimental change for speeding up the local start time. It slightly improves.

src/Diagnostics.DataProviders/ConfigurationFactory.cs
rekhaswaminathan: Oh nvm. I don't think it will make a difference then

rekhaswaminathan (approved):

src/Diagnostics.DataProviders/ConfigurationFactory.cs
michimune: It turns out that we don't support container apps for national clouds yet. I will set it up once we get there.

ShekharGupta1988: Can we use some garbled url instead of an actual one for mock purposes?
trying to avoid accidental leakage of these urls via code base

ShekharGupta1988 (approved):

PR #814, opened 2022-01-19
changed type to List<(string, bool)> and added DevOpsGetBranchesConve…

DarrelDonald:
…rter

### DIFF
--- a/src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
+++ b/src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
@@ -77,7 +77,7 @@ namespace Diagnostics.RuntimeHost.Controllers
         [HttpGet(UriElements.DevOpsGetBranches)]
         public async Task<IActionResult> GetBranchesAsync(string resourceUri)
         {
-            object response = await _devOpsClient.GetBranchesAsync(resourceUri, this.HttpContext.Request.Headers[RequestIdHeaderName]);
+            List<(string, bool)> response = await _devOpsClient.GetBranchesAsync(resourceUri, this.HttpContext.Request.Headers[RequestIdHeaderName]);
             return Ok(response);
         }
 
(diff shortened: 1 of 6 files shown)


ShekharGupta1988 (approved):

src/Diagnostics.RuntimeHost/SerializationConvertors/AsRuntimeTypeConverter.cs
```diff
@@ -20,6 +22,8 @@ public class AsRuntimeTypeConverter<T> : JsonConverter<T>
             };
 
             options.Converters.Add(new ExceptionConverter());
+            options.Converters.Add(new DevOpsGetBranchesConverter());
+            options.Converters.Add(new DevOpsMakePRConverter());
```
michimune, line 26: I don't think you need these lines. You will need them if the DevOps responses need to be handled with AsRuntimeTypeConverter. But looking at your changes in Startup.cs, it's not the case.
It won't be harmful to leave them here, however minimizing the number of Converter objects is essential for performance. Can you try removing them and see if it still works?

michimune (approved):

michimune (approved):

src/Diagnostics.RuntimeHost/Controllers/DevOpsController.cs
```diff
@@ -77,7 +77,7 @@ public async Task<IActionResult> GetFileContentAsync(string filePathInRepo, stri
         [HttpGet(UriElements.DevOpsGetBranches)]
         public async Task<IActionResult> GetBranchesAsync(string resourceUri)
         {
-            object response = await _devOpsClient.GetBranchesAsync(resourceUri, this.HttpContext.Request.Headers[RequestIdHeaderName]);
+            List<(string, bool)> response = await _devOpsClient.GetBranchesAsync(resourceUri, this.HttpContext.Request.Headers[RequestIdHeaderName]);
```
rekhaswaminathan, line 80: I do see other methods in this controller that return Object type (MakePullRequest, GetFileContent, PushChanges). Are those working as expected?

DarrelDonald: GetFileContent and PushChanges are working as expected. I made changes for MakePullRequest in an update to the PR

rekhaswaminathan (approved):

PR #820, opened 2022-01-24
Allow a query to be executed on a few app service clusters

nmallick1:
Allow a query to be executed on a few app service clusters instead of all. This eliminated the need to perform a cross cluster union and enables scenarios where a query should be executed in few regions. This is being created to enable certain scenarios for static webapps.

### DIFF
--- a/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
@@ -153,7 +153,7 @@ namespace Diagnostics.DataProviders
             {"wawseas", new Tuple<string, string>("waws-prod-hk1-029", "EastAsia") }
         };
 
-        public async Task<DataTable> ExecuteQueryOnFewAppAppServiceClusters(List<string> appServiceClusterNames, string query, string operationName)
+        public async Task<DataTable> ExecuteQueryOnFewAppServiceClusters(List<string> appServiceClusterNames, string query, string operationName)
         {
             if (string.IsNullOrWhiteSpace(operationName))
             {
--- a/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
@@ -167,7 +167,7 @@ namespace Diagnostics.DataProviders
 
             if (appServiceClusterNames?.Count < 1)
             {
-                throw new ArgumentNullException(nameof(appServiceClusterNames), "App service kust cluster names list cannot be empty. Please supply at least one kusto cluster name where the query should be executed.");
+                throw new ArgumentNullException(nameof(appServiceClusterNames), "App service kusto cluster names list cannot be empty. Please supply at least one kusto cluster name where the query should be executed.");
             }
 
             List<Task<DataTable>> queryTask = new List<Task<DataTable>>();
--- a/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
+++ b/src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
@@ -144,6 +144,7 @@ namespace Diagnostics.DataProviders
 
         private Dictionary<string, Tuple<string, string>> clusterStampNameMapping = new Dictionary<string, Tuple<string, string>>() 
         {
+            //One random stamp name per region.
             {"wawswus", new Tuple<string, string>("waws-prod-bay-153", "WestUS") },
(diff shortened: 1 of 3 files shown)


src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -140,6 +140,67 @@ public async Task<DataTable> ExecuteQuery(string query, string stampName, string
...
+        }
+
+        private Dictionary<string, Tuple<string, string>> clusterStampNameMapping = new Dictionary<string, Tuple<string, string>>() 
+        {
+            {"wawswus", new Tuple<string, string>("waws-prod-bay-153", "WestUS") },
+            {"wawseus", new Tuple<string, string>("waws-prod-blu-189", "EastUS") },
+            {"wawscus", new Tuple<string, string>("waws-prod-dm1-187", "CentralUS") },
+            {"wawsweu", new Tuple<string, string>("waws-prod-am2-329", "WestEurope") },
+            {"wawsneu", new Tuple<string, string>("waws-prod-db3-169", "NorthEurope") },
+            {"wawseas", new Tuple<string, string>("waws-prod-hk1-029", "EastAsia") }
```
rekhaswaminathan, line 152: are these stamps specific to Static Web Apps only?

rekhaswaminathan (approved):

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
michimune: Please add a comment explaining what these stamps are.

michimune (approved):

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
nmallick1: Nope, these are just random stamp names so that ExecuteQuery is able to pickup the correct cluster based on the failover logic. One random stamp name per region.

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -140,6 +140,67 @@ public async Task<DataTable> ExecuteQuery(string query, string stampName, string
...
+        };
+
+        public async Task<DataTable> ExecuteQueryOnFewAppAppServiceClusters(List<string> appServiceClusterNames, string query, string operationName)
```
hforeste, line 155: nitpik; naming `ExecuteQueryOnFew**AppApp**ServiceClusters`

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -140,6 +140,67 @@ public async Task<DataTable> ExecuteQuery(string query, string stampName, string
...
+            }
+
+            if (string.IsNullOrWhiteSpace(query))
+            {
+                throw new ArgumentNullException(nameof(query), "Query cannot be empty. Please supply a query to execute.");
+            }
+
+            if (appServiceClusterNames?.Count < 1)
+            {
+                throw new ArgumentNullException(nameof(appServiceClusterNames), "App service kust cluster names list cannot be empty. Please supply at least one kusto cluster name where the query should be executed.");
```
hforeste, line 169: nitpik; Spelling

App service **kust** cluster names list cannot be empty. Please supply at least one kusto cluster name where the query should be executed.

hforeste:
```suggestion
                throw new ArgumentNullException(nameof(appServiceClusterNames), "App service kusto cluster names list cannot be empty. Please supply at least one kusto cluster name where the query should be executed.");

```

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
hforeste: So this is a superset over the `ExecuteQuery` method except detector author doensn't need to provide `StampName` since SWA don't have concepts of stamps?

nmallick1: That is correct @hforeste. SWA had a requirement where they wanted to query BlueRidgeDiagnostics table which does not exist in wawsneu, wawsweu, and wawseas clusters. The data however can be in any of the other regions. The answer is to union the clusters where the table is known to exist. Union however fails because it is a cross cluster query. This method is a counterpart of ExecuteQueryOnAllAppServiceClusters but like the name suggests, you can target which clusters the query should run on and the queries are fired directly at those clusters eliminating the need for union via cross cluster queries.

PR #824, opened 2022-01-27
Logging change

rekhaswaminathan:


src/Diagnostics.RuntimeHost/Program.cs
```diff
@@ -55,6 +55,10 @@ public static IHostBuilder CreateHostBuilder(string[] args)
                 webbuilder.ConfigureAppConfiguration((context, config) =>
                 {
                     var builtConfig = config.Build();
+                    config.AddEnvironmentVariables()
+                        .AddCommandLine(args)
+                        .Build();
+
```
hforeste, line 61: So by adding this earlier in the config we are saying that environment variables and command line arguments will hold lesser precedence than the appsettings captured in the encrypted json file, right?

src/Diagnostics.RuntimeHost/Startup.cs
```diff
@@ -85,36 +85,39 @@ public void ConfigureServices(IServiceCollection services)
...
                         {
+                            context.Request.Headers.TryGetValue(HeaderConstants.RequestIdHeaderName, out var requestId);
                             var allowedAppIds = Configuration["SecuritySettings:AllowedAppIds"].Split(",").Select(p => p.Trim()).ToList();
                             var claimPrincipal = context.Principal;
                             var incomingAppId = claimPrincipal.Claims.FirstOrDefault(c => c.Type.Equals("appid", StringComparison.CurrentCultureIgnoreCase));
                             if (incomingAppId == null || !allowedAppIds.Exists(p => p.Equals(incomingAppId.Value, StringComparison.OrdinalIgnoreCase)))
                             {
                                 context.Fail("Unauthorized Request");
-                                DiagnosticsETWProvider.Instance.LogRuntimeHostMessage($"AAD Authentication failed because incoming app id was not allowed");
+                                DiagnosticsETWProvider.Instance.LogRuntimeHostMessage($"AAD Authentication failed because incoming app id was not allowed, Request id {requestId}");
```
hforeste, line 95: Isn't the request id logged in the RequestID column?

src/Diagnostics.RuntimeHost/Program.cs
rekhaswaminathan: Oops. I did not realize that. I will revert this back. I moved it up to remove the confusion with `Secrets:Enabled` flag. Env variables & command line arguments should take more precedence than encrypted json.

src/Diagnostics.RuntimeHost/Startup.cs
hforeste: Oh i see, why not add a new requestid column to the logger? 
https://github.com/Azure/Azure-AppServices-Diagnostics/blob/e305a03d3046a358ec538392d791f995545af8dd/src/Diagnostics.Logger/ETWProvider/DiagnosticsETWProvider.cs#L99-L103

src/Diagnostics.RuntimeHost/Program.cs
hforeste: What is the confusion you are trying to clear with `Secrets:Enabled`? Maybe we can do a different change to help

rekhaswaminathan: Remember we used to replace the `Secrets:Enabled `flag to false when deploying to prod endpoints? I thought moving this up will make the env variables take precedence and not go to keyvault to fetch secrets in the next line. But we have encrypted settings that comes after that, which I missed

src/Diagnostics.RuntimeHost/Startup.cs
rekhaswaminathan: This event id is used many places like Services which needs additional code change to send the request id. I did not want to bring too many changes. This is rare cases when AAD fails, and we need to know the reason

hforeste: Ah i see.

hforeste (approved):

PR #825, opened 2022-01-28
updates for kusto diag leaders

yifguoMSFT:
- change GetClusterNameFromStamp function to use new diagleader when updating a missing region:cluster mapping
- add shadowing for wawswus and wawsneu
- turn on agg cluster for all regions
- not use weak consistency for all diagleader clusters

### DIFF
--- a/src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
+++ b/src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
@@ -87,8 +87,17 @@ namespace Diagnostics.DataProviders
                             {
                                 var regionName = ((string)dr["Region"]).ToLower();
                                 var clusterName = (string)dr["ClusterName"];
-                                _configuration.RegionSpecificClusterNameCollection.TryAdd(regionName, clusterName + "follower");
-                                _configuration.FailoverClusterNameCollection.TryAdd(regionName, clusterName);
+                                if (_configuration.WawsPrimaryToDiagLeaderClusterMapping != null && _configuration.WawsPrimaryToDiagLeaderClusterMapping.ContainsKey(clusterName))
+                                {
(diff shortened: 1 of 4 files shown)


rekhaswaminathan (approved):

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
```diff
@@ -87,8 +87,17 @@ public async Task<string> GetClusterNameFromStamp(string stampName)
...
+                                if (_configuration.WawsPrimaryToDiagLeaderClusterMapping != null && _configuration.WawsPrimaryToDiagLeaderClusterMapping.ContainsKey(clusterName))
+                                {
+                                    string diagLeaderCluster = _configuration.WawsPrimaryToDiagLeaderClusterMapping[clusterName];
+                                    _configuration.RegionSpecificClusterNameCollection.TryAdd(regionName, diagLeaderCluster);
+                                    _configuration.FailoverClusterNameCollection.TryAdd(diagLeaderCluster, clusterName);
+                                }
+                                else 
+                                {
+                                    _configuration.RegionSpecificClusterNameCollection.TryAdd(regionName, clusterName + "follower");
+                                    _configuration.FailoverClusterNameCollection.TryAdd(clusterName + "follower", clusterName);
```
nmallick1, line 99: Do we still need the follower cluster? The ultimate goal is to move away from followers and the failover cluster is the primary waws clusters as usual.

nmallick1 (approved):

src/Diagnostics.DataProviders/KustoHeartBeatService/KustoHeartBeatService.cs
yifguoMSFT: I think until we completely remove the dependency on followers, we still need this. I will remove this when we move to diag leaders completely.

PR #827, opened 2022-02-04
Support new RP for container apps

michimune:
This PR supports container app's new resource provider type.

### DIFF
--- a/src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/ContainerAppFilter.cs
+++ b/src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/ContainerAppFilter.cs
@@ -6,4 +6,11 @@
         {
         }
     }
+
+    public class MicrosoftAppContainerAppFilter : ResourceFilterBase
+    {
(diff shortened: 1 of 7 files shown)


yifguoMSFT (approved):

src/Diagnostics.ModelsAndUtils/Attributes/ResourceFilter/ContainerAppFilter.cs
```diff
@@ -6,4 +6,11 @@ public ContainerAppFilter(bool internalOnly = true) : base(ResourceType.Containe
         {
         }
     }
+
+    public class MicrosoftAppContainerAppFilter : ResourceFilterBase
+    {
+        public MicrosoftAppContainerAppFilter(bool internalOnly = true) : base(ResourceType.ContainerApp, internalOnly)
```
rekhaswaminathan, line 12: Will this still work if its `ResourceType.ContainerApp`?

michimune: I searched all the places that the ResourceType property is referenced. It seems to work.

rekhaswaminathan (approved):

PR #833, opened 2022-02-09
Add sap support topic Ids for runtimehost

cindylovescoding:


src/Diagnostics.ModelsAndUtils/Attributes/SupportTopic.cs
```diff
@@ -14,14 +14,24 @@ public class SupportTopic : Attribute, IEquatable<SupportTopic>
         /// See <see href="http://aka.ms/selfhelppreview"/>
         public string Id;
 
+        /// <summary>
+        /// Sap Support Topic Id
+        /// </summary>
+        public string SapSupportTopicId;
```
nmallick1, line 20: Should we use a property setter here and make sure that the supplied value is a GUID for Sap id's?

nmallick1 (approved):

src/Diagnostics.ModelsAndUtils/Attributes/SupportTopic.cs
cindylovescoding: Nazeef you mean that we should validate that it is a GUID before parsing SapId? 
I think it's fine as long as the SapId string in detector metadata is matching the SapId in the request. But I am open to change this if you think it is needed.

rekhaswaminathan:
@cindylovescoding Do you want to add this sap support id as part of metadata to the storage?

src/Diagnostics.ModelsAndUtils/Attributes/SupportTopic.cs
```diff
@@ -14,14 +14,24 @@ public class SupportTopic : Attribute, IEquatable<SupportTopic>
...
+
         /// <summary>
         /// Unique resource Id.
         /// </summary>
         public string PesId;
 
         public bool Equals(SupportTopic other)
         {
-            return (this.Id == other.Id && this.PesId == other.PesId);
+            return (this.Id == other.Id && this.PesId == other.PesId || this.SapSupportTopicId == other.SapSupportTopicId && this.SapProductId == other.SapProductId);
```
ShekharGupta1988, line 34: this should be :

((this.Id == other.Id && this.PesId == other.PesId) || (this.SapSupportTopicId == other.SapSupportTopicId && this.SapProductId == other.SapProductId))

src/Diagnostics.Scripts/EntityInvoker.cs
```diff
@@ -329,14 +329,14 @@ private void Validate()
                 {
                     this._entryPointDefinitionAttribute.SupportTopicList.ToList().ForEach(item =>
                     {
-                        if (string.IsNullOrWhiteSpace(item.Id))
+                        if (string.IsNullOrWhiteSpace(item.Id) || string.IsNullOrWhiteSpace(item.SapSupportTopicId))
```
ShekharGupta1988, line 332: Thinking about this change little bit:
- do we want to throw Script/Compile exception right now if the SAP Id is not present? This will block detector authors to publish their detectors next week.

How about this:
- In the beginning, We don't throw exceptions for this  SAP Id and let detector authors publish the detector changes
- When we have done the UI work and migration of support topic id to SAP Id in detector code, then we make this mandatory and throw exception on missing SAP Ids

src/Diagnostics.ModelsAndUtils/Attributes/SupportTopic.cs
cindylovescoding: "&&" has higher priority than "||" right Shekhar? So I guess they work the same?
For code readability , what you suggested is definitely better to understand.

cindylovescoding:
> @cindylovescoding Do you want to add this sap support id as part of metadata to the storage?

> 

That is correct Rekha.

rekhaswaminathan:
> That is correct Rekha.

@cindylovescoding  You have to make a change in `DiagEntity.cs` and `DiagEntityHelper.cs` files to include the SAP Support ID. You can follow the existing SupportTopicList as example.

src/Diagnostics.Scripts/EntityInvoker.cs
cindylovescoding: My intention here was to check at least **SupportTopicId** or **SapSupportTopicId** should exist. I put the check incorrectly but I was planning to do this so we don't block detector publish with author only specifying SapSupportTopicId :
`if (string.IsNullOrWhiteSpace(item.Id) && string.IsNullOrWhiteSpace(item.SapSupportTopicId))`

cindylovescoding: @ShekharGupta1988 Do you think we should allow authors publishing detector with only SapSupportTopicId? Or we should discard checking for SAP Ids?

src/Diagnostics.ModelsAndUtils/Attributes/SupportTopic.cs
nmallick1: Naah, it's fine.. I was just trying to ensure we catch a user who accidentally makes an error in Copy+Paste and then wonders why the detector is not being triggered. We can at least ensure that that the value is a GUID and avoid some percentage of issues.

cindylovescoding:
> > That is correct Rekha.
> 
> @cindylovescoding You have to make a change in `DiagEntity.cs` and `DiagEntityHelper.cs` files to include the SAP Support ID. You can follow the existing SupportTopicList as example.

Hey @rekhaswaminathan , "sapSupportTopic" will not be a new property for DiagEntity, but it will be part of the extended SupportTopicList. 

The current SupportTopicList is like:
[{id: "32758309", pesId: "32758309"}, {id: "32457411", pesId: "32457411"}]

With new SapSupportTopicId, the new SupportTopicList will be changed to something like this:
[{id: "32758309", pesId: "32758309", sapProductId: "d40f17bb-8b19-117c-f69a-d1be4187f657", sapSupportTopicId:"d40f17bb-8b19-117c-f69a-d1be4187f657" }, {id: "32457411", pesId: "32457411"}]

In this case, do we still need to change  `DiagEntity.cs` and `DiagEntityHelper.cs` ?

src/Diagnostics.Scripts/EntityInvoker.cs
ShekharGupta1988: Oh I see.. my bad. So basically one of the Id (either support topic id or SAp Id) should be present. Thats fine Cindy.
But the problem would be if some detector author provide just SAP Id and not support topic id. We will publish the detector but not surface it in case submission.

So until we do the whole wiring , its better to make support topic id as a mandatory field and we will change it once we migrate all detectors to SAP Id.

ShekharGupta1988 (approved):

rekhaswaminathan (approved):

src/Diagnostics.Scripts/EntityInvoker.cs
cindylovescoding: Makes sense Shekhar! Will do the change!

PR #845, opened 2022-02-26
Kusto related miscellaneous changes

yifguoMSFT:
- KustoDataProvider.ExecuteQueryOnHighPerfClusterWithFallback adds only backup query to meta data
- DiagnosticsRequestMiddleware extract detector name from uri and provide it to DataProviderContext
- use `detectorId_#` as the operation name for unnamed kusto queries


src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
```diff
@@ -352,14 +380,31 @@ public async Task<string> GetAggHighPerfClusterNameByStampAsync(string stampName
...
+                return await ExecuteClusterQuery(aggQuery, highPerfCluster, _kustoMap.MapDatabase(_configuration.DBName) ?? _configuration.DBName, requestId, operationName, addMetaData: false);
+            }
+        }
+
+
+        private string GenerateOperationName()
+        {
+            if (_detectorId != null)
+            {
+                return _detectorId + $"_{unnamedOperationCount++}";
```
nmallick1, line 403: Do you think we should have a delimiter other than "_" given that underscore is used commonly while naming something? 

It can help us in future analytics if we desire to.

nmallick1 (approved):

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
yifguoMSFT: I see, I am thinking the pattern like "detectorName_N" as the query name for the unnamed queries, what about "detectorName.N"? I think we cannot use "detectorName;N" as the client activity id uses semicolon as delimiter already.

nmallick1: How about a "|" symbol? Just want to avoid using something a real detector author may use and cause conflict with our reporting in future

yifguoMSFT: I see, if so what about "detectorName.N(auto generated)"?

nmallick1 (approved):

src/Diagnostics.DataProviders/DataProviders/KustoDataProvider.cs
yifguoMSFT: discussed offline and decided to apply the pattern "detectorId.unnamedQueryX"

